PR #5, opened 2018-02-07
adding an `IamDataFrame.to_excel()` function

danielhuppmann:
In the spirit of making the IamDataFrame more similar to a `pandas.DataFrame`, and because I typed these two lines way to often...

This PR also adds a unit test for the `timeseries()` function and makes sure that one warning is raised when over-writing meta-data during the unit tests.

### DIFF
--- a/pyam_analysis/core.py
+++ b/pyam_analysis/core.py
@@ -537,8 +537,7 @@ class IamDataFrame(object):
                 meta = meta[keep_col_match(meta[col], values)]
             return return_df(meta, display, idx_cols)
 
-    def to_excel(self, excel_writer, filters={}, exclude_cat=[],
-                 sheet_name='data', index=False, **kwargs):
+    def to_excel(self, excel_writer, sheet_name='data', index=False, **kwargs):
         """Write timeseries data to Excel using the IAMC template convention
         (wrapper for `pandas.DataFrame.to_excel()`)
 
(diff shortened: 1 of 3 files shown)


danielhuppmann:
rebased on the latest version including the plotting library

pyam_analysis/core.py
```diff
@@ -536,6 +537,30 @@ def metadata(self, meta=None, name=None, filters={},
...
+            filter by model, scenario, region, variable, level, year, category
+            see function _select() for details
+        exclude_cat: None or list of strings, default []
+            exclude all scenarios from the listed categories
+        sheet_name: string, default 'data'
+            name of the sheet that will contain the (filtered) IamDataFrame
+        index: boolean, default False
+            write row names (index)
+        """
+        df = self.timeseries(filters, exclude_cat).reset_index()
```
gidden: not sure if you want to think about how this will work in the future, the internal data representation is long-form (rather than wide-form). I might think that the call we want to make is as follows:

```
(df.filter(filters=foo)
    .add_metadata() # I think I added this function, but users can optionally do it here
    .to_wide()
    .capitalize_columns() # there might be a better pandas function here to call directly?
    .to_excel(*args, **kwargs) # can just pass them directly to pandas at this point
)
```

pyam_analysis/core.py
```diff
@@ -536,6 +537,30 @@ def metadata(self, meta=None, name=None, filters={},
...
+        df.columns = [str.title(i) if type(i) == str else i
```
gidden: see previous notes on investigating a `capitalize_columns()` function (which may not be necessary if we want to expose `df.rename()`.

tests/test_core.py
```diff
@@ -81,7 +90,10 @@ def test_category_pass():
 
 
 def test_load_metadata(test_ia):
-    test_ia.load_metadata(os.path.join(here, 'testing_metadata.xlsx'))
+    with pytest.warns(Warning) as record:
```
gidden: might be worth testing what warning we actually get. Nice pattern, haven't used it before =)

pyam_analysis/core.py
danielhuppmann: agree on using the `filter()` function, but disagree on adding the meta-data - the idea is to get something that conforms 100% to the "old" template, so that it can be used by the current web databases or the `Riamc` library...

Also, adding (let's say) 10 meta-data entries for a scenario with thousands of variables and ~10 timesteps roughly doubles the resulting file size.

danielhuppmann: re long vs. wide - that's what I have been (ab)using the `timeseries()`` function for. My hunch is that long is more efficient if we have timeseries data that is sparse in years...

danielhuppmann: re capitalize columns - having both strings and ints in the column header is causing problems for shorter notation...

danielhuppmann:
thanks @gidden for the comments, have cleaned it up to make it more pandas-style...

PR #6, opened 2018-02-08
Plots

gidden:
Adds some plotting functionality, we should discuss what else needs to be added here.


danielhuppmann:
Thanks @gidden for this really nice plotting library!

PR #8, opened 2018-02-20
use `ixmp.TimeSeries.timeseries()` filters, make IamDataFrame from pandas.DataFrame

danielhuppmann:
This PR increases integration with pandas: IamDataFrames can now be initialized directly from a `pandas.DataFrame`, and `append()` now also takes DataFrames as an argument. IamDataFrame can also be appended directly to another IamDataFrame.

This PR also implements full filtering when loading timeseries data from an `ixmp.TimeSeries` object.

### DIFF
--- a/pyam_analysis/core.py
+++ b/pyam_analysis/core.py
@@ -70,15 +70,16 @@ class IamDataFrame(object):
 
         Parameters
         ----------
-        data: ixmp.TimeSeries, ixmp.Scenario or data file
-            an instance of an TimeSeries or Scenario
-            (these two option require the 'ixmp' package as a dependency)
-            or a file with IAMC-style data
+        data: ixmp.TimeSeries, ixmp.Scenario, pandas.DataFrame or data file
+            an instance of an TimeSeries or Scenario (requires `ixmp`),
+            or pandas.DataFrame or data file with IAMC-format data columns
         regions: list
             list of regions to be imported
         """
         # import data from pandas.DataFrame or read from source
-        if has_ix and isinstance(data, ixmp.TimeSeries):
+        if isinstance(data, pd.DataFrame):
+            self.data = format_data(data)
+        elif has_ix and isinstance(data, ixmp.TimeSeries):
             self.data = read_ix(data, regions)
         else:
             self.data = read_data(data, regions, **kwargs)
@@ -92,23 +93,25 @@ class IamDataFrame(object):
         self.cat_color = {'uncategorized': 'white', 'exclude': 'black'}
         self.col_count = 0
 
-    def append(self, other, regions=None):
-        """Read timeseries data and append to IamDataFrame
+    def append(self, other, regions=None, **kwargs):
+        """Import or read timeseries data and append to IamDataFrame
 
         Parameters
         ----------
-        data: ixmp.TimeSeries, ixmp.Scenario or data file
-            this option requires the 'ixmp' package as a dependency
-            or a file with IAMC-style data
+        other: ixmp.TimeSeries, ixmp.Scenario, pandas.DataFrame or data file
+            an instance of an TimeSeries or Scenario (requires `ixmp`),
+            or pandas.DataFrame or data file with IAMC-format data columns
         regions: list
             list of regions to be imported
         """
         new = copy.deepcopy(self)
 
-        if has_ix and isinstance(other, ixmp.TimeSeries):
+        if isinstance(other, pd.DataFrame):
+            df = format_data(other)
+        elif has_ix and isinstance(other, ixmp.TimeSeries):
             df = read_ix(other, regions)
         elif os.path.isfile(other):
-            df = read_data(other, regions)
+            df = read_data(other, regions, **kwargs)
         else:
             raise ValueError("arg '{}' not recognized as valid source"
                              .format(other))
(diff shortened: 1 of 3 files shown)


danielhuppmann:
no more conflicts...

pyam_analysis/core.py
```diff
@@ -65,71 +65,94 @@ class IamDataFrame(object):
     (including validation of values, completeness of variables provided)
     as well as a number of visualization and plotting tools."""
 
-    def __init__(self, data, regions=None, **kwargs):
+    def __init__(self, data, regions=None, variables=None, units=None,
```
gidden: if I read this correctly, you can actually get rid of all of these ix kwargs. If you simply use **kwargs, you can do read_ix(..., **kwargs) and it will pass straight through.

pyam_analysis/core.py
```diff
@@ -65,71 +65,94 @@ class IamDataFrame(object):
...
+            self.data = read_data(data, regions, **kwargs)
 
-            # define a dataframe for categorization and other meta-data
-            self._meta = return_index(self.data, mod_scen,
-                                      drop_duplicates=True)
-            self.reset_category(True)
+        # define a dataframe for categorization and other meta-data
+        self._meta = return_index(self.data, mod_scen,
+                                  drop_duplicates=True)
+        self.reset_category(True)
```
gidden: use explicit argument name (`reset_exclude=True`)

pyam_analysis/core.py
```diff
@@ -65,71 +65,94 @@ class IamDataFrame(object):
...
+            filter by regions
+        variables : list of strings
+            filter by variables (only with ixmp objects)
+        units : list of strings
+            filter by units (only with ixmp objects)
+        years : list of integers
+            filter by years (only with ixmp objects)
         """
-        new = IamDataFrame(data=self)
+        new = copy.deepcopy(self)
```
gidden: you should be able to say

```
new = copy.deepcopy(self) if not inplace else self
```

and ignore the inplace logic later on

gidden: and we should probably develop a consistent style here. In other places I've used `ret` (short for `return`).

pyam_analysis/core.py
danielhuppmann: was on my to-do list as part of the category clean-up...

pyam_analysis/core.py
```diff
@@ -65,71 +65,94 @@ class IamDataFrame(object):
...
+        """Import or read timeseries data and append to IamDataFrame
 
         Parameters
         ----------
-        data: ixmp.TimeSeries, ixmp.Scenario or data file
-            this option requires the 'ixmp' package as a dependency
-            or a file with IAMC-style data
-        regions: list
-            list of regions to be imported
+        other: pyam-analysis.IamDataFrame, ixmp.TimeSeries, ixmp.Scenario,
```
gidden: it seems like there's a lot of conversion and stuff going on here. I would suggest sectioning that logic into two possible functions: `iam.read_data` (already exists) and `iam.to_dataframe` (which i think needs to be written. The first would read data from external sources, the second would translate in-memory data into an `iam.DataFrame`. The user would then be responsible for converting their data and providing it into these functions. The functions would therefore only take other iam.DataFrame objects.

pyam_analysis/core.py
```diff
@@ -675,7 +697,8 @@ def filter(self, filters, inplace=False):
             names=('model', 'scenario')
         )
         ret._meta = ret._meta.loc[idx]
-        return ret
+        if not inplace:
```
gidden: why was this added?

pyam_analysis/core.py
```diff
@@ -765,21 +787,15 @@ def read_ix(ix, regions=None):
     regions: list
         list of regions to be loaded from the database snapshot
     """
-    if not has_ix:
-        error = 'this option depends on the ixmp package'
-        raise SystemError(error)
     if isinstance(ix, ixmp.TimeSeries):
```
gidden: normally, its best to put the error checking first (`if condition: raise Error()`) and then have the rest of the function and top-level indentation.

pyam_analysis/core.py
```diff
@@ -755,7 +777,7 @@ def line_plot(self, *args, **kwargs):
 # %% auxiliary function for reading data from snapshot file
 
 
-def read_ix(ix, regions=None):
+def read_ix(ix, regions=None, variables=None, units=None, years=None):
```
gidden: here again, all additional arguments can be collapsed to `**kwargs`

pyam_analysis/core.py
gidden: ah, looks like the `format_data` function does mostly this.

pyam_analysis/core.py
```diff
@@ -803,26 +819,40 @@ def read_data(fname, regions=None, *args, **kwargs):
     else:
         df = pd.read_excel(fname, *args, **kwargs)
 
-    df = df.rename(columns={c: str(c).lower() for c in df.columns})
+    return(format_data(df))
```
gidden: no parens for return

pyam_analysis/core.py
```diff
@@ -765,21 +787,15 @@ def read_ix(ix, regions=None):
     regions: list
         list of regions to be loaded from the database snapshot
     """
-    if not has_ix:
-        error = 'this option depends on the ixmp package'
-        raise SystemError(error)
     if isinstance(ix, ixmp.TimeSeries):
-        df = ix.timeseries()
+        df = ix.timeseries(iamc=False, regions=regions, variables=variables,
```
gidden: the docs say this could also take a `Scenario`, is that no longer true?

pyam_analysis/core.py
```diff
@@ -803,26 +819,40 @@ def read_data(fname, regions=None, *args, **kwargs):
     else:
         df = pd.read_excel(fname, *args, **kwargs)
 
-    df = df.rename(columns={c: str(c).lower() for c in df.columns})
+    return(format_data(df))
+
+
+def format_data(data, regions=None):
```
gidden: suggest taking out regions argument, this should be filtered by the user later if desired.

pyam_analysis/core.py
danielhuppmann: because pandas logic says `return None` if `inplace=True`

pyam_analysis/core.py
```diff
@@ -803,26 +819,40 @@ def read_data(fname, regions=None, *args, **kwargs):
...
+
+    Parameters
+    ----------
+    data: pandas.DataFrame
+        dataframe to be converted to an IamDataFrame
+    regions: list
+        list of regions to be loaded from the database snapshot
+    """
+    # format columns to lower-case and check that all required columns exist
+    data = data.rename(columns={c: str(c).lower() for c in data.columns})
```
gidden: should consistently use inplace argument (you use it below)

pyam_analysis/core.py
danielhuppmann: `Scenario` is a subclass of `TimeSeries`, so this should return `True` in either case (I think?). I included both ixmp classes explicitly in the docs because I don't think that too many users will ever come in contact with the `TimeSeries` class.

pyam_analysis/core.py
danielhuppmann: I think that a function that takes many kinds of inputs is preferable from a user perspective, even if the docstrings are long. I think that internally, it's not quite as messy as it could have been...

pyam_analysis/core.py
danielhuppmann: I took that check out because `read_ix()` will only be called internally from functions that check for an `ixmp.TimeSeries` instance - and if someone calls this function from outside, they better know which packages they have installed and imported...

pyam_analysis/core.py
gidden: Yes, good catch... I just read way too much why they do it. I don't like it, but they have good reasons.

pyam_analysis/core.py
gidden: I don't think we disagree here. I'm arguing that there should be one or more "external functions" like `pd.read_csv` that support reading other datatypes and that all internal functions should assume (and check) that the type is an iam data frame. We can leave this for further cleanup if you wish.

pyam_analysis/core.py
gidden: Yes, I'm referring to the below else block: 
```
     else:
         error = 'arg ' + ix + ' not recognized as valid ixmp class'
         raise ValueError(error)
```

Which can be moved to the top of the function, and all other indentation at top-level.

pyam_analysis/core.py
gidden: Sounds good.

gidden:
Merging this as is, we can do further clean up tomorrow as needed.

PR #9, opened 2018-02-22
Refactor step3

gidden:
This should complete the refactor we've been working on. We should probably update the notebook too...


gidden:
Hey @danielhuppmann, I went ahead and updated the notebook. It may be most important to start there and see what you think. The main thing I ran into was our use of "uncategorized" data and plotting. I added the ability to shortcut the `categorize` function specifically for uncategorized data. However, we may want to rethink this... not sure. You'll see an example in the tutorial as well.

danielhuppmann:
HI @gidden , many thanks! Working through this tutorial and the cleaned-up code, so much beauty!

But there is one major break from the previous implementation, which only became apparent when I extended the unit-test data to implement checks for the refactored `metadata()` feature.

In the previous implementation, using `validate()` over a multi-year timeseries failed if *any* at least one value was outside of the bounds, while `categorize()` only passes (assign the category) if *all* values are within the range (i.e., the `max()` satisfies an upper bound rather than the `min()`) . Currently, `categorize()` assigns a category according to the minimum of an upper bound.

Also, I'd think that the range checks should be implemented as upper/lower-or-equal for the inclusive option.

I'll issue a PR with some proposals and a failing unit test to illustrate - maybe you can look at your reworked `_apply_criteria()` implementation?

gidden:
Sounds good @danielhuppmann. Given that all updates are built on top of this PR, I will merge this one so we can focus on smaller code reviews.

PR #11, opened 2018-02-22
migrate and refactor `metadata()` function

danielhuppmann:
This PR reimplements the `metadata()` to add columns to the metadata table, taking pd.Series, lists, ints or str's.

Depends on PR #9 for general refactoring and the PR into `gidden:refactor_step3` concerning the changed (wrong) behaviour of the `validate()` function.

PR #12, opened 2018-02-23
Refactor metadata rebase

gidden:
rebasing #11


gidden:
Reproducing @danielhuppmann's comment from #9 below 

```
HI @gidden , many thanks! Working through this tutorial and the cleaned-up code, so much beauty!

But there is one major break from the previous implementation, which only became apparent when I extended the unit-test data to implement checks for the refactored metadata() feature.

In the previous implementation, using validate() over a multi-year timeseries failed if any at least one value was outside of the bounds, while categorize() only passes (assign the category) if all values are within the range (i.e., the max() satisfies an upper bound rather than the min()) . Currently, categorize() assigns a category according to the minimum of an upper bound.

Also, I'd think that the range checks should be implemented as upper/lower-or-equal for the inclusive option.

I'll issue a PR with some proposals and a failing unit test to illustrate - maybe you can look at your reworked _apply_criteria() implementation?
```

gidden:
Hey @danielhuppmann. Due to the conflicts from squash and merge #9, I have taken your branch changes and rebased them here. I did a few things:

- I moved the test dataframe into testing_utils.py for easier use of testing out different data
- I shortened test model and scenario names (easier to fit 1 row on 1 line)
- I maintained the previous dependence of tests on the original test dataframe

I suspect the issue you raise is currently being masked by point 3. 

I do agree on the change for the range checks, however.

gidden:
I think this now conforms to your requirements, @danielhuppmann. I updated relevant tests to include and pass the tests you wrote with additional data. Note, that I have updated our "missing data" parameter (was `None`) to conform to [panda's guidelines](https://pandas.pydata.org/pandas-docs/stable/missing_data.html) (now is `np.nan`)

danielhuppmann:
thanks @gidden ! looking good!

PR #15, opened 2018-02-23
test out plus in regex

gidden:
add `+` to psuedo_regex


danielhuppmann:
what's the use case for that one?

gidden:
as discussed, we have variables like "CEDS+|..."

danielhuppmann:
thanks for the clarification and the tests!

PR #16, opened 2018-02-26
always use pseudo-regex in `pattern_match()`

danielhuppmann:
This PR changes the behaviour of the filter to add the option to select by exact match of model, scenario, region name, or use `*` as a wildcard.

### DIFF
--- a/pyam_analysis/core.py
+++ b/pyam_analysis/core.py
@@ -406,15 +406,14 @@ def _apply_filters(data, meta, filters):
 
         elif col == 'variable':
             level = filters['level'] if 'level' in filters.keys() else None
-            keep_col = pattern_match(data[col], values, True, level)
+            keep_col = pattern_match(data[col], values, level)
 
         elif col in ['year']:
             keep_col = years_match(data[col], values)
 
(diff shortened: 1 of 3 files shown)


pyam_analysis/core.py
```diff
@@ -406,15 +406,14 @@ def _apply_filters(data, meta, filters):
...
+            keep_col = pattern_match(data[col], values, level)
 
         elif col in ['year']:
             keep_col = years_match(data[col], values)
 
         elif col in ['level']:
             if 'variable' not in filters.keys():
-                keep_col = pattern_match(data['variable'], '*',
-                                         pseudo_regex=True, level=values)
+                keep_col = pattern_match(data['variable'], '*', level=values)
```
gidden: one question here - what is this if/then checking for? should there be an error if `level` is given and `variable` is not?

danielhuppmann:
If both `level` and `variable` is in the dictionary, this is checked anyway 10 lines above.
This line here is only applied if only `level` is given.

gidden:
ok, lgtm

PR #17, opened 2018-02-27
include check for required variables

danielhuppmann:
This PR adds a function to check whether required variables are included in the data.

Other miscellanous changes:
 - allow `unit` as argument for `filter()`
 - rework the `_check_rows()` function
 - rename options for `return_type` in `check_rows()` to `any` and `all`

### DIFF
--- a/pyam_analysis/core.py
+++ b/pyam_analysis/core.py
@@ -300,9 +300,8 @@ class IamDataFrame(object):
         filters: dict
             The following columns are available for filtering:
              - metadata columns: filter by category assignment in metadata
-             - 'model', 'scenario', 'region': takes a string or list of strings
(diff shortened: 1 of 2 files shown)


danielhuppmann:
@gidden, I changed the args vis-avis our earlier discussion, because passing over a filter as dictionary had some non-intuitive implications/results.

I may try to also implement a `iam.require_variables(df, ...)` version, too, at a later point.

gidden:
lgtm, let's chat about the outer function before I pull.

*very* nice refactor on `check_rows()` btw =)

danielhuppmann:
@gidden, top-level function included and unit-tested...

gidden:
lgtm!

PR #19, opened 2018-02-27
Plots

### LINKED ISSUE #21: add bar charts
initially should support years on x-axis and stacking of either regions or sector, unit on y-axis label.

### LINKED ISSUE #22: add pie chart
should be the same as bar chart but limited in one dimension (ie. only one bar)

gidden:
ready to merge now. closes #21 closes #22


gidden:
updating CI with cartopy in just a moment

gidden:
well that was more painful than I would have hoped..

gidden:
ready for review @danielhuppmann

gidden:
Should now be (finally) good to go @danielhuppmann. Yes, you are correct that currently building docs requires a file I have locally. That will be fixed when we address #20

danielhuppmann:
looks not just good to me, but outright awesome! thanks so much @gidden!

Issue #26, opened 2018-03-02
read data from `xlsx` fails due to `lru`


danielhuppmann:
Reading from `xlsx` fails after merging #19 , trying to identify the issue identifies `@lru_cache()` as the culprit.

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-3b2a587960f9> in <module>()
----> 1 df = iam.IamDataFrame(data='../data/postparis_dbsnapshot.xlsx', regions=['World'])

C:\ProgramData\Anaconda3\lib\site-packages\pyam_analysis\core.py in __init__(self, data, **kwargs)
     55             self.data = read_ix(data, **kwargs)
     56         else:
---> 57             self.data = read_files(data, **kwargs)
     58 
     59         # define a dataframe for categorization and other meta-data

C:\ProgramData\Anaconda3\lib\site-packages\pyam_analysis\utils.py in read_files(fnames, *args, **kwargs)
    116     for fname in fnames:
    117         logger().info('Reading {}'.format(fname))
--> 118         df = read_pandas(fname, *args, **kwargs)
    119         dfs.append(format_data(df))
    120 

TypeError: unhashable type: 'list'
```

Also, the current implemententation will not work if reading from multiple files with mixed wide vs. long formats (because `format_data()` is called after merging the imported dataframes.

danielhuppmann:
closed by #27

PR #27, opened 2018-03-02
removing `lru` failure to read xlsx, fix bug when reading multiple files with wide/long format

danielhuppmann:
This is a quick fix for #26.

@gidden: Happy to update that if you can enlighten me on why this `lru_cache` fixture might be confused about lists...


danielhuppmann:
The issues wasn't the file name in the arguments, but calling the `IamDataFrame()` with a list of regions as filter.

Suggest to remove `lru_cache` for reading IamDataFrame data from xlsx and ixmp.

PR #29, opened 2018-03-02
Rmap

### LINKED ISSUE #23: add complete cmip6 region mapping
and figure out how to best do the model checking

note we will need to likely add SSD for all models by copying SDN

### LINKED ISSUE #25: make map regions its own function
include option for r5, iso, etc.

remove it from region_plot function?

gidden:
closes #25 
closes #23


gidden:
wasn't using `--mpl`. these now pass on my machine @danielhuppmann should be ready to go

danielhuppmann:
thanks @gidden , install and unit tests work fine on my machine now

PR #31, opened 2018-03-05
support mapping for multiple regions

gidden:


danielhuppmann:
@gidden is this good to be merged?

pyam_analysis/core.py
```diff
@@ -516,40 +517,41 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
         inplace : bool, default False
             if True, do operation inplace and return None
         """
-        model = self.data['model'].unique()
-        if len(model) > 1:
-            raise ValueError(
-                'map_regions only supports mapping for a single model')
-        model = model[0]
+        models = self.data['model'].unique()
```
danielhuppmann: this could use `self.meta.get_levels(0)`...

gidden:
should be good to go, updated

PR #34, opened 2018-03-05
update tutorial notebook, re-insert auxiliary `models()` etc. functions

danielhuppmann:
This PR cleans up some inconsistencies in the tutorial notebook following the refactoring and inserting of the plotting library. It also makes some minor editorial changes to the index page and docstrings.

The PR also re-introduces the auxiliary `models()` etc. functions for an easier interface for non-expert users.

PR #35, opened 2018-03-05
Doc

gidden:


danielhuppmann:
interesting behaviour on Windows - when running `make html`, Windows is asking me which program I'd like to use to open `include.mk`...

Then, it fails with the error message `Config directory doesn't contain a conf.py file` (while I fixed in #34).

danielhuppmann:
after merging in #34, fails with

```
Exception occurred:
  File "C:\ProgramData\Anaconda3\lib\site-packages\sphinx_gallery\gen_gallery.py
", line 329, in sumarize_failing_examples
    "\n" + "-" * 79)
ValueError: Here is a summary of the problems encountered when running the examp
les

Unexpected failing examples:
C:\Users\huppmann\Documents\GitHub\pyam\doc\source\_examples\plot_regions.py fai
led leaving traceback:
Traceback (most recent call last):
  File "C:\Users\huppmann\Documents\GitHub\pyam\doc\source\_examples\plot_region
s.py", line 22, in <module>
    df.region_plot()
  File "C:\ProgramData\Anaconda3\lib\site-packages\pyam\core.py", line 593, in r
egion_plot
    ax = plotting.region_plot(df, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\pyam\plotting.py", line 154,
in region_plot
    crs = crs or cartopy.crs.PlateCarree()
AttributeError: 'NoneType' object has no attribute 'crs'
```

gidden:
This error is because you don't have cartopy installed... we should probably make that more obvious.

gidden:
And it looks like I got the windows make file a bit wrong, here is a potentially good source: https://msdn.microsoft.com/en-us/library/x6bt6xe7.aspx

PR #36, opened 2018-03-05
fixing `validate()` function if both upper and lower check are given

danielhuppmann:
This PR fixes an issue with the `validate()` function that does not correctly fail when checking both upper and lower bounds. Unit test to illustrate the failing test added.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -149,7 +149,7 @@ def test_validate_both(meta_df):
     obs = meta_df.validate({'Primary Energy': {'lo': 2.0, 'up': 6.5}},
                            exclude=False)
     assert len(obs) == 2
-    assert obs['year'].values[0] == 2010
+    assert list(obs['year'].values) == [2005, 2010]
 
 
 def test_validate_year(meta_df):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -145,6 +145,13 @@ def test_validate_lo(meta_df):
     assert obs['year'].values[0] == 2005
 
 
+def test_validate_both(meta_df):
+    obs = meta_df.validate({'Primary Energy': {'lo': 2.0, 'up': 6.5}},
+                           exclude=False)
+    assert len(obs) == 2
+    assert obs['year'].values[0] == 2010
+
+
 def test_validate_year(meta_df):
     obs = meta_df.validate({'Primary Energy': {'up': 5.0, 'year': 2005}},
                            exclude=False)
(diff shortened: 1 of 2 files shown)


gidden:
Not sure exactly what's being fixed here - happy to have you explain it to me tomorrow!

danielhuppmann:
@gidden, the original use case of the `validate()` function was to check whether a value was outside a range given by an upper or lower bound. For example:
`df.validate({variable, {'up': x * 1.5, 'lo': x * 0.5}` checks whether the value of variable is within 50% deviation from a reference `x`.

The refactoring changed the behaviour to only fail the test if *both*  upper and lower bound are violated.

It is similar to the problem between `all` and `any` in the return-test. Categorize if *all*  values pass the test, fail the validation if *any* value fails the test.

gidden:
Ok, I think I got that much. But I also thought it had to validate all years in a timeseries? I'm curious what are the exact values you expect in the test. I didn't see any timeseries in the test df for which all values match the criteria you described. Perhaps I am wrong though!

danielhuppmann:
Ok, got it - changed the behaviour of a unit test to cover the case that only the upper or lower bound is violated. Thanks for catching this @gidden !

tests/test_core.py
```diff
@@ -138,13 +138,21 @@ def test_validate_up(meta_df):
...
     assert obs['year'].values[0] == 2010
 
+    assert list(meta_df['exclude']) == [False, False] # assert none excluded
+
 
 def test_validate_lo(meta_df):
-    obs = meta_df.validate({'Primary Energy': {'lo': 2.0}}, exclude=False)
+    obs = meta_df.validate({'Primary Energy': {'up': 8, 'lo': 2.0}})
     assert len(obs) == 1
     assert obs['year'].values[0] == 2005
```
gidden: Last comment here, and I think it will be quick to implement. I believe that `a_scenario` is the one being caught here, correct? Could you test the scenario too (for clarity for potential future tests)?

tests/test_core.py
```diff
@@ -138,13 +138,21 @@ def test_validate_up(meta_df):
...
 
 
+def test_validate_both(meta_df):
+    obs = meta_df.validate({'Primary Energy': {'up': 6.5, 'lo': 2.0}})
+    assert len(obs) == 2
+    assert list(obs['year'].values) == [2005, 2010]
```
gidden: and here as well, I suspect it should also be 

```
assert list(obs['scenario'].values) == ['a_scenario', 'a_scenario2']
```

gidden:
Final comments, then should be good to go

PR #37, opened 2018-03-05
make cartopy and geopandas louder if not there

gidden:
This should fix the issue you raised in #35

### DIFF
--- a/pyam/plotting.py
+++ b/pyam/plotting.py
@@ -29,6 +29,7 @@ except ImportError:
 
 
 from pyam.run_control import run_control
+from pyam.utils import requires_package
 
 from pandas.plotting._style import _get_standard_colors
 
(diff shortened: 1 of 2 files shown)


danielhuppmann:
I appreciate the sentiment, but this doesn't have any impact on the behaviour on my PC...

should I go ahead and merge anyway, assuming that the problems on my machine are triggered by other, more fundamental issues?

gidden:
Well, we have a few options. One it to figure out why this isn't working as expected on your machine. A second is to make cartopy and geopandas hard dependencies. That could be problematic because cartopy can be a pain in the ass due to some other deps. Finally, I can reimplement this in a less "elegant" way. I'd like to go with option one if possible. Wait to merge until we can both sit down, if that's ok.

gidden:
Rebased. I also checked behavior on my side, and it indeed fails without `geopandas` installed with an import error saying "Requires geopandas".

PR #39, opened 2018-03-06
update install logo

gidden:


gidden:
Should have been in the rename pr!

PR #41, opened 2018-03-07
edits used for ceds work

gidden:
- removes 'year' column index name from `timeseries()`
- guarantees map cols are always lower


danielhuppmann:
lgtm, thanks @gidden !

PR #45, opened 2018-03-07
fixes for `pandas.combine_first()`, empty filtered IamDataFrames, general index for `metadata()`

danielhuppmann:
this PR brings in fixes for a few minor but annoying issues:
 - `pd.combine_first()` drops the dtype setting of the `meta['exlcude']` column, causing `|=` to fail
 - `filter()` shows a warning, if an empty IamDataFrame is returned
 - the  top-level functions `validate()` and `require_variable()` show the empty-IamDataFrame warning instead of performing a meaningless comparison
 - `metadata()` now accepts series with MultiIndex `['model', 'scenario', ...]` - this means that the output of the `timeseries()` function can be directly used as input for `metadata()` or with `apply()`. A unit test was adapted to illustrate/check the new behaviour. A ValueError is raised if the reduced MultiIndex has duplicates.
 - drop duplicates in `_meta_idx()` to yield correct counts in error messages

Some logger messages have also been updated and harmonized...

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -419,6 +419,9 @@ class IamDataFrame(object):
             pd.unique(list(zip(ret.data['model'], ret.data['scenario']))),
             names=('model', 'scenario')
         )
+        if len(idx) == 0:
+            logger().warning('Filtered IamDataFrame is empty!')
+
         ret.meta = ret.meta.loc[idx]
         if not inplace:
             return ret
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -263,7 +263,14 @@ class IamDataFrame(object):
             category column name (if not given by data series name)
         """
         if isinstance(meta, pd.Series) and \
-                meta.index.names == ['model', 'scenario']:
+                meta.index.names[0:2] == ['model', 'scenario']:
+            # reduce index dimentsions to model-scenario if necessary
+            if len(meta.index.names) > 2:
+                drop = list(range(2, len(meta.index.names)))
+                meta.index = meta.index.droplevel(drop)
+                if meta.index.duplicated():
+                    raise ValueError("non-unique ['model', 'scenario'] index!")
+            # check if trying to add model-scenario index not existing in self
             diff = meta.index.difference(self.meta.index)
             if not diff.empty:
                 error = "adding metadata for non-existing scenarios '{}'!"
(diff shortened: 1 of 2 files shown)


danielhuppmann:
added one more unit test and fixed the duplicated-check.

should be good to go, @gidden

pyam/core.py
```diff
@@ -351,17 +360,19 @@ def require_variable(self, variable, unit=None, year=None, exclude=False):
         keep = _apply_filters(self.data, self.meta, criteria)
         idx = self.meta.index.difference(_meta_idx(self.data[keep]))
 
-        if len(idx) == 0:
-            logger().info('All scenarios have the required variable')
+        i = len(idx)
```
gidden: can you change this variable to `n` or `nidx` to denote it is a length?

pyam/core.py
```diff
@@ -620,7 +635,7 @@ def region_plot(self, **kwargs):
 
 
 def _meta_idx(data):
-    return data[META_IDX].set_index(META_IDX).index
+    return data[META_IDX].drop_duplicates().set_index(META_IDX).index
```
gidden: can you add a test for this behavior?

pyam/core.py
```diff
@@ -263,13 +263,22 @@ def metadata(self, meta, name=None):
             category column name (if not given by data series name)
         """
         if isinstance(meta, pd.Series) and \
-                meta.index.names == ['model', 'scenario']:
+                meta.index.names[0:2] == ['model', 'scenario']:
```
gidden: this seems like it's getting a bit complicated. what about something like:

```
if isinstance(meta, pd.Series) and meta.index.isin(['model', 'scenario']):
    meta = (meta
                 .reset_index()
                 .loc[:, ['model', 'scenario', meta.name]]
                 .set_index(['model', 'scenario'])
                )
```

danielhuppmann:
@gidden, hope that testing for the length of the returned dataframe is good enough?

danielhuppmann:
I'm not sure if this is less complicated - it certainly has more lines now. But it has the benefit of taking pd.Series even if the index is not in the usual order... (and I adapted a unit test to check for this feature).

pyam/core.py
danielhuppmann: test added...

danielhuppmann:
@gidden, hold off a minute, in case you are looking at it right now...

gidden:
Holding off..

danielhuppmann:
good to go (I think)!

gidden:
Ok, looks great!

PR #47, opened 2018-03-15
change default behaviour of legend-display in `line_plot()`

danielhuppmann:
This PR changes the default behaviour of the `line_plot()` legend display - currently, the default is to show the legend even with hundreds of lines, which makes the plot small and hard to read.

Suggestion for new `legend` args options:
 - `None` (default), which displays the legend if less than 13 legend entries are shown
 - `boolean` with behaviour as currently implemented


gidden:
lgtm!

PR #49, opened 2018-03-19
WIP - first trial of a `scatter()`plot

danielhuppmann:
This PR adds a scatter figure to the plotting library.

This function does not directly use the `mpl.scatter()` function for two reasons:
 - `mpl.scatter()` does not take both string and tuple colors, causing issues with the default-color features
 - we will want a feature for more flexible scatter-and-line charts, for example connecting the markers across scenarios (from the same model) in a scatter plot. Therefore, this function anticipates this by adding a `groupby` arg.

This new function has some overlap with the `line_plot()`function, but overloading the line-plotting even more doesn't sound like the best strategy to me...


danielhuppmann:
@gidden, I assume that this test fails because the seaborn or mpl packages I used for generating the test scatterplot are not the same as used in CI - but how can I find out which the correct packages are?

danielhuppmann:
this function currently only works on metadata, because of the issue identified in #51.

gidden:
one pr left =)

danielhuppmann:
closing in favour of #95

PR #50, opened 2018-03-19
unique model map regions

### LINKED ISSUE #48: map_regions fails with multiple scenarios
due to multiple model looping

gidden:
fixes #48


danielhuppmann:
looks good to me, but wouldn't it make more sense to give meaningful name to the tests, rather than the issue number?

PR #52, opened 2018-03-19
check aggregate

danielhuppmann:
This PR adds a function to check whether the sub-categories of a timeseries (e.g., `foo|bar1`, `foo|bar2`) match the values of the variable of the level above (e.g., `foo`). This helps to identify reporting errors and missing entries.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -475,7 +475,7 @@ class IamDataFrame(object):
             return ret
 
     def check_aggregate(self, var_total, var_sector=None, units=None,
-                        exclude=True, threshold=0.1, multiplier=1):
+                        exclude=False, threshold=0.1, multiplier=1):
         """Check whether the timeseries values match the aggregation
         of sub-categories
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -425,8 +425,7 @@ class IamDataFrame(object):
             logger().info(msg.format(len(df), len(self.data)))
 
             if exclude_on_fail and len(idx) > 0:
-                logger().info('{} non-valid scenario{} will be excluded'
-                              .format(len(idx), '' if len(idx) == 1 else 's'))
+                self._exclude(df)
 
             return df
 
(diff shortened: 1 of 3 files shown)


gidden:
Some small changes requested. I think this is also a feature that would be very useful to document with examples (especially because I am not immediately sure how to use it). Perhaps it is worth starting an "Examples" page or something similar in the documentation to show how to use this function? I would imagine we will add lots more too, basically like the current plotting gallery but for other operations.

danielhuppmann:
@gidden, thanks for the feedback! should be good to go...

tests/test_core.py
```diff
@@ -202,6 +202,30 @@ def test_validate_top_level(meta_df):
     assert list(meta_df['exclude']) == [False, True]
 
 
+def test_check_aggregate_fail(meta_df):
+    obs = meta_df.check_aggregate('Primary Energy', exclude=True)
+    print(obs)
```
gidden: rm print

pyam/core.py
```diff
@@ -392,20 +393,60 @@ def validate(self, criteria={}, exclude=False):
...
 
-            if exclude and len(idx) > 0:
-                logger().info('{} non-valid scenario{} will be excluded'
-                              .format(len(idx), '' if len(idx) == 1 else 's'))
+            if exclude:
+                self._exclude(df)
 
             return df
 
+    def check_aggregate(self, var_total, var_sector=None, units=None,
```
gidden: a few questions on function names, feel free to disagree:

- `var_total` to `variable`
- `var_sector` to `components`

pyam/core.py
```diff
@@ -392,20 +393,60 @@ def validate(self, criteria={}, exclude=False):
...
+                        exclude=False, threshold=0.1, multiplier=1):
```
gidden: Is there a use case for the multiplier?

pyam/core.py
```diff
@@ -392,20 +393,60 @@ def validate(self, criteria={}, exclude=False):
...
+        if var_sector is None:
+            var_sector = self.filter({'variable': '{}|*'.format(var_total),
+                                      'level': 0}).variables()
+
+        df_total = _aggregate_by_variables(self.data, var_total, units)
+        df_sector = _aggregate_by_variables(self.data, var_sector, units)
+        n = len(df_total)
+
+        diff = df_total - multiplier * df_sector
+        diff = diff[abs(diff) > threshold]
```
gidden: perhaps you could use here

```
np.isclose(df_total, df_sector, **kwargs)
```

and then allow tolerances and such in the kwargs

see https://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.isclose.html

pyam/core.py
danielhuppmann: one use case in the SR1.5 context is to check that values are equal but with different signs (like CCS quantity vs. carbon removed must be equal, but one is usually reported as positive and the other as negative). Or going forward, making sure that timeseries in different units are equal...

danielhuppmann:
thanks @gidden for the suggestion with `np.isclose()` - a lot nice!also incorporated your renaming suggestions

rebased on top of PR #55, ready for review

gidden:
This might be easiest to cherry pick now

```
git checkout master
git checkout -b new-branch
git cherry-pick 26d63e0..bc41dd1
```

danielhuppmann:
rebased on `2695bc`

danielhuppmann:
should be good to be merged, @gidden

gidden:
lgtm!

PR #54, opened 2018-03-22
check for unique regions in plots

danielhuppmann:
This PR adds the `region` column to be checked for unique values in pie/bar/stack plots.


gidden:
lgtm!

Issue #40, opened 2018-03-06
filtering by metadata column fails if comparing string and `nan`


danielhuppmann:
Using a metadata column for filtering by a string fails if not the entire column is defined (i.e., has `nan` entries.

danielhuppmann:
This can be solved (as a quickfix) by `df.meta.col[[not isinstance(i, str) for i in df.meta.col]] = ''`.

But it should be checked and formatted in `load_metadata()`.

PR #55, opened 2018-03-26
allow using index as arg in `set_meta()` (refactored from `metadata()`)

danielhuppmann:
This PR allows using an index as an argument to the `set_meta()` function (previously `metadata()`), to assign a value in a meta column to all scenarios satisfying some filters. Previously, I would do the filter, use the index of the meta DataFrame to generate a pd.Series with the value, and pass that to `set_meta()`. This is now internalised in the function.

So the new use case would be `df.set_meta('some_category_name', 'some_column', df.filter('{some filters}').meta.index)`.

Last, setting column values in meta to `uncategorised` (from the pre-refactoring days, currently not operational as far as I can tell) is fixed - if a new column is added to `meta` and the data type of the value is a string, the default of that column is set to `uncategorized`, and to `np.NaN` otherwise.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -494,7 +494,6 @@ class IamDataFrame(object):
             logger().warning('Filtered IamDataFrame is empty!')
 
         ret.meta = ret.meta.loc[idx]
-        ret.index = ret.meta.index
         if not inplace:
             return ret
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -324,13 +324,6 @@ class IamDataFrame(object):
             if arg:
                 run_control().update({kind: {name: {value: arg}}})
 
-        if criteria == 'uncategorized':
-            self.meta[name].fillna(value, inplace=True)
-            msg = "{} of {} scenarios are uncategorized."
-            logger().info(msg.format(np.sum(self.meta[name] == value),
-                                     len(self.meta)))
-            return  # EXIT FUNCTION
-
         # find all data that matches categorization
         rows = _apply_criteria(self.data, criteria,
                                in_range=True, return_test='all')
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -267,37 +267,40 @@ class IamDataFrame(object):
         index: pd.MultiIndex
             index to be used for setting metadata column
         """
-        if index is not None:
-            meta = pd.Series(data=meta, index=index, name=name)
-        if isinstance(meta, pd.Series) and \
-                set(['model', 'scenario']).issubset(meta.index.names):
-            meta.name = name or meta.name
+        # create pd.Series from index and meta if provided
+        _meta = meta if index is None else pd.Series(data=meta, index=index)
+
+        # append by index of pd.Series if possible
+        if isinstance(_meta, pd.Series) and \
+                set(['model', 'scenario']).issubset(_meta.index.names):
+            _meta.name = name or _meta.name
             # reduce index dimensions to model-scenario only
-            meta = (meta
-                    .reset_index()
-                    .loc[:, ['model', 'scenario', meta.name]]
-                    .set_index(['model', 'scenario'])
-                    )
+            _meta = (_meta
+                     .reset_index()
+                     .loc[:, ['model', 'scenario', _meta.name]]
+                     .set_index(['model', 'scenario'])
+                     )
             # raise error if index is not unique
-            if meta.index.duplicated().any():
+            if _meta.index.duplicated().any():
                 raise ValueError("non-unique ['model', 'scenario'] index!")
             # check if trying to add model-scenario index not existing in self
-            diff = meta.index.difference(self.meta.index)
+            diff = _meta.index.difference(self.meta.index)
             if not diff.empty:
                 error = "adding metadata for non-existing scenarios '{}'!"
                 raise ValueError(error.format(diff))
 
             self._add_meta_column(name, meta)
-            self.meta = meta.combine_first(self.meta)
+            self.meta = _meta.combine_first(self.meta)
             #  quickfix for pandas.combine_first(), issue #7509
             self.meta['exclude'] = self.meta['exclude'].astype('bool')
             return  # EXIT FUNCTION
 
-        if isinstance(meta, pd.Series):
-            name = name or meta.name
-            meta = meta.tolist()
+        # append every other column type as list
+        if isinstance(_meta, pd.Series):
+            name = name or _meta.name
+            _meta = _meta.tolist()
 
-        self.meta[name] = meta
+        self.meta[name] = _meta
 
     def categorize(self, name, value, criteria,
                    color=None, marker=None, linestyle=None):
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -262,9 +262,13 @@ class IamDataFrame(object):
             (by `['model', 'scenario']` index if possible)
         name: str
             category column name (if not given by meta pd.Series.name)
-        index: pd.MultiIndex
+        index: pyam.IamDataFrame or pd.MultiIndex
             index to be used for setting metadata column
         """
+        # use meta.index if index arg is an IamDataFrame
+        if index is not None and isinstance(index, IamDataFrame):
+            index = index.meta.index
+
         # create pd.Series from index and meta if provided
         _meta = meta if index is None else pd.Series(data=meta, index=index)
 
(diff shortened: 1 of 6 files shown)


danielhuppmann:
solves #40

danielhuppmann:
Thanks @gidden for the comments!

Resetting the index of either the `data` or `meta` frames does not make sense in our framework (because some functions require the index of `meta` to be `['model', 'scenario']` and `data` needs to have a given set of columns).

Your proposed alternative `meta_index()` could just as easily be achieved using `meta.index` directly. I guess it's easiest to just remove the `index` property and merge the other features of this PR, and then do the more thorough review of nice-to-have pandas features later.

danielhuppmann:
As discussed previously, I removed the `IamDataFrame.index` - for now, let's continue to use `IamDataFrame.meta.index` directly.

Also, I refactored the function `metadata()` to `set_meta()`, closes #59

pyam/core.py
```diff
@@ -261,34 +261,43 @@ def metadata(self, meta, name=None):
...
         name: str
             category column name (if not given by meta pd.Series.name)
+        index: pd.MultiIndex
+            index to be used for setting metadata column
         """
-        if isinstance(meta, pd.Series) and \
-                set(['model', 'scenario']).issubset(meta.index.names):
-            meta.name = name or meta.name
+        # create pd.Series from index and meta if provided
+        _meta = meta if index is None else pd.Series(data=meta, index=index)
```
gidden: I don't think I see any need here to use `_meta`, perhaps just use `meta`?

pyam/core.py
```diff
@@ -261,34 +261,43 @@ def metadata(self, meta, name=None):
...
+
```
gidden: this may be a bit fancy, but I would suggest changes as follows:

```
try:
    append_by_idx = set(['model', 'scenario']).issubset(_meta.index.names)
except:
    append_by_idx = False

if not append_by_idx:
    if not name and not hasattr(meta, 'name'):
        raise ValueError('Must pass a name or use a pd.Series')
    name = name or _meta.name
    _meta = list(meta)
   self.meta[name] = meta

# then follow with the more complicated bit at normal indentation
```

pyam/core.py
```diff
@@ -317,28 +326,26 @@ def categorize(self, name, value, criteria,
...
+        # update metadata dataframe
+        self._add_meta_column(name, value)
+        self.meta.loc[idx, name] = value
+        msg = '{} scenario{} categorized as `{}: {}`'
+        logger().info(msg.format(len(idx), '' if len(idx) == 1 else 's',
+                                 name, value))
+
+    def _add_meta_column(self, name, value):
+        """Add a metadata column, set to `uncategorized` if str else np.nan"""
+        if name is not None and name not in self.meta:
```
gidden: this should probably raise an error if either of these are true instead of failing silently if they are not

pyam/core.py
```diff
@@ -261,34 +261,43 @@ def metadata(self, meta, name=None):
...
+        if isinstance(_meta, pd.Series) and \
+                set(['model', 'scenario']).issubset(_meta.index.names):
+            _meta.name = name or _meta.name
             # reduce index dimensions to model-scenario only
-            meta = (meta
-                    .reset_index()
-                    .loc[:, ['model', 'scenario', meta.name]]
-                    .set_index(['model', 'scenario'])
-                    )
+            _meta = (_meta
```
gidden: I think this can be `meta = meta.drop(index=[other indicies])[name]`

pyam/core.py
```diff
@@ -261,34 +261,43 @@ def metadata(self, meta, name=None):
...
+                     .loc[:, ['model', 'scenario', _meta.name]]
+                     .set_index(['model', 'scenario'])
+                     )
             # raise error if index is not unique
-            if meta.index.duplicated().any():
+            if _meta.index.duplicated().any():
                 raise ValueError("non-unique ['model', 'scenario'] index!")
             # check if trying to add model-scenario index not existing in self
-            diff = meta.index.difference(self.meta.index)
+            diff = _meta.index.difference(self.meta.index)
```
gidden: this should probably be combined with the below line as you do in 281

pyam/core.py
```diff
@@ -317,28 +326,26 @@ def categorize(self, name, value, criteria,
...
+            return  # EXIT FUNCTION
+
+        # update metadata dataframe
+        self._add_meta_column(name, value)
+        self.meta.loc[idx, name] = value
+        msg = '{} scenario{} categorized as `{}: {}`'
+        logger().info(msg.format(len(idx), '' if len(idx) == 1 else 's',
+                                 name, value))
+
+    def _add_meta_column(self, name, value):
```
gidden: Ok, if I understand this correctly, I would call this function `_new_meta_column()` and have it with a kwarg like `fill_nan=True`. Then you could call it with `_new_meta_column(name, fill_nan=not isstr(value))`.

You can obviously play with the fill_nan or fill_uncategorized as you prefer, but it wasn't clear how the value was being used in the calling function.

pyam/core.py
```diff
@@ -437,6 +444,7 @@ def filter(self, filters, keep=True, inplace=False):
             logger().warning('Filtered IamDataFrame is empty!')
 
         ret.meta = ret.meta.loc[idx]
+        ret.index = ret.meta.index
```
gidden: is this a holdover from before?

pyam/core.py
gidden: this is just for `_add_meta_column` right? I think that can be ameliorated with the suggested change to that function below.

pyam/core.py
danielhuppmann: that wouldn't work when passing over an index with more than 2 levels, need to bring it downb to the META_IDX first

pyam/core.py

pyam/core.py
danielhuppmann: agree on the renaming, but for the arg, the whole point was to move the checking for the type to the auxiliary function, not into the args when calling the function

pyam/core.py
danielhuppmann: reason for keeping `diff` explicitly is for using it in the ValueError message

pyam/core.py
danielhuppmann: was a bit of an effort and not sure if this is necessarily prettier - but implemented

pyam/core.py
danielhuppmann: another consideration, the `s[name]` doesn't work on `pd.Series`, so the reindexing also takes care of a harmonization to a `pd.DataFrame`

danielhuppmann:
rebased and implemented most suggestions
should be good to be re-reviewed, @gidden

pyam/core.py
```diff
@@ -261,35 +261,57 @@ def metadata(self, meta, name=None):
...
+                                                     name=name)
+
+        try:
+            append_by_idx = set(META_IDX).issubset(_meta.index.names)
+        except AttributeError:
+            append_by_idx = False
+
+        # if not possible to append by index
+        if not append_by_idx:
+            if isinstance(meta, collections.Iterable) and not isstr(meta):
```
gidden: Not blocking, but I recently implemented `isscalar` in message_ix: https://github.com/iiasa/message_ix/blob/master/message_ix/utils.py#L29

tutorial/pyam_first_steps.ipynb
```diff
@@ -449,9 +449,10 @@
     "We now use the categorization feature of the ``pyam`` package to group scenarios by temperature outcome by the end of the century.\n",
     "\n",
     "The first cell sets the ``'Temperature'`` categorization to the default `\"uncategorized\"`.\n",
-    "This may be helpful in this tutorial if you are going back and forth between cells to reset the assignment.\n",
+    "This is not necessary per se (setting a meta column via the categorization will mark all non-assigned rows as `\"uncategorized\"` (if the value is a string) or `np.nan`.\n",
+    "Still, having this cell may be helpful in this tutorial if you are going back and forth between cells to reset the assignment.\n",
     "\n",
-    "The function `categorize()` takes `color` and similar arguments, which can then be used by teh plotting library."
+    "The function `categorize()` takes `color` and similar arguments, which can then be used by th plotting library."
```
gidden: typo "th"?

gidden:
Just a typo, then good to go!

danielhuppmann:
thanks @gidden for catching the typo!

re the `is scalar()` function, note that the use case here is different than `is scalar()` - pd.Series (and maybe other array types) need to be cast to lists for the later workflow, while strings and scalars can be appended as is. Anyway, I added both your `isscalar()` function (because we'll probably need it at some point anyway) and a new `islistable()` function.

gidden:
thanks @danielhuppmann!

PR #56, opened 2018-03-28
google group add

gidden:


danielhuppmann:
thanks @gidden for setting up the google group!

Issue #57, opened 2018-03-28
Raw download from SSP Database can't be read


znicholls:
I think this is related to issue #40 

If you download data directly from the [SSP Database](https://tntcat.iiasa.ac.at/SspDb/dsd?Action=htmlpage&page=series) e.g.
[iamc_db_emissions_CO2.xlsx](https://github.com/IAMconsortium/pyam/files/1857947/iamc_db_emissions_CO2.xlsx), then try to read it in you get the following error

```python
xlsx_file_name = os.path.join('..','data','iiasa-ssp-portal','iamc_db_emissions_CO2.xlsx')
df = pyam.IamDataFrame(data=xlsx_file_name)
INFO:root:Reading `../data/iiasa-ssp-portal/iamc_db_emissions_CO2.xlsx`
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-8-af699d821ed6> in <module>()
      1 xlsx_file_name = os.path.join('..','data','iiasa-ssp-portal','iamc_db_emissions_CO2.xlsx')
----> 2 df = pyam.IamDataFrame(data=xlsx_file_name)

/home/zebedee/anaconda2/lib/python2.7/site-packages/pyam/core.pyc in __init__(self, data, **kwargs)
     57             self.data = read_ix(data, **kwargs)
     58         else:
---> 59             self.data = read_files(data, **kwargs)
     60 
     61         # define a dataframe for categorization and other metadata indicators

/home/zebedee/anaconda2/lib/python2.7/site-packages/pyam/utils.pyc in read_files(fnames, *args, **kwargs)
    116         logger().info('Reading `{}`'.format(fname))
    117         df = read_pandas(fname, *args, **kwargs)
--> 118         dfs.append(format_data(df))
    119 
    120     return pd.concat(dfs)

/home/zebedee/anaconda2/lib/python2.7/site-packages/pyam/utils.pyc in format_data(df)
    135         df = pd.melt(df, id_vars=IAMC_IDX, var_name='year',
    136                      value_vars=numcols, value_name='value')
--> 137     df['year'] = pd.to_numeric(df['year'])
    138 
    139     # drop NaN's

/home/zebedee/anaconda2/lib/python2.7/site-packages/pandas/core/tools/numeric.pyc in to_numeric(arg, errors, downcast)
    131             coerce_numeric = False if errors in ('ignore', 'raise') else True
    132             values = lib.maybe_convert_numeric(values, set(),
[...]

PR #58, opened 2018-04-12
Add capability to read raw download from SSP database

znicholls:
Solves issue #57 

The solution here isn't very clever and I may have broken some naming conventions but all the tests seem to pass, including the new one. I'm happy to fix these issues if you can tell me how you'd like that done.


pyam/utils.py
```diff
@@ -129,6 +129,18 @@ def format_data(df):
         missing = list(set(IAMC_IDX) - set(df.columns))
         raise ValueError("missing required columns `{}`!".format(missing))
 
+    if 'notes' in df.columns:
+        logger().info('Ignoring notes column in dataframe')
+        df.drop(columns='notes',inplace=True)
+        # remove SSP Public Database row
+        database_label_rows = [
```
danielhuppmann: I'd suggest to drop all models starting with `@`, because there are many other (non-public) database instances from which one can download such snapshot files.

pyam/utils.py
```diff
@@ -129,6 +129,18 @@ def format_data(df):
...
+            'SSP Public Database' in model
+            for model in df.model
+        ]
+        df.drop(df[database_label_rows].index, inplace=True)
+
+
+
```
danielhuppmann: please remove extra empty lines

pyam/utils.py
znicholls: When you say all models starting with @, do you mean drop any items in the 'model' column which start with @ or something different?

danielhuppmann: yes, I meant drop all rows where the `model` entry starts with `@`

pyam/utils.py
```diff
@@ -129,6 +129,18 @@ def format_data(df):
         missing = list(set(IAMC_IDX) - set(df.columns))
         raise ValueError("missing required columns `{}`!".format(missing))
 
+    if 'notes' in df.columns:
+        logger().info('Ignoring notes column in dataframe')
+        df.drop(columns='notes',inplace=True)
```
gidden: space between args, please (pep 8)

pyam/utils.py
```diff
@@ -129,6 +129,18 @@ def format_data(df):
...
+        # remove SSP Public Database row
+        database_label_rows = [
+            'SSP Public Database' in model
+            for model in df.model
+        ]
+        df.drop(df[database_label_rows].index, inplace=True)
```
gidden: I think this can be a one-liner:

```
df = df[~df.model.str.contains('SSP Public Database')]
```

gidden:
Just a few comments from my side. Looks good! Thanks Zeb!

pyam/utils.py
gidden: From what I remember of the database version, this could also be:

```
df = df.iloc[:-2]
```

Just a thought!

danielhuppmann: I'd argue against blindly removing the last column - we may want to add more columns in later snapshot-export versions (e.g., user/timestamp of upload).

The cleanest way would be to remove all columns that are not in the `IAMC_IDX` or years (i.e., int).

znicholls: `df = df.iloc[:-2]` removes just the last row right? I'm not sure how best to implement removing any columns not in `IAMC_IDX` so will leave that up to you to add if you want.

znicholls:
Thanks both. My python is rusty so apologies for various abuses of style and multi-line solutions to one-line problems etc.

gidden:
Lgtm, I'll let @danielhuppmann make the final decision and merge.

pyam/utils.py
```diff
@@ -129,6 +129,14 @@ def format_data(df):
         missing = list(set(IAMC_IDX) - set(df.columns))
         raise ValueError("missing required columns `{}`!".format(missing))
 
+    if 'notes' in df.columns:
+        logger().info('Ignoring notes column in dataframe')
+        df.drop(columns='notes', inplace=True)
+        df = df[
+            (~df.model.str.contains('SSP Public Database'))
```
danielhuppmann: Sorry for raising this again, but you shouldn't need checking both the database name and the `@` - just checking for the `@` should be enough.

znicholls: No need to apologise, I wasn't thinking (long day)

znicholls:
Alright I'll fix in the morning with whatever logic you recommend given your knowledge of the various databases this is likely to interact with

danielhuppmann:
@znicholls sorry that this last hoop to jump through is such a pain! I concur with Matt that checking for the string `database` is the best solution at this point...

danielhuppmann:
thanks @znicholls!

znicholls:
Thanks for hand holding, hopefully next contribution will be smoother!

PR #60, opened 2018-04-19
Filter by kwargs

danielhuppmann:
This PR changes the behaviour of the `filter()` to take keyword arguments instead of a dictionary, which is usually easier to type and process. A warning is shown if a dictionary is passed to the refactored `filter()` function as the first arg to indicate that this use is deprecated.

All tests and notebooks are refactored.


pyam/core.py
```diff
@@ -406,12 +406,16 @@ def validate(self, criteria={}, exclude=False):
 
             return df
 
-    def filter(self, filters, keep=True, inplace=False):
+    def filter(self, keep=True, inplace=False, **kwargs):
```
gidden: We need to decide if we want to make this API change now. How many users do we have right now? The safest way is to do the following:

- make the argument `filters=None`
- add the following check at the beginning of the function:

```
if filters is not None:
    warnings.warn('filters keyword argument in filters() is deprecated and will be removed in the next release')
    kwargs.update(filters)  
```

pyam/core.py
```diff
@@ -760,8 +766,7 @@ def validate(df, *args, **kwargs):
         filter by data & metadata columns, see function 'filter()' for details,
         filtering by 'variable'/'year' is replaced by arguments of 'criteria'
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
+    fdf = df.filter(**kwargs.pop('filters', {}))
```
gidden: then this would just be df.filter(**kwargs)

pyam/core.py
```diff
@@ -778,8 +783,7 @@ def require_variable(df, *args, **kwargs):
     filters: dict, optional
         filter by data & metadata columns, see function 'filter()' for details
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
+    fdf = df.filter(**kwargs.pop('filters', {}))
```
gidden: same here

pyam/core.py
```diff
@@ -798,8 +802,7 @@ def categorize(df, *args, **kwargs):
         filter by data & metadata columns, see function 'filter()' for details,
         filtering by 'variable'/'year' is replaced by arguments of 'criteria'
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
+    fdf = df.filter(**kwargs.pop('filters', {}))
```
gidden: and here

danielhuppmann:
Implemented the API change with deprecation warning as suggested.

Also changed the kwargs of `validate()` and `require_variable()` to `exclude_on_fail` instead of `exclude` to prevent overlap with `exclude` column in the meta df when filtering.

@gidden, should be good to go!

pyam/core.py
```diff
@@ -770,61 +778,51 @@ def _apply_criteria(df, criteria, **kwargs):
...
-        filter by data & metadata columns, see function 'filter()' for details,
-        filtering by 'variable'/'year' is replaced by arguments of 'criteria'
+    args and kwargs: see IamDataFrame.validate() and .filter() for details
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
+    fdf = df.filter(**kwargs)
     if len(fdf.data) > 0:
-        vdf = fdf.validate(*args, **kwargs)
+        vdf = fdf.validate(criteria, exclude_on_fail)
```
gidden: use explicit kwargs here e.g. `criteria=criteria`

pyam/core.py
```diff
@@ -770,61 +778,51 @@ def _apply_criteria(df, criteria, **kwargs):
...
-    filters: dict, optional
-        filter by data & metadata columns, see function 'filter()' for details
+    args and kwargs: see IamDataFrame.validate() and .filter() for details
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
+    fdf = df.filter(**kwargs)
     if len(fdf.data) > 0:
-        vdf = fdf.require_variable(*args, **kwargs)
+        vdf = fdf.require_variable(variable, unit, year, exclude_on_fail)
```
gidden: same here

pyam/core.py
```diff
@@ -770,61 +778,51 @@ def _apply_criteria(df, criteria, **kwargs):
...
-    filters: dict, optional
-        filter by data & metadata columns, see function 'filter()' for details,
-        filtering by 'variable'/'year' is replaced by arguments of 'criteria'
+    args and kwargs: see IamDataFrame.categorize()  and .filter() for details
     """
-    filters = kwargs.pop('filters', {})
-    fdf = df.filter(filters)
-    fdf.categorize(*args, **kwargs)
+    fdf = df.filter(**kwargs)
+    fdf.categorize(name, value, criteria, color, marker, linestyle)
```
gidden: here as well (where they are kwargs)

gidden:
Thanks @danielhuppmann !

PR #62, opened 2018-04-26
Add rename function

OFR-IIASA:


pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+        ----------
+        mapping: dict
+            for each column where entries should be renamed, provide current name and target name
+            {<column name>: {<current_name_1>: <target_name_1>, <current_name_2>: <target_name_2>}}
+        inplace: bool, default False
+            if True, do operation inplace and return None
+        """
+        ret = copy.deepcopy(self) if not inplace else self
+        for col in mapping:
+            if col in ['region', 'variable', 'unit']:
```
gidden: This can be `if not and save an indention

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+                ret.data.loc[:, col] = self.data.loc[:, col].replace(mapping[col])
+            else:
+                raise ValueError('renaming by {} not supported!'.format(col))
+
+        ret.data = ret.data.groupby(IAMC_IDX + ['year']).sum().reset_index()
```
gidden: @danielhuppman this is your YEAR_IDX right?

danielhuppmann: not quite, `YEAR_IDX` as proposed in #52 (not yet merged) is again different - we should probably come up with a better structure for the index-column collections at some point down the line...

danielhuppmann: How about ``LONG_IDX = IAMC_IDX + ['year']`` (i.e., `['model', 'scenario', 'region', 'variable', 'unit', 'year']` for all columns except `value`) in the `data` table?

gidden:
thanks @OFR-IIASA !

PR #63, opened 2018-04-27
Add conv unit function

OFR-IIASA:


pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
 
             return df
 
+    def conv_unit(self, unit_conv, inplace=False):
+        """ Converts units based on provided unit conversion factors
+
+        Parameters
+        ----------
+        unit_conv: dict
+            for each unit for which a conversion should be carried out, provide current unit and target unit and conversion factor
```
danielhuppmann: please pep8

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
 
             return df
 
+    def conv_unit(self, unit_conv, inplace=False):
+        """ Converts units based on provided unit conversion factors
```
danielhuppmann: please remove empty space before "Converts"

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+        Parameters
+        ----------
+        unit_conv: dict
+            for each unit for which a conversion should be carried out, provide current unit and target unit and conversion factor
+            {<current unit>: [<target unit>, <conversion factor>]}
+        inplace: bool, default False
+            if True, do operation inplace and return None
+        """
+        ret = copy.deepcopy(self) if not inplace else self
+        for unt in unit_conv:
```
danielhuppmann: suggest to rename `unt` to `unit` for readability

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+            if not type(unit_conv[unt][1]) in [int, np.integer, float]:
+                raise ValueError(
+                    'unit conversion factor for {} is not numeric!'.format(unt))
+            ret.data.loc[ret.data['unit'] == unt, 'value'] = ret.data.loc[ret.data['unit']
```
danielhuppmann: the expression `ret.data['unit'] == unt` is used three times, better to starts with 

```
rows = ret.data['unit'] == unt
```

and then use `rows` in the subsequent two lines?

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
 
             return df
 
+    def conv_unit(self, unit_conv, inplace=False):
```
danielhuppmann: why not use `convert_unit()` - still short and concise, but also immediately obvious?

danielhuppmann:
@gidden, I am not sure if the input arg format using a dictionary of str -> list is most appropriate here - would a tuple be better? Or some other input args structure?

pyam/core.py
gidden: Agreed

pyam/core.py
gidden: This is used multiple times, suggest updating to 
```
where = ret.data['unit'] == unt
```

pyam/core.py
gidden: suggest looping as 

```
for current_unit, (new_unit, factor) in unit_conv.items():
    do stuff
```

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+        """ Converts units based on provided unit conversion factors
+
+        Parameters
+        ----------
+        unit_conv: dict
```
gidden: rename perhaps to `mapping` or `conversion_mapping`

pyam/core.py
```diff
@@ -405,6 +405,28 @@ def validate(self, criteria={}, exclude=False):
...
+            for each unit for which a conversion should be carried out, provide current unit and target unit and conversion factor
+            {<current unit>: [<target unit>, <conversion factor>]}
+        inplace: bool, default False
+            if True, do operation inplace and return None
+        """
+        ret = copy.deepcopy(self) if not inplace else self
+        for unt in unit_conv:
+            if not type(unit_conv[unt][1]) in [int, np.integer, float]:
```
gidden: here just use pd.as_numeric(factor) and let the error pass through to user if there is one

pyam/core.py
gidden: this can be more concise:

```
ret.data.loc[where, 'value'] *= factor
```

gidden:
looks good, a few more suggestions added to those by @danielhuppmann

gidden:
this will probably need a rebase

pyam/core.py
gidden: i.e., 

```
factor = pd.as_numeric(factor)
```

pyam/core.py
```diff
@@ -427,6 +427,26 @@ def rename(self, mapping, inplace=False):
         if not inplace:
             return ret
 
+    def convert_unit(self, conversion_mapping, inplace=False):
+        """Converts units based on provided unit conversion factors
+
+        Parameters
+        ----------
+        conversion_mapping: dict
+            for each unit for which a conversion should be carried out, provide current unit and target unit and conversion factor
```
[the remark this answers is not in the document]
OFR-IIASA: i did. but maybe it doesnt catch it because it is documentation and not part of the code?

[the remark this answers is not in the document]
gidden: Ok, can you just manually break this then at 79 chars? Something like

```
        Parameters
        ----------
        conversion_mapping: dict
            for each unit for which a conversion should be carried out, provide current unit 
            and target unit and conversion factor
```

gidden:
lgtm, @danielhuppmann ?

gidden:
thanks @OFR-IIASA !

PR #64, opened 2018-06-21
add windows ci

gidden:


gidden:
should be g2g after this round of ci @danielhuppmann

danielhuppmann:
thanks @gidden for this heroic effort!

PR #65, opened 2018-06-22
add a `filter_by_meta() ` function

danielhuppmann:
This PR implements a function to join a generic pd.DataFrame (indexed on `['model', 'scenario']`) with specific metadata columns from an IamDataFrame and apply filters on each column.

This is useful in the following workflow:
1. retrieve a pd.DataFrame from an IamDataFrame using `timeseries()`
2. do some computation on the df
3. add relevant metadata like categories from the original IamDataFrame
4. continue working on the df, eg using `groupby()`on the categories

This is sort-of possible with `as_pandas()` and `filter()`, but quite cumbersome, because `as_pandas()` returns value-columns (years)  and metadata columns (Step 1 above). So one would have to make some bypass for step 2 to just operate on the year-columns. I found the workflow described above more flexible and also easier to read (because less bypassing required).

The use of `col=None` for only joining a column without filter is modelled on `slice(None)` of `pd.MultiIndex`

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -960,31 +960,42 @@ def check_aggregate(df, variable, components=None, units=None,
         return vdf
 
 
-def filter_by_meta(data, df, **kwargs):
-    """Add columns from IamDataFrame.meta table to a pd.DataFrame and apply
-    filters (optional)
+def filter_by_meta(data, df, join_meta=False, **kwargs):
+    """Filter by and join meta columns from an IamDataFrame to a pd.DataFrame
 
     Parameters
     ----------
-    data: pd.DataFrame instance, index must include `['model', 'scenario']`
-        the DataFrame to which meta columns are to be joined
+    data: pd.DataFrame instance
+        DataFrame to which meta columns are to be joined,
+        index or columns must include `['model', 'scenario']`
     df: IamDataFrame instance
-        the IamDataFrame from which meta columns are joined
+        IamDataFrame from which meta columns are filtered and joined (optional)
+    join_meta: bool, default False
+        join selected columns from `df.meta` on `data`
     kwargs:
         meta columns to be joined, where `col=...` applies filters
         by the given arguments (using `utils.pattern_match()`) and `col=None`
         joins the column without filtering
     """
-    data = data.copy()
-    idx = data.index.copy()
-    data.index = data.index.droplevel(list(set(idx.names) - set(META_IDX)))
-    data = data.join(df.meta[list(kwargs.keys())])
-    data.index = idx
+    meta = df.meta[list(kwargs)].copy()
 
-    # filter by joined columns
-    keep = np.array([True] * len(data))
+    # filter meta by columns
+    keep = np.array([True] * len(meta))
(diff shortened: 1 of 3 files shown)


gidden:
Ok, conceptually I get it now. Do we want to join the metadata though? Or perhaps that should be optional?

pyam/core.py
```diff
@@ -955,3 +955,33 @@ def check_aggregate(df, variable, components=None, units=None,
...
+    data = data.copy()
+    idx = data.index.copy()
+    data.index = data.index.droplevel(list(set(idx.names) - set(META_IDX)))
+    data = data.join(df.meta[list(kwargs.keys())])
+    data.index = idx
+
+    # filter by joined columns
+    keep = np.array([True] * len(data))
+    for col, values in kwargs.items():
+        if values is not None:
```
gidden: I see here this enables `integer=None` below. But wouldn't it be easier to just not have `integer` in the call at all?

gidden:
Ok, I think I see where the implementation is going. It might be simpler to

- subset metadata for values in kwargs first
- extract the model/scenario index from the subset dataframe
- return data subset with subset index

Instead of joining the dataframes first, then doing the subset (also more computationally expensive)

gidden:
And can you please also add to release notes! =)

pyam/core.py
danielhuppmann: I used the `column=None` as a shorthand for joining the meta column with applying a filter, similar to `slice(None)` of `pd.MultiIndex`.

danielhuppmann: To explain the rationale again - it is often useful to have that information for further processing like `groupby()`.

danielhuppmann:
Re whether joining is optional: I would make it the default (because in my use cases, I usually want to filter and do something else), but we can make it optional to not have it joined (just perform the filter). What would be a good kwarg-name, `join_meta`?

danielhuppmann:
note to myself: if no index is given, check if `model` and `scenario` columns exist and create index from that

gidden:
let's talk about this briefly when we meet tomorrow morning.

danielhuppmann:
sticklered, rebased, release-noted, error-messaged, and a more efficient workflow (first filter, then join optionally)

and it now also accepts `['model', 'scenario']` as columns instead of index and returns the df in the same way as it was inputted (and yes, there is a test for that, too)

@gidden, good to go once the tests pass

pyam/core.py
```diff
@@ -958,3 +958,49 @@ def check_aggregate(df, variable, components=None, units=None,
...
+    df: IamDataFrame instance
+        IamDataFrame from which meta columns are filtered and joined (optional)
+    join_meta: bool, default False
+        join selected columns from `df.meta` on `data`
+    kwargs:
+        meta columns to be joined, where `col=...` applies filters
+        by the given arguments (using `utils.pattern_match()`) and `col=None`
+        joins the column without filtering
+    """
+    print(data.columns)
```
gidden: rm print statements (sorry!)

gidden:
thanks @danielhuppmann!!

PR #66, opened 2018-06-25
bugfix for `interpolate()` - duplication of values

danielhuppmann:
This PR fixes a bug in the ` interpolate` function, where values are duplicated if they are already defined. The fix only interpolates (and adds to the `IamDataFrame.data` only years that are missing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,2 +1,4 @@
 
 # Next Release
+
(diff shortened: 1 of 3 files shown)


gidden:
Hey @danielhuppmann, looks good to me. However, could you add a test that fails without this fix? Can you also add this to the release notes?

gidden:
Probably easiest to rebase on master to get the release notes, FYI.

danielhuppmann:
rebased, extended the test for the `interpolate()` function, added a release note

@gidden, should be good to go!

gidden:
thanks @danielhuppmann !

PR #82, opened 2018-08-14
WIP - do not merge - add an `append` kwarg to the `rename()` function

danielhuppmann:

- [ ] Tests Added
- [x] Documentation Added

# WIP - do not merge

Needs more careful checks whether this new feature doesn't duplicate entries

# Description

This PR adds the option to "rename and append" using the `rename()` function, instead of only replacing values.

For example, having timeseries for `Primary Energy|Coal` and `Primary Energy|Gas`, the `rename()` function can now add the timeseries `Primary Energy` using

```
df.rename({'variable':
    {'Primary Energy|Coal': 'Primary Energy'
     'Primary Energy|Gas': 'Primary Energy'}},
    append=True)
```

This function also works for aggregating across regions.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- (#82)[https://github.com/IAMconsortium/pyam/pull/82] Adding the option to append (instead of replace) in `rename()`
 - (#81)[https://github.com/IAMconsortium/pyam/pull/81] Bug-fix when using `set_meta()` with unnamed pd.Series and no `name` kwarg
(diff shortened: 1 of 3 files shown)


gidden:
what's the status here?

danielhuppmann:
I didn't get it to work as expected yet, because redoing the renaming would potentially multiply the values. Like rename-and-append `a=1` and `b=1` to `c=2`, and then do it again might return `c=4`. Needs some more thinking on warning messages.

Still on my to-do list but with medium priority, and not required for the sr15 work.

danielhuppmann:
closing in favour of #178

PR #85, opened 2018-09-09
Filter by string with `nan` and better loading of metadata from file

### LINKED ISSUE #40: filtering by metadata column fails if comparing string and `nan`

### LINKED ISSUE #69: Loading metadata from file should filter down to available scenarios
Currently, loading a metadata file adds models and scenarios even if no data for them exists in the `IamDataFrame`. This should be downfiltered before combining (with appropriate logger messages).

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description

This PR fixes two issues: first, the function `load_metadata()` for importing from file now filters down to valid scenarios (included in the IamDataFrame) and includes meaningful logging.

Second, so far, there was a problem when using `filter()` using a string if the meta column had `np.nan`, because comparing string to `nan` doesn't work with `pd.Series.isin()` in combination with `regexp`.

Now, all `np.nan` are replaced by empty strings when applying a filter with a string query. This makes sense (sort of) because searching for any string (i.e, `*`) now returns the `nan` entry, and every other query does not. It also allows filtering for uncategorised entries by setting `df.filter(col='')`.

By extension, setting the meta columns by default to `uncategorized` is now removed - this was a quick fix for the issue now solved. Before, `pyam` tried to guess whether the meta values were strings or numbers - now, it always sets new meta columns to `np.nan`.

closes #40 
closes #69


danielhuppmann:
@gidden, any ideas what the problem of @stickler-ci is?

gidden:
looks great!

PR #88, opened 2018-09-12
Add region sum check

znicholls:
This PR adds a couple of simple methods that follow from our experience with the CMIP6 database. `check_aggregate_regions` checks that the world totals equal the regional sums, taking into account things like International Shipping which are only given on a world basis but have to be included when checking the sums. `check_internal_consistency` is a simple wrapper which just checks whether the database is internally consistent (regional sums add to world, sectoral sums add totals) and uses `check_aggregate` and `check_aggregate_regions`. 

@gidden @danielhuppmann it would be great if you could have a look at this at some point and let me know what else you'd like to change. I've added a new notebook to show you and future users what is going on (https://github.com/IAMconsortium/pyam/blob/3c124c25fdbc9fc74ccf712578f22b9283ca05a1/tutorial/checking_databases.ipynb). 

- [x] Tests Added
- [x] Documentation Added

# Adding to RELEASE_NOTES.md

```
```


znicholls:
Also happy to somehow move this into the `validate` method if that's better?

gidden:
I gave this a once-over, @danielhuppmann I *think* (don't remember) that you wrote the initial aggregation function. If so, can you please give this a thorough review?

pyam/core.py
```diff
@@ -499,7 +500,13 @@ def check_aggregate(self, variable, components=None, units=None,
             components = self.filter(variable='{}|*'.format(variable),
                                      level=0).variables()
 
-        # filter and groupby data, use `pd.Series.align` for machting index
+        if not len(components):
+            msg = '{} - cannot check aggregate because it has no components'
+            logger().info(msg.format(variable))
+
```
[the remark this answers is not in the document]
znicholls: good pick up. as an aside, have you thought of using a code formatter to handle all this stuff?

pyam/core.py
```diff
@@ -499,7 +500,13 @@ def check_aggregate(self, variable, components=None, units=None,
             components = self.filter(variable='{}|*'.format(variable),
                                      level=0).variables()
 
-        # filter and groupby data, use `pd.Series.align` for machting index
+        if not len(components):
```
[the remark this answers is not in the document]
znicholls: The reason I'd favour not raising an error is that you then have to wrap all your checks in try except blocks if you want to just throw a table which has variables with no components at it. If you just print a message and return None then things can keep running. Happy either way though and happy to put the try except in the `internal_consistency` method so the logic is kept separate.

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+
+            diff = pd.concat([diff], keys=[region], names=['region'])
+
+            return diff.unstack().rename_axis(None, axis=1)
+
+    def check_internal_consistency(self, **kwargs):
+        #TODO: test this and write docstring, kwargs passed to np is close
+        inconsistent_vars = {}
+        for variable in self.variables():
+            diff_agg = self.check_aggregate(
```
gidden: I think this call can be all one line. In general, we prefer to be as efficient as possible in vertical space unless using that space increases readability. Here and below, I think you're safe for a one-liner.

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+
+            diff_regional = self.check_aggregate_regions(
+                variable,
+                **kwargs
+            )
+            if diff_regional is not None:
+                inconsistent_vars[variable + "-regional"] = diff_regional
+
+        if inconsistent_vars:
+            return inconsistent_vars
```
gidden: Since you define it initially as an empty dict, I was just return it (and not check)

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
         if len(diff):
-            msg = '{} of {} data points are not aggregates of components'
-            logger().info(msg.format(len(diff), len(df_variable)))
+            msg = '{} - {} of {} data points are not aggregates of components'
+            logger().info(msg.format(variable, len(diff), len(df_variable)))
 
             if exclude_on_fail:
                 self._exclude_on_fail(diff.index.droplevel([2, 3]))
 
+            diff = pd.concat([diff], keys=[variable], names=['variable'])
```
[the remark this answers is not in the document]
znicholls: I just wanted to add the variable back onto the data table which is returned. For the wrapping function, you want to be able to see which variables failed (again, happy to move this wrapping to `internal_consistency` to separate the logic)

pyam/core.py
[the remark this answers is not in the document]
gidden: We use `stickler` which should pick up most things automatically. Perhaps my request here is not strictly PEP8, more of general style regarding vertical line efficiency (see comment below)

pyam/core.py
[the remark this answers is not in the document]
gidden: maybe @danielhuppmann has an opinion. I saw another function he wrote with a option like `fail_on_error` or something..

pyam/core.py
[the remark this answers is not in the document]
znicholls: ah ok. stickler isn't running on this for some reason...

pyam/core.py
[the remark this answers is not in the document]
gidden: so is this the same as diff.set_index('variable')?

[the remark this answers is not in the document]
gidden: In any case, if it's for a specific use case, it is probably best to keep this function the same and adjust where you need it elsewhere.

gidden:
also @znicholls regarding the failing tests on travis, I have one fix in there. If #91 passes, I will merge that in as well. Once that happens, you can rebase this to get the test fixes in.

pyam/core.py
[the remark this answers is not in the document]
gidden: looks like we're over our limit for reviews, or were the last time this was run...

znicholls:
> also @znicholls regarding the failing tests on travis, I have one fix in there. If #91 passes, I will merge that in as well. Once that happens, you can rebase this to get the test fixes in.

Sweet, stickler reviews can also be tested when that happens

pyam/core.py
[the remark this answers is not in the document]
znicholls: > so is this the same as `diff.set_index('variable')`?

It may well be (I don't know what I'm doing half the time with pandas), I'm just trying to add an extra row to the output index.

> In any case, if it's for a specific use case, it is probably best to keep this function the same and adjust where you need it elsewhere.

Cool so pending danielhuppmann's thoughts a lot will probably get moved into `check_internal_consistency`

tutorial/pyam_first_steps.ipynb
```diff
@@ -182,6 +182,31 @@
...
+    "Using keyword `keep=False` allows you to discard everything that is found by the filter rather than keeping it"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df.filter(region=\"World\", keep=False).data[['region']].drop_duplicates()"
```
danielhuppmann: There is a function `df.regions()` which does the same thing as `df.data[['region']].drop_duplicates()`

tutorial/pyam_first_steps.ipynb
```diff
@@ -182,6 +182,31 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df.filter(region=\"World\", keep=True).data[['region']].drop_duplicates()"
```
danielhuppmann: Use `df.regions()`, see previous comment

tests/conftest.py
```diff
@@ -31,6 +31,91 @@
 )
 
 
+CHECK_AGG_DF = pd.DataFrame([
```
danielhuppmann: This looks to me like a much larger dataframe than what would be needed to run the unit tests...?

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+            variable to be checked for matching aggregation of regional data
+        region: str
+            region to be checked for matching aggregation of regional data
+        components: list of str, default None
+            list of regions, defaults to all regions except region
+        units: str or list of str, default None
+            filter variable and components for given unit(s)
+        exclude_on_fail: boolean, default False
+            flag scenarios failing validation as `exclude: True`
+        multiplier: number, default 1
```
danielhuppmann: The multiplier is sometimes useful when comparing across variables when the variables are naturally in different units (e.g., carbon vs. carbon dioxide, EJ vs. GWa). I don't think that there is a use case over regions.

pyam/core.py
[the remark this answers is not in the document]
danielhuppmann: I'd rather see everything that is internally inconsistent at once rather than get an error message every time the internal-consistency finds an issue, and I have to fix it and run the function again

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+
+        # filter and groupby data, use `pd.Series.align` for matching index
+        df_region, df_components = (
+            _aggregate_by_regions(var_df.data, region, units)
+            .align(_aggregate_by_regions(var_df.data, components, units))
+        )
+
+        # if it's a World comparison, we need to add in variables that might
+        # have been missed e.g. shipping and aviation
+        if region == "World":
```
danielhuppmann: I don't think it's a good idea to hard-code any region name, not even ' World' - and I'm not sure we even need this. If we first make sure that all variables aggregate by region, and then check that within each region, they add up to the "parent" variable, the variable only reported at the global level would naturally be included there (and not at the regional level).

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+            logger().info(msg.format(variable, len(diff), len(df_region)))
+
+            if exclude_on_fail:
+                self._exclude_on_fail(diff.index.droplevel([2, 3]))
+
+            diff = pd.concat([diff], keys=[region], names=['region'])
+
+            return diff.unstack().rename_axis(None, axis=1)
+
+    def check_internal_consistency(self, **kwargs):
```
danielhuppmann: from my experience with the SR1.5 database, there are so many variables that don't aggregate up "naturally", like `Final Energy|Gas` and `Final Energy|Transport`, that this function would be pretty much useless in general applications. I'd rather pass over a list of variables (with mapping to components were necessary) that should be checked rather than assuming a blanket "check everything".

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+        units: str or list of str, default None
+            filter variable and components for given unit(s)
+        exclude_on_fail: boolean, default False
+            flag scenarios failing validation as `exclude: True`
+        multiplier: number, default 1
+            factor when comparing variable and sum of components
+        kwargs: passed to `np.isclose()`
+        """
+        # default components to all variables one level below `variable`
+        var_df = self.filter(variable=variable, level=0)
```
danielhuppmann: this does not select all variables "one level below `variable`"

pyam/core.py
```diff
@@ -510,14 +517,125 @@ def check_aggregate(self, variable, components=None, units=None,
...
+
+        if components is None:
+            components = var_df.filter(region=region, keep=False).regions()
+
+        if not len(components):
```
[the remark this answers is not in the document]
danielhuppmann: I'd rather not throw errors for the first inconsistency you find...

tutorial/pyam_first_steps.ipynb
znicholls: oops good pickup

tests/conftest.py
znicholls: ye so I wasn't completely sure how best to do this. Basically I wanted a fake dataset where I had a variable with components, a variable with no components, a variable like CO2 which has data at the world level which needs to go in the world total but isn't in the regions and at least 2 regions (to check the aggregate against the world) so seemed like I need this...

pyam/core.py
znicholls: Nice, will adjust the logic

pyam/core.py
znicholls: That may be the case for energy but for an emissions user (like us), these aggregations have to be the case. Maybe this should take a filter keyword so it could just look at variables that start with 'Emissions'?

gidden:
Hey @znicholls and @danielhuppmann what is our status here? @znicholls we're hoping to package up a release soon. This would be a great addition!

pyam/core.py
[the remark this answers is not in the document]
znicholls: ok so I'll leave it then

pyam/core.py
```diff
@@ -499,7 +500,13 @@ def check_aggregate(self, variable, components=None, units=None,
             components = self.filter(variable='{}|*'.format(variable),
                                      level=0).variables()
 
-        # filter and groupby data, use `pd.Series.align` for machting index
+        if not len(components):
+            msg = '{} - cannot check aggregate because it has no components'
+            logger().info(msg.format(variable))
+
+            return None
```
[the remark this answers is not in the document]
znicholls: given the above, will change to just `return`

pyam/core.py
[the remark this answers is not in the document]
znicholls: > so is this the same as `diff.set_index('variable')`?

I think it would be except that `diff` doesn't have a `variable` column at this point so we can't use `set_index` and instead have to do this `pd.concat`.

> In any case, if it's for a specific use case, it is probably best to keep this function the same and adjust where you need it elsewhere.

My opinion is still that it would be better to actually see the thing you're checking consistency for in the output but will leave resolution up to @danielhuppmann

pyam/core.py
znicholls: Having re-thought this, here's my use case where it's needed. You want to check that World, Emissions|CO2 is correct. Hence you just add up all the regions for Emissions|CO2. You find that the regional sum is less than World because you're missing variables. I don't see how else to do this ('World' is unique in this respect I think). 

One option would be to make `check_aggregate_regions` a private method which is only used as part of `check_internal_consistency`, in which case your logic would always hold.

znicholls:
Have gone through and answered all the questions I can. Will await @danielhuppmann's thoughts then can update and tidy up.

pyam/core.py
```diff
@@ -510,14 +517,115 @@ def check_aggregate(self, variable, components=None, units=None,
...
+            region to be checked for matching aggregation of regional data
+        components: list of str, default None
+            list of regions, defaults to all regions except region
+        units: str or list of str, default None
+            filter variable and components for given unit(s)
+        exclude_on_fail: boolean, default False
+            flag scenarios failing validation as `exclude: True`
+        kwargs: passed to `np.isclose()`
+        """
+
```
danielhuppmann: remove empty line here

pyam/core.py
danielhuppmann: thanks for setting `World` via a kwarg

danielhuppmann: yes, sorry, I misunderstood this piece of code, I though you had changed this... I think it would be preferably to explicitly state which variables need to be added at the aggregate level (`region='World'`) to get to the sum of components...

znicholls: Ah ok. I did think about doing it that way. The reason I didn't is that it's often hard to predict that a priori. For example, some models have a regional breakdown of aviation (so you don't need to add it again) whilst others don't (so you do need to add aviation in when checking the world total), similarly for aggricultural emissions for CO2. 

Doing it this way ensures that any variables which are in WORLD, but don't have a regional breakdown provided, get included when we check the WORLD total variable. 

I know it's nasty so am happy to clear it up however you think is best.

znicholls:
@gidden @danielhuppmann happy to add docs etc. to this today, we just need to resolve the last open conversation

pyam/core.py
danielhuppmann: but I'd still think that you could make that logic part of a keyword argument - if you check for a particular region and the subregions (if provided, default all other regions) don't have a particular variable, you include it from the region - otherwise, you aggregate up.

one could think of having World, continents and countries, where some variables are only reported at the continental level. So you would first check that for each continent, internal consistency is given, and then one checks internal consistency from continents to global totals...

znicholls:
pending tests, this is good to go from my end

danielhuppmann:
thanks @znicholls for a really nice new feature!

@gidden, can you take a final look and merge the PR?

gidden:
This is awesome, and even has a tutorial! Thanks a bunch, @znicholls

gidden:
waiting for CI then will merge

znicholls:
> This is awesome

Nah pyam is awesome, this is just an add on!

PR #90, opened 2018-09-17
Panel label

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description

This PR adds a function to the plotting library to set a "panel label", which is by default placed in the top-left corner of an Axes object.


gidden:
failure could be because of the freetype version, but not sure.

danielhuppmann:
I think that failure is due to wrong matplotlib version (2.1.2 on CI, 2.2.2 locally for generating the test figure) - but 2.1.2 doesn't run on Mac for some framework problem. @gidden, can you generate the test figure and PR into my branch?

gidden:
done

tests/test_plotting.py
```diff
@@ -388,3 +388,11 @@ def test_stack_plot_other(plot_df):
          .stack_plot(ax=ax, stack='scenario', cmap='viridis', title='foo')
          )
     return fig
+
+
+@pytest.mark.skipif(IS_WINDOWS, reason=WINDOWS_REASON)
```
gidden: this needs to be deleted now. apologies, I was thrashing in CI and removed this from other functions.

gidden:
one final bit, then g2g

danielhuppmann:
good to go, @gidden!

danielhuppmann:
now with release notes...

PR #93, opened 2018-09-18
initialise IamDataFrame with index

### LINKED ISSUE #77: IamDataFrame constructor should take pd.DataFrame with index
Currently, the `IamDataFrame.__init__()` can take a `pd.DataFrame` with columns `['model', 'scenario' , ...]`, but it fails if these identifiers are given as an index.

This would be a nice-to-have feature so that the `pd.DataFrame` returned by the `timeseries()` function can be turned back into an `IamDataFrame`. This currently requires a `reset_index()`.

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR extends the IamDataFrame constructor to take a pd.DataFrame with a meaningful index, for example the dataframe returned by `IamDataFrame.timeseries()`. Closes #77

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -169,6 +169,10 @@ def format_data(df):
             df['scenario'] = scen.apply(
                 lambda s: '-'.join(s.split('-')[1:]).strip())
 
+    # reset the index if meaningful entries are included there
+    if not list(df.index.names) == [None]:
+        df.reset_index(inplace=True)
+
     # format columns to lower-case and check that all required columns exist
     if not set(IAMC_IDX).issubset(set(df.columns)):
         missing = list(set(IAMC_IDX) - set(df.columns))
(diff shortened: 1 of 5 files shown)


gidden:
is this good to go?

pyam/utils.py
```diff
@@ -169,6 +169,10 @@ def format_data(df):
             df['scenario'] = scen.apply(
                 lambda s: '-'.join(s.split('-')[1:]).strip())
 
+    # reset the index if meaningful entries are included there
+    if not list(df.index.names) == [None]:
```
gidden: here I think you could do just `if df.index.names:`

danielhuppmann: just tried this, seems that this returns `True` even if no names are given

danielhuppmann:
thanks @gidden, good to go after the tests pass!

gidden:
Ah, of course, it has an element. Oh well.

PR #94, opened 2018-09-20
set meta by pd.DataFrame

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

In the existing implementation of `set_meta()`, passing over a non-valid index (e.g., a pd.DataFrame) was simply ignored without warning or error. This PR is a clean-up in the sense that if there is an `index` arg, it will be used and an error raised if it cannot be coerced in a meaningful way.

In addition, the function now also takes a `pd.DataFrame` with columns `['model', 'scenario']` as valid `index` arg.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -297,6 +297,10 @@ class IamDataFrame(object):
         if not isinstance(index, pd.MultiIndex):
             raise ValueError('index cannot be coerced to pd.MultiIndex')
 
+        # raise error if index is not unique
+        if index.duplicated().any():
+            raise ValueError("non-unique ['model', 'scenario'] index!")
+
         # create pd.Series from meta, index and name if provided
         meta = pd.Series(data=meta, index=index, name=name)
         meta.name = name = name or meta.name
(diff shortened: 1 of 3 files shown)


gidden:
Ok, I think I see the point here, but one question. The test as is has the following lines

```
    df = pd.DataFrame([
        ['a_model', 'a_scenario', 'a_region1', 1],
    ], columns=['model', 'scenario', 'region', 'col'])
     meta_df.set_meta(meta=0.3, name='meta_values', index=df)
```

Would the existing implementation work if the user provided the following?

```
    df = pd.DataFrame([
        ['a_model', 'a_scenario', 'a_region1', 1],
    ], columns=['model', 'scenario', 'region', 'col'])
     meta_df.set_meta(meta=0.3, name='meta_values', index=df[['model', 'scenario']])
```

pyam/core.py
```diff
@@ -272,53 +272,55 @@ def set_meta(self, meta, name=None, index=None):
...
 
-        # use meta.index if index arg is an IamDataFrame
-        if index is not None and isinstance(index, IamDataFrame):
-            index = index.meta.index
-
-        # create pd.Series from meta, index and name if provided
-        _meta = meta if index is None else pd.Series(data=meta, index=index,
-                                                     name=name)
+        # check if meta has a valid index and use it for further workflow
+        if hasattr(meta, 'index') and hasattr(meta.index, 'names') \
```
gidden: do we need to check if `index is None` here?

gidden:
Ok, answered my own question by looking back through. Looks great to me!

Issue #51, opened 2018-03-19
`as_pandas()` returns bigger-than-necessary dataframes


danielhuppmann:
When the function `line_plot()` calls `as_pandas()` to retrieve the joined data+meta dataframe and then plots two metadata columns, it paints each marker once per timestep (because the rows in the joined dataframe are duplicated if you ignore year and value).

It might make more sense to pass the required columns to `as_pandas()` and then let it decide whether to join data and meta, or return data or meta only.

danielhuppmann:
this has actually been fixed in `line_plot()` but the issue persists in `stack_plot()`, `bar_plot()` and `pie_plot()`...

PR #95, opened 2018-09-23
add a scatter plot to the plotting library

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a function to draw a scatter plot using values from two columns of the `meta` dataframe.


gidden:
looks like issues with freetype again. let me see if I can figure out a way to update the linux ci...

danielhuppmann:
thanks @gidden!

gidden:
I played around in #97, but it seems like no matter what I do, matplotlib cannot find the updated freetype version. E.g.,

```
$ freetype-config --ftversion
2.8.1
```

```
$ pytest --mpl
============================================================================== test session starts ===============================================================================
platform linux -- Python 3.5.2, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
Matplotlib: 2.2.0
Freetype: 2.6.1
rootdir: /home/gidden/work/iiasa/pyam, inifile:
plugins: mpl-0.9
```

gidden:
I can generate the image and PR as last time, but we should either dig into this issue further in the future or I can show you how to do this with docker containers =)

gidden:
expected fig as generated on linux

test_scatter

pyam/core.py
```diff
@@ -745,6 +745,13 @@ def pie_plot(self, *args, **kwargs):
         ax = plotting.pie_plot(df, *args, **kwargs)
         return ax
 
+    def scatter(self, *args, **kwargs):
+        """Plot a scatter chart using metadata columns
+
+        see pyam.plotting.scatter() for all available options
+        """
+        return plotting.scatter(self.meta.reset_index(), *args, **kwargs)
```
gidden: So this seems inconsistent with our other plotting functions. why is this `meta` by default? shouldn't we also be able to plot scatter data from the `data` df?

danielhuppmann: The way I implemented (and used) it so far is this works for `meta` columns only. In the the unit test, I take some data and convert it to meta columns, then display those.

Assume we want to create a scatter plot between two variables over time - how would you imagine to pass over the `x` and `y` args?

danielhuppmann:
Summary from offline discussion with @gidden for allowing `scatter()` not just to use `meta` columns: add a kwarg `dim` , default to `dim='meta'`, which will allow to specify where to expect the `x` and `y` content (e.g, variable, region, scenario). This still assumes that `x` and `y` are coming from the same dimension (for now)...

gidden:
Did we figure out the issue with use `as_pandas()`? If possible, it might be nicer to have the fully merged dataset with an interface similar to `bar_plot()` like

```
def scatter(x, y, col=None)
```

With

```
x: column name or element name within a column
y: column name or element name within a column
col: if x or y is given as element name, select elements from this column
```

then you could do things like

```
df.scatter('year', 'variable')
df.filter('variable'='Primary Energy').scatter('year', 'variable')
df.scatter('Primary Energy', 'GDP', col='variable')
```

danielhuppmann:
The over-large `as_pandas()` issue (#51) is not resolved. I attempted it, but this is not as simple as I had thought, or my approach was not elegant enough.

I used the present implementation of `scatter()` in the SR1.5 finalisation and it works as expected.

Suggest to merge this PR as is (unless there are specific code improvements) to include it in the first release and move the proposal for a better API to a new issue.

gidden:
@danielhuppmann can you give me an example dataset? I would be happy to try to implement it. Fundamentally, I think scatter plots should be able to plot variable data.

danielhuppmann:
you could use the data from the tutorial, see 
`https://github.com/IAMconsortium/pyam/blob/master/tutorial/tutorial_AR5_data.csv`

gidden:
closing in favor of #108

PR #101, opened 2018-10-01
add a function to retrieve years in which a series crosses a threshold

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This PR adds a `pyam.cross_threshold()` function to retrieve a list of years in which a `pd.Series` (e.g., a timeseries row) crosses a given threshold.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- (#101)[https://github.com/IAMconsortium/pyam/pull/101] add function `cross_threshold()` to determine years where a timeseries crosses a given threshold
 - (#94)[https://github.com/IAMconsortium/pyam/pull/94] `set_meta()` can take pd.DataFrame (with columns `['model', 'scenario']`) as `index` arg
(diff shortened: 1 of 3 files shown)


gidden:
this looks good @danielhuppmann, but it's a little confusing because the crossing could be in the positive (below to above) or negative (above to below). I think that information should be included in what is returned.

Also, maybe add a test for one that does not cross a threshold?

danielhuppmann:
thanks @gidden!

Added a test where `cross_threshold()` returns an empty list (because not crossing the threshold).

On returning the direction of crossing, not sure what would be suitable return format - any ideas? Alternatively, the function could take an additional optional `from=['above', 'below']` kwarg that filters the returned list.

gidden:
Yes, I would add a kwarg like `direction`.

danielhuppmann:
I was thinking of `direction` at first, too, but how would you then name the options? `direction=up` is ambiguous. That's why I thought `from` would be more intuitive...

danielhuppmann:
@gidden, added a kwarg `direction=['from above', 'from below']` and three more tests for that feature...

Should be good to go once the tests pass

PR #103, opened 2018-10-09
fix bug when timeseries columns are formatted as float

### LINKED ISSUE #104: Add checks that years have to be integers
PR #103 fixes a bug of the `cumulative()` and `fill_series()` functions when timeseries have columns that are defined as floats instead of integers.

This would be great to also check in the `IamDataFrame.data.year` column when initializing a new instance, renaming or appending data.

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR fixes a bug with the timeseries functions `fill_series()` and `cumulative()` when the columns (years) are defined as floats instead of integers. An auxiliary function is introduced to cast columns to `int` and raise an error if non-meaningful years are present (like `2020.5`).

It also casts the `year` column in new instances of an `IamDataFrame` as `int`, closes #104.

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -262,3 +262,17 @@ def years_match(data, years):
     """
     years = [years] if isinstance(years, int) else years
     return data.isin(years)
+
+
+def cast_years_to_int(x, index=False):
+    """formatting series or timeseries columns to int and checking validity"""
+    _x = x.index if index else x
+    cols = list(map(int, _x))
+    error = _x[cols != _x]
+    if not error.empty:
+        raise ValueError('invalid values `{}`'.format(list(error)))
+    if index:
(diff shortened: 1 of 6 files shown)


danielhuppmann:
stickler happy on first try, hurrray!

pyam/utils.py
```diff
@@ -262,3 +262,17 @@ def years_match(data, years):
     """
     years = [years] if isinstance(years, int) else years
     return data.isin(years)
+
+
+def cast_years_to_int(x, index=False):
+    """formatting series or timeseries columns to int and checking validity"""
```
gidden: can we get some docs on what `index` does here?

gidden:
lgtm, just curious about the `index` docs

PR #112, opened 2018-10-16
Proposal for notebook

danielhuppmann:
Inspired by https://www.ipcc.ch/publications_and_data/ar4/wg1/en/figure-spm-5.html


gidden:
building a new pr off this commit

PR #114, opened 2018-10-19
extend `append()` function to allow adding data

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR extends the `append()` function to allow adding data for existing scenarios in an IamDataFrame. Until now, the function only accepted data for model-scenarios that were not in the index of the IamDataFrame to which data was appended.

In practice, I often wanted to add data to an existing scenario, like pull out GDP and population data, compute a timeseries GDP per capita, and append it. I the implemented this by 
```
df.data = df.data.append(other.data)
```
The drawback here is that this creates duplicated timeseries data in the IamDataFrame if one executes this line more than once.

Now, the function accepts new timeseries data as long as there is no overlap with existing data.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -179,8 +179,16 @@ class IamDataFrame(object):
             # if not ignored, check that overlapping meta dataframes are equal
             if not ignore_meta_conflict:
                 cols = [i for i in other.meta.columns if i in ret.meta.columns]
-                pd.testing.assert_frame_equal(ret.meta.loc[intersect, cols],
-                                              other.meta.loc[intersect, cols])
+                if not ret.meta.loc[intersect, cols].equals(
+                        other.meta.loc[intersect, cols]):
+                    conflict_idx = (
+                        pd.concat([ret.meta.loc[intersect, cols],
+                                   other.meta.loc[intersect, cols]]
+                                  ).drop_duplicates().index.drop_duplicates()
+                        )
+                    msg = 'conflict in `meta` for scenarios {}'.format(
+                        [i for i in pd.DataFrame(index=conflict_idx).index])
+                    raise ValueError(msg)
 
             cols = [i for i in other.meta.columns if i not in ret.meta.columns]
             _meta = other.meta.loc[intersect, cols]
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -184,8 +184,9 @@ class IamDataFrame(object):
                     conflict_idx = (
                         pd.concat([ret.meta.loc[intersect, cols],
                                    other.meta.loc[intersect, cols]]
-                                  ).drop_duplicates().index.drop_duplicates()
-                        )
+                                  ).drop_duplicates()
+                        .index.drop_duplicates()
+                                   )
                     msg = 'conflict in `meta` for scenarios {}'.format(
                         [i for i in pd.DataFrame(index=conflict_idx).index])
                     raise ValueError(msg)
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -169,16 +169,19 @@ class IamDataFrame(object):
         diff = other.meta.index.difference(ret.meta.index)
         intersect = other.meta.index.intersection(ret.meta.index)
 
-        # join other.meta for new scenarios
-        if not diff.empty:
-            ret.meta = ret.meta.append(other.meta.loc[diff, :], sort=False)
-
         # merge other.meta columns not in self.meta for existing scenarios
         if not intersect.empty:
-            _meta = other.meta.loc[intersect, [i for i in other.meta.columns
-                                               if i not in ret.meta.columns]]
+            cols = [i for i in other.meta.columns if i not in ret.meta.columns]
+            _meta = other.meta.loc[intersect, cols]
             ret.meta = ret.meta.merge(_meta, how='outer',
                                       left_index=True, right_index=True)
+
+        # join other.meta for new scenarios
+        if not diff.empty:
+            # sorting not supported by ` pd.append()`  prior to version 23
+            sort_kwarg = {} if int(pd.__version__.split('.')[1]) < 23 \
+                else dict(sort=False)
+            ret.meta = ret.meta.append(other.meta.loc[diff, :], **sort_kwarg)
 
         # append other.data (verify integrity for no duplicates)
         ret.data.set_index(LONG_IDX, inplace=True)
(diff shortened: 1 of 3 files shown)


danielhuppmann:
@gidden, an API question for you: assume I have two IamDataFrames with the same model/scenario and conflicting meta column entries, like

``data.meta``

|    |    |exclude| col1 |
|--|---|------|------|
|m | s | False | a       |

``other.meta``

|    |    |exclude| col1 | col2 |
|--|---|------|------|------|
|m | s | False | b       | 1       |

The way I implemented it now, the values in `meta` are taken from `data.meta` wherever possible, and only columns that don't exist in `data.meta` are appended from `other.meta`.

``data.append(other).meta``

|    |    |exclude| col1 | col2 |
|--|---|------|------|------|
|m | s | False | a       | 1       |

That means the `other.meta.col1 = b` is simply ignored in `append()`. Not sure if that is the ideal behaviour here - any ideas?

danielhuppmann:
@gidden, what is the `pandas` version on CI? Seems that it doesn't know `pd.append(sort=False)`...

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

gidden:
It looks like sort is "New in version 0.23.0.". The question now is whether we want to support backwards pandas dependency. I would suggest one of two options:

- check pandas version here with `_kwargs = dict(sort=True) if pands.__version__[1] >= 23 else {}`
- check pandas version on import with 
```
import pandas
if pandas.__version__[1] < 23:
    raise ImportError('Pandas version must be >= 0.23.0')
```

We can discuss which is best.

danielhuppmann:
GitHub doesn't want to run the unit tests

gidden:
Hmm...

gidden:
try `git commit --amend && git push -f origin ...`

danielhuppmann:
seems that the GitHub PR page is lagging behind the GitHub fork page... last commit on the branch there is https://github.com/danielhuppmann/pyam/commit/a23ae3916f0c30d1571e5b6d4eb82155f8aca1ea, but this isn't updated on this page

danielhuppmann:
updated now, but GitHub seems to have forgotten that it wants to run some tests...

pyam/core.py
```diff
@@ -147,28 +147,58 @@ def variables(self, include_units=False):
...
+        intersect = other.meta.index.intersection(ret.meta.index)
 
-        # check that any model/scenario is not yet included in IamDataFrame
-        ret.meta = ret.meta.append(other.meta, verify_integrity=True)
+        # merge other.meta columns not in self.meta for existing scenarios
+        if not intersect.empty:
+            # if not ignored, check that overlapping meta dataframes are equal
+            if not ignore_meta_conflict:
+                cols = [i for i in other.meta.columns if i in ret.meta.columns]
+                pd.testing.assert_frame_equal(ret.meta.loc[intersect, cols],
```
gidden: I would place this in a `try/except` so that we can provide the user with a specific error message. and/or use [`df.equals()` ](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.equals.html)

pyam/core.py
```diff
@@ -147,28 +147,58 @@ def variables(self, include_units=False):
...
             other = IamDataFrame(other, **kwargs)
+            ignore_meta_conflict = True
+
+        diff = other.meta.index.difference(ret.meta.index)
+        intersect = other.meta.index.intersection(ret.meta.index)
 
-        # check that any model/scenario is not yet included in IamDataFrame
-        ret.meta = ret.meta.append(other.meta, verify_integrity=True)
+        # merge other.meta columns not in self.meta for existing scenarios
+        if not intersect.empty:
```
gidden: this can be a single line `if not intersect.empty and not ignore_meta_conflict`

gidden:
looking good, just a few suggested changes

pyam/core.py
danielhuppmann: note quite the same: The way it's implemented now, columns that are only in `other` are always appended to `self.meta`... In addition, there is an option to check that overlapping columns must be equal.

pyam/core.py
danielhuppmann: changed to using `df.equals` and a specific error message with a list of model-scenarios that have conflicts in `meta`

pyam/core.py
gidden: Ah i see, more indentation underneathe

PR #117, opened 2018-10-23
Docs

gidden:
After this, I think we are ready to go.


gidden:
If I have time, I'll add a test for the bars..

PR #122, opened 2018-10-24
Split out base class

znicholls:

- [x] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

This splits out the key, reusable components of `pyam.IamDataFrame` into a subclass, `pyam._PyamDataFrame` which can then be superclassed for other users. The only reason I haven't moved all the methods is because some clearly don't apply to climate data (e.g. `check_internal_consistency` would have to be re-written to somehow know that you shouldn't add `Atmospheric Concentrations|CO2` and `Atmospheric Concentrations|CH4` but should add `Radiative Forcing|CO2` and `Radiative Forcing|CH4`).

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -39,11 +39,9 @@ from pyam.utils import (
 from pyam.timeseries import fill_series
 
 
-class IamDataFrame(object):
-    """This class is a wrapper for dataframes following the IAMC format.
-    It provides a number of diagnostic features (including validation of data,
-    completeness of variables provided) as well as a number of visualization
-    and plotting tools.
+class _PyamDataFrame(object):
+    """Base class for pyam dataframes.
+    This class provides a number of data analysis tools, in particular data categorisation, filtering and visualistation tools
     """
 
     def __init__(self, data, **kwargs):
@@ -68,10 +66,6 @@ class IamDataFrame(object):
             self.data = read_ix(data, **kwargs)
         else:
             self.data = read_files(data, **kwargs)
-
-        # cast year column to `int` if necessary
-        if not self.data.year.dtype == 'int64':
-            self.data.year = cast_years_to_int(self.data.year)
 
         # define a dataframe for categorization and other metadata indicators
         self.meta = self.data[META_IDX].drop_duplicates().set_index(META_IDX)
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Looking at this PR, I'm still not convinced why we need to create a generic subclass.

The example about `check_internal_consistency()` requiring additional info is equally valid in IAMs, where you often have variable trees like
```
Primary Energy|Fossil
Primary Energy|Coal
Primary Energy|Non-Biomass Renewables
Primary Energy|Hydro
```
[yes, one could create maybe strictly hierarchical trees, but this has been met with quite some push-back in the community]

So what I think would be required for `check_internal_consistency()` to be useful in any applied project is a project-specific mapping (file) that specifies the aggregation tree or other rules (sum, weighted average, etc.).

[as an aside, I talked to @tscheypidi a while back to make this mapping file readable by `pyam` and PIK's R `iamc` package]

znicholls:
> I'm still not convinced why we need to create a generic subclass

I'm not convinced either. However, the advantages I see right now are:

1. whilst we don't have the ability to add extra meta columns, this provides a simple workaround which recognises the different conventions in the IAM and SCM communities
1. once you have a base class (and an easy way to test its subclasses, which is introduced in #124), it is easy to introduce more subclasses with whatever custom functionality you need. This can be useful in future and given that this plus #124 would do it, the hardest step is done and so not using this ability seems a shame.

> useful in any applied project is a project-specific mapping

This would be a good reason to put that function into the base class then. My question would be, do we have one universal mapping file which knows about all variables (probably?), plan b could be that the subclasses know the mappings relevant to their communities?

gidden:
@znicholls, if you think this is the best way to implement the climate modeling, then I'm ok with it.

I would suggest, though, making two new modules, e.g., `iam.py` and `scm.py` and putting the respective dataframes there. `core.py` is already getting pretty long. Plus, since there will be structural differences in the two indicies, it is probably best to keep their definitions separate.

znicholls:
In hindsight a better way may become more obvious but I think this is the right first step.

Good suggestion re splitting, will do!

znicholls:
I'm assuming that maintaining Python2 compatibility is essential?

gidden:
Yup

znicholls:
superceeded by #134

PR #127, opened 2018-10-26
Add variable infilling

znicholls:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This PR adds a `add_missing_variable` method to `_PyamDataFrame` to allow users to easily add variables which aren't in their dataframes. The variables can simply be added as zeros, or made to follow the lead trajectory of another gas (a poor man's version of Aneris). 

This functionality is useful if your climate model requires a variable that isn't produced in your scenario. It's basically a cheat way of saying, fill with zeros before you pass to your climate model (rather than hoping that the climate model does something sensible with missing data).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,13 +1,14 @@
 
 # Next Release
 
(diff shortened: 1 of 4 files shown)


znicholls:
If implemented, will go in OpenSCM

PR #134, opened 2018-11-02
Split out base class

znicholls:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Splits out a base class to make room for adding other dataframes, on top of `IamDataFrame`, for other communities.

I have left the check functions in `IamDataFrame` because we know how they should work for the IAM community, but other communities will have to think harder. I think it makes more sense to leave them there to start, then merge them into the base class later (as we work out the best way), rather than include them and have them throw all sorts of cryptic errors for users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 *xlsx
 *png
 
+# logo made when running tests
+doc/source/tutorials/logo.pdf
+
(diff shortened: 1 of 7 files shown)


znicholls:
if implemented, will go in OpenSCM

PR #139, opened 2018-11-19
Paper

gidden:
let me know what you think @danielhuppmann


danielhuppmann:
would it make sense to link `pyam` to 
https://doi.org/10.5281/zenodo.1470400 instead of the doi of release v0.1.0?

PR #143, opened 2018-11-19
Editorial suggestions for paper draft

danielhuppmann:
This PR suggests some changes to the paper draft.

### DIFF
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -64,3 +64,11 @@
 	 year={2018},
 	 month={Oct}
 }
+
+@misc{huppmann_notebooks_2018,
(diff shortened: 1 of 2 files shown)


gidden:
Hey @danielhuppmann looks good. The only reason I didn't also include the notebook ref before is that I don't know how self-referential adverse JOSS is, so tried to minimize... I have to run to RBP, but will pull/review after

danielhuppmann:
Agree that we don’t want to cite too much of our own work - but if that’s an issue, I’d rather remove the NCC reference.

gidden:
Ok, I'll go over it once more, thanks!

PR #145, opened 2018-11-27
updating for versioneer

gidden:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This PR uses [`versioneer`](https://github.com/warner/python-versioneer) to automatically generate appropriate version numbers. If on a tag, it will provide the current tag as the version (for packaging). If on a certain commit, it will name the commit like so:

```
$ python -c 'import pyam; print(pyam.__version__)'
0.1.2+6.g04b4940
```


setup.cfg
```diff
@@ -1,2 +1,9 @@
 [bdist_wheel]
 universal=1
+
+[versioneer]
+VCS = git
```
znicholls: might want to add `style = pep440` here? Not sure what exactly it means but it seems to be in my other projects...

setup.py
```diff
@@ -23,27 +23,6 @@
...
-        ]
-        for d in dirs:
-            if os.path.exists(d):
-                print('removing {}'.format(d))
-                shutil.rmtree(d)
-
-
 # thank you https://stormpath.com/blog/building-simple-cli-interfaces-in-python
 class RunTests(Command):
     """Run all tests."""
```
znicholls: Is this class redundant now that you're using `"cmdclass": versioneer.get_cmdclass(),` below? Looking at https://stormpath.com/blog/building-simple-cli-interfaces-in-python again, I think it needs to be setup something like https://github.com/znicholls/netcdf-scm/blob/master/setup.py for this to actually do anything.

setup.cfg
gidden: pep440 is the default style, we can add it or not... doesn't really matter. it simply means that the version you get is pep440 compliant.

setup.py
gidden: Yes, it is. I wasn't sure what exactly to do with it, since I didn't add it (I believe you did, right?). 

In any case, I can mirror your netcdf-scm style or we can just remove it. I always run `pytest tests` and not `python setup.py test`. Do you have a preference?

znicholls:
Pep440, toss a coin

Command class, I think copy netcdfscm. I think you have to have the command
class in there and this seems to be the way to make it work..

gidden:
ok should be g2g @znicholls

znicholls:
lgtm

PR #148, opened 2018-12-05
minor typo

Chilipp:
Fixed typo in first steps tutorial notebook.

Concerns software review in https://github.com/openjournals/joss-reviews/issues/1095


gidden:
Hi @Chilipp this is actually intentional. In the text above, we state that an empty array will be returned without the wildcard. Wildcards aren't necessarily well known by all people using these tools (many are newish to some programming topics), so we want to make sure they see this before potentially getting frustrated.

Chilipp:
All right, thank you 👍

PR #153, opened 2018-12-06
trying coveralls

gidden:


gidden:
ok, let's merge this then I can pr the badge in once it hits master

PR #171, opened 2019-01-11
compatibility with `ixmp`

danielhuppmann:

- [x] Tests Added - Not applicable, no unit tests for ixmp exist yet
- [x] Documentation Added - Only internal refactoring and bugfix

# Description of PR

This PR fixes a bug when loading timeseries data from an `ixmp.Scenario` object. Also, it refactors the use of the `ixmp` class, moving to an own file. This mitigates a potential circular dependency and allows `ixmp` and `message_ix` to depend on `pyam.utils` for some features. Note that `ixmp` is not intended as a dependency, only as an optional package.


gidden:
lgtm! we can discuss other options later as this is just moves a function around

PR #174, opened 2019-01-22
add a function `compare()` to compare two IamDataFrames

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR introduces a function `compare()` to compare the `data` tables of two IamDataFrames using the `np.isclose()` function.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 # Next Release
 
 - [#175](https://github.com/IAMconsortium/pyam/pull/175) Update link to tutorial in readme.md 
+- [#174](https://github.com/IAMconsortium/pyam/pull/174) Add a function `difference()` to compare two IamDataFrames
 - [#171](https://github.com/IAMconsortium/pyam/pull/171) Fix a bug when reading from an `ixmp.TimeSeries` object, refactor to mitigate circular dependency
 - [#162](https://github.com/IAMconsortium/pyam/pull/162) Add a function to sum and append timeseries components to an aggregate variable
 - [#152](https://github.com/IAMconsortium/pyam/pull/152) Fix bug where scatter plots did not work with property metadata when using two variables (#136, #152)
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1418,3 +1418,23 @@ def filter_by_meta(data, df, join_meta=False, **kwargs):
         data.index.name = None
 
     return data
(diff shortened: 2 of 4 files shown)


gidden:
lgtm, one suggestion though: should we add an additional option(s)? Specifically, I'm thinking difference, relative difference, and anything else? so you would see the difference in the test is `(2 - 1) = 1` and `(2 - 1) / 2 = 0.5`?

danielhuppmann:
The comparison is using `np.isclose()` and you can pass the kwargs to it through the `compare()` function. So absolute and relative tolerance can be specified, though you have to keep in mind that `np.isclose()` is not symmetric, so `np.isclose(a, b)` does not necessary equal `np.isclose(b, a)`. Still, this is good enough for my use case...

@gidden, can you specify a use case that requires a more elaborate treatment?

gidden:
I see that `np.isclose` is used. I was suggesting two separate things I suppose:

1. the definition of close could be in absolute magnitude or relative (within 50%, for example)
2. it is likely useful to see what the difference is, thus adding a column which provides `a - b` in addition to `a` and `b`

Another thought is this API/feel is close to looking like [`pd.subtract`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.subtract.html) and could be confused with `pd.diff`.

Perhaps we should take the pandas route and supply a `subtract` function. Then we could make a `compare` function which houses what you describe here. One possible function signature could be:

```
pyam.compare(left, right, left_label='left', right_label='right', kind='diff', drop_close=True, **kwargs)
```

danielhuppmann:
rebased to resolve merge conflicts

danielhuppmann:
I see the point of broader `compare()` function (not so sure that subtracting makes sense or that it can be confused with  'first-difference' as in `pd.diff()`).

Anyway, I refactored to `compare()` and added `drop_close` as a kwarg. Other kwargs such as `kind` (or maybe `method`?) can be added at a later point.

danielhuppmann:
Another idea for a use case of this function (jotted down here for future reference):

`pyam.compare(df, on={'scenario': ('scenario_a', 'scenario_b')})`

which would show the difference between two scenarios with different names within on IamDataFrame (whereas the current implementation compares scenarios with the same name across two IamDataFrames).

gidden:
Awesome, thanks!

PR #177, opened 2019-01-24
Refactor to allow subclasses to set their own time format

znicholls:

- [x] Tests Added

```
```

# Description of PR

This PR moves the formatting of the time column from `pyam/utils.py` into a method of `IamDataFrame`. This allows subclasses to avoid converting to pandas datetime if their timespan is greater than the pandas maximum of 584 years (see https://stackoverflow.com/questions/32888124/pandas-out-of-bounds-nanosecond-timestamp-after-offset-rollforward-plus-adding-a).

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -23,6 +23,7 @@ from pyam.utils import (
     read_files,
     read_pandas,
     format_data,
+    cast_years_to_int,
     pattern_match,
     years_match,
     month_match,
@@ -72,6 +73,8 @@ class IamDataFrame(object):
             _data = read_ix(data, **kwargs)
         else:
             _data = read_files(data, **kwargs)
+
+        _data = self._format_data_time_col(_data)
         self.data, self.time_col, self.extra_cols = _data
         self._LONG_IDX = IAMC_IDX + [self.time_col] + self.extra_cols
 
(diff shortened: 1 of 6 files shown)


pyam/core.py
```diff
@@ -83,6 +86,19 @@ def __init__(self, data, **kwargs):
         if 'exec' in run_control():
             self._execute_run_control()
 
+    def _format_data_time_col(self):
```
gidden: I don't like adding two functions with almost entirely the same name. I think we have two options:

- make a single function that can be inherited by subclasses that must handle both `'year'` and `'time'`
- make two functions, one explicitly for year and one explicitly for time

gidden: Or, appending to option first: move default behavior for year out and only have one function for time

gidden: Thinking of it more, I like option two better. But honestly would prefer if we could get rid of `'year'` entirely...

tests/test_core.py
```diff
@@ -65,6 +65,49 @@ def test_init_df_with_extra_col(test_pd_df):
                                   tdf, check_like=True)
 
 
+@pytest.mark.xfail(reason=(
+    "pandas datetime is limited to the time period of ~1677-2262, see "
+    "https://stackoverflow.com/a/37226672"
+))
+def test_init_df_long_timespan(test_pd_df):
```
gidden: a little pedantic, but could you make three tests named

- `test_init_datetime`
- `test_init_datetime_subclass`
- `test_init_datetime_long_timespan`

And possibly a short comment noting why the subclass test passes

pyam/core.py
danielhuppmann: I think that a substantial part of our users and use cases will still revolve around yearly data, so I'd very much like to retain the option to do `df.filter(year=2020)`. Plus we should cater to the use case of having categorical sub-annual specifications like `['summer-day', 'winter-night']`.

If this can be maintained while removing `year` on the backend, fine by me.

gidden:
Agreed on your points!

pyam/core.py
znicholls: Ok will go for option two. I don't see a way to get rid of year and match the sub-annual specifications yet so will leave that in for now and leave that upgrade to you as I think you know the use cases better.

tests/test_core.py
znicholls: nice suggestion, done

gidden:
looks great! Thanks @znicholls

PR #178, opened 2019-01-25
Adding an option to `rename(append=True)`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR extends the `rename()` function to include an `append` kwarg (which appends the renamed data to `self` or the returned object).

As part of the extension, the logic was slightly reworked: previously, the function replaced all entries iterating over `mapping` (column by column, ie. an 'or'  logic). Now, only data that is touched where all given column renaming mappings apply ('and'  logic).

The previous implementation can still easily be achieved by calling the new function multiple times, i.e., 
```
(
    df
    .rename({'col1': {'old_col1_value': 'new_col1_value'}})
    .rename({'col2': {'old_col2_value': 'new_col2_value'}})
)
```

To simplify the use of this function going forward, the function now also accepts the column rename mappings as kwargs (as well as mapping of mappings), so these two operations are equivalent:
```
df.rename({'col1': {'old_col1_value': 'new_col1_value'}})
df.rename(col1={'old_col1_value': 'new_col1_value'})
```


danielhuppmann:
rebased to resolve merge conflicts

danielhuppmann:
figured out travis is on pandas 22.0 and appveyor is on pandas 24.0, and something in the latest pandas release causes a few unit tests to fail (no issues up to 23.4)

danielhuppmann:
The upper bound on the pandas version seems specified in `requirements.py` (`"pandas>=0.21.0, <=0.23.4",`) seems to be ignored by AppVeyor, so the hard-code line in `appveyor.yml` is required for the time being.

danielhuppmann:
@OFR-IIASA, you implemented the `rename()` function first - can you take a look at this PR?

gidden:
Hi all - initial thoughts/comments. I'm a bit worried about extending our APIs with otherwise common operations. I see that this has already happened in `aggregate`, so it may have been my fault for letting that slip through.

If I understand correctly, these two commands would be equivalent if this PR is merged:

(please tell me if I'm wrong)

```
df.rename(dict, append=True)
```

vs. 

```
df.append(df.rename(dict))
```

If this is true, then I would say -1 to the change of adding another kwarg, when you have an even shorter one-liner to do so. Otherwise, tell me what I'm not understanding, please!

danielhuppmann:
Thanks @gidden for the comment: there is a difference between these two, because `df.append(df.rename(dict))` would lead to duplicates of all data not renamed in the operation - and because of the `groupby.sum`, the values would double without an error message.

To do this correctly, you would have to do:
```
df.append(df.filter(dict1).rename(dict2))
```
where `dict1` is `{'col': ['current_value']}` and `dict2` is `{'col': {'current_value': 'new_value'}}`.

I believe that having these two dictionaries is a sufficient hassle (and potential for errors, see the doubling of values) to warrant its own kwarg.

gidden:
Ok, I think I have been convinced!

gidden:
@OFR-IIASA let me know when you are done and I can take a look as well after

gidden:
PR is in!

pyam/core.py
```diff
@@ -518,35 +518,72 @@ def validate(self, criteria={}, exclude_on_fail=False):
 
             return df
 
-    def rename(self, mapping, inplace=False):
+    def rename(self, mapping={}, inplace=False, append=False, **kwargs):
```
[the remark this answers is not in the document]
danielhuppmann: thanks @OFR-IIASA for the review. I was unable to replicate this error, but I added a few unit tests in the process (see 
https://github.com/IAMconsortium/pyam/pull/178/commits/28738d7c1e1dd4931963c1b2db7a6f2cd765283b). Can you check if it still occurs?

danielhuppmann:
@gidden, can you take a few minutes to review? not sure how to deal with the possible bug identified by @OFR-IIASA...

gidden:
> see here: [28738d7](https://github.com/IAMconsortium/pyam/commit/28738d7c1e1dd4931963c1b2db7a6f2cd765283b) (the link was also provided in my response to the review comment)
> 
> for clarification, the tests don't have the `filter()` addition for conciseness, but I also ran manual trials with that and still wasn't able to reproduce the error

Ok, thanks! @OFR-IIASA could you please provide the exact code snippet you used that failed? If I read your example, you do something like: 

```
df2 = df1.filter(variable='x').rename(append=True, variable={'x': 'y'}) # works
df1.filter(variable='x').rename(variable={'x': 'y'}) # doesn't work
```

But I am not sure exactly what you did from your above text.

appveyor.yml
```diff
@@ -28,6 +28,7 @@ build: false
 
 test_script:
   - pip install -e .[tests]
+  - pip install pandas==0.23.4
```
gidden: please add a comment here as to why this is needed

requirements.py
```diff
@@ -2,7 +2,7 @@
     "argparse",
     "numpy",
     "requests",
-    "pandas >=0.21.0",
+    "pandas>=0.21.0, <=0.23.4",
```
gidden: is this still needed?

pyam/core.py
```diff
@@ -1302,6 +1336,14 @@ def _apply_criteria(df, criteria, **kwargs):
     return df
 
 
+def _make_index_from_data(df):
```
gidden: this could probably be a bit more generic, like:

```
def _make_index(df, cols=['model', 'scenario']):
    return pd.MultiIndex.from_tuples(
       pd.unique(list(zip(df[col] for col in cols)),
       names = tuple(cols)
)
```

pyam/core.py
```diff
@@ -518,35 +518,73 @@ def validate(self, criteria={}, exclude_on_fail=False):
...
+            >> {<column_name>: {<current_name_1>: <target_name_1>,
+            >>                  <current_name_2>: <target_name_2>}}
+            or kwargs as `column_name={<current_name_1>: <target_name_1>, ...}`
         inplace: bool, default False
             if True, do operation inplace and return None
+        append: bool, default False
+            if True, append renamed timeseries to IamDataFrame
         """
+        # combine `mapping` arg and mapping kwargs, ensure no rename conflicts
+        mapping = mapping or {}
```
gidden: should we add a deprecation warning for mapping?

requirements.py
danielhuppmann: this is how I believe it should be specified, but it seems to be ignored by AppVeyor.

gidden: Ah yes, but do we still need to require <=0.23? #181 fixes this then? sorry to just trying to wrap my head around what is going on between the two prs

danielhuppmann: Unfortunately, #181 only fixes the easy parts for now. I suggest to merge this in, then I'll focus on the remaining issues with pandas 0.24.0 (which are in the statistics module and related to a changed functioning of `combine_first()`, so unrelated to this PR).

gidden: Ok, sounds good.

pyam/core.py
danielhuppmann: I went back and forth about that a few times, but in the end, I thought that it would be better to keep both options permanently for mimicking the multiple options in [pd.DataFrame.rename()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.rename.html).

pyam/core.py
```diff
@@ -518,35 +518,72 @@ def validate(self, criteria={}, exclude_on_fail=False):
...
+            raise ValueError(msg)
+        mapping.update(kwargs)
+
+        # changing index and data columns can cause model-scenario mismatch
+        if any(i in mapping for i in META_IDX)\
+                and any(i in mapping for i in ['region', 'variable', 'unit']):
+            msg = 'Renaming index and data cols simultaneously not supported!'
+            raise ValueError(msg)
+
+        filters = {}
```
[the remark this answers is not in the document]

appveyor.yml
danielhuppmann: done

pyam/core.py
danielhuppmann: nice, thanks @gidden!

PR #187, opened 2019-02-04
Fix `line_plot()` to work with `nan` in meta columns

danielhuppmann:

- [x] Tests Added (extended existing `line_plot()` test with metadata, first failed, then passed)
- [x] Documentation Added

# Description of PR

The function `line_plot()` drops all rows where any value is `nan` in the data or meta tables, even if the columns in questions aren't used. This PR adds a function `join_meta()` where only columns (in particular those relevant to the plotting library) are joined to the dataframe which is passed to the plotting functions.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -331,11 +331,11 @@ class IamDataFrame(object):
                 cols.add(value)
 
         return (
-                self.data
-                .set_index(META_IDX)
-                .join(self.meta[list(cols)])
-                .reset_index()
-               )
+            self.data
+            .set_index(META_IDX)
+            .join(self.meta[list(cols)])
+            .reset_index()
+        )
 
     def timeseries(self, iamc_index=False):
         """Returns a pd.DataFrame in wide format (years or timedate as columns)
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -325,7 +325,7 @@ class IamDataFrame(object):
         kwargs: arguments passed to plotting library, meta columns (as values)
            are joined to returned dataframe
         """
-        cols = set()
+        cols = set(['exclude'])
         for arg, value in kwargs.items():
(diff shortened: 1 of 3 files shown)


gidden:
Nice! I'm a bit curious how you see the use of this `join_meta()` function. 

Is it something that "normal/every day" users will want with some regularity? How should it be used vis-a-vis `as_pandas()`? Should it replace it? Should it be used with the `with_metadata` option? 

I think now that the package is getting more mature (and bigger), we should try to avoid code bloat wherever possible and reuse as much as possible. Thus my intuition says that `as_pandas` should probably use this functionality somehow.

danielhuppmann:
As discussed offline with @gidden, I added the feature for column discovery to the `as_pandas()` function, so if `with_metadata` is a dictionary, it will only merge meaningful columns, keeping the existing API unchanged.

Also, if `with_metadata=False`, the function now returns a copy of the `data` rather than the data frame itself.

gidden:
looks like one of the win ci builds failed on an http request, rebuilding

pyam/core.py
```diff
@@ -1079,7 +1068,7 @@ def line_plot(self, x='year', y='value', **kwargs):
 
         see pyam.plotting.line_plot() for all available options
         """
-        df = self.join_meta(x=x, y=y, **kwargs)
+        df = self.as_pandas(with_metadata=kwargs)
```
gidden: do we need to include `x` and `y` here? or no?

danielhuppmann: I don't think we do - if `x` or `y` are variables, they will be included in the returned dataframe anyway. And I don't think we can do `x` or `y` as metadata columns currently?

gidden:
thanks @danielhuppmann !

PR #189, opened 2019-02-05
Fix `scatter()` to work with `nan` in meta columns

danielhuppmann:

- [x] Tests Added (tests continue to work)
- [x] Documentation Added

# Description of PR

This PR extends the fix for the ` line_plot()` feature with `nan` in meta columns (see #187) to the `scatter` function.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -324,6 +324,7 @@ class IamDataFrame(object):
             return self.data.copy()
 
     def _discover_meta_cols(self, **kwargs):
+        """Discover applicable meta columns from kwargs"""
         cols = set(['exclude'])
         for arg, value in kwargs.items():
             if isstr(value) and value in self.meta.columns:
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -310,10 +310,8 @@ class IamDataFrame(object):
            meaningful meta columns from values (in key-value)
         """
         if with_metadata:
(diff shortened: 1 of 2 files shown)


pyam/core.py
```diff
@@ -311,21 +311,25 @@ def as_pandas(self, with_metadata=False):
...
                 self.data
                 .set_index(META_IDX)
-                .join(self.meta[list(cols)])
+                .join(self.meta[cols])
                 .reset_index()
             )
         else:
             return self.data.copy()
 
+    def _discover_meta_cols(self, **kwargs):
```
gidden: even thought its private, could you please add a docstring explaining what this function does?

pyam/core.py
```diff
@@ -311,21 +311,25 @@ def as_pandas(self, with_metadata=False):
         """
         if with_metadata:
             if isinstance(with_metadata, dict):
```
gidden: now this can be a one-liner:

```
cols = self._discover_meta_cols(**with_metadata) if isinstance(with_metadata, dict) else self.meta.columns
```
```

gidden:
one last request, then g2g!

pyam/core.py
danielhuppmann: well, two lines or >80 ;-)

gidden:
thanks @danielhuppmann !

PR #196, opened 2019-02-18
[TEST] Trying to relax ci install reqs

gidden:
Here I am testing the issues with pandas>24, which have been observed on windows, but not linux thus far.


gidden:
@danielhuppmann it looks like we can relax the pandas constraint (at least on CI), does this jive with your understanding?

PR #198, opened 2019-02-19
Add tutorial plotting with aggregation

rossursino:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

Adds a tutorial demonstrating how to make stack plots after aggregating variables with pyam. In the process it updates the `stack_plot` functionality to better handle negative emissions. [This notebook](https://github.com/IAMconsortium/pyam/blob/59da398348ac05a2930e8d119b9e58ba712bb7c6/doc/source/tutorials/plot_aggregated_variables.ipynb) illustrates the end result. The cell with "7" next to it of [this notebook](https://github.com/IAMconsortium/pyam/blob/f394fc21052a1d203816d039e4b8ee8178f0b8b5/doc/source/tutorials/plot_aggregated_variables.ipynb) shows the problem with the current implementation. In particular, agg just disappears once the cars subsector goes negative and thereafter the plot is useless.


gidden:
Hi @rossursino thanks for the PR! Let us know when it's ready for review.

rossursino:
@gidden Just working with @znicholls to get this passing the tests then it will be ready for review.

znicholls:
@gidden what's the trick for making expected figures in the right way i.e. what call to `plt.savefig` do you se?

gidden:
> @gidden what's the trick for making expected figures in the right way i.e. what call to `plt.savefig` do you use?

In short, make the test look like the others and generate them as stated in `tests/README.md`.

Slightly longer - the mpl pytest plugin has a "generate expected figures" mode which will run the tests just to generate the figures (they do the `savefig` for you). In the test itself, you have to return a `matplotlib.Figure` object. Everything about this is pretty finicky, so I have done my best to standardize as many "defaults" as possible via the `MPL_KWARGS` dict in the test suite. The biggest issue I have run in to is the `Freetype` library version - all figures at the moment are generated with `v2.6.1`. Which version you have depends on how you installed matplotlib - in the CI, for instance, we have to [monkey around a bit](https://github.com/IAMconsortium/pyam/blob/master/.travis.yml#L30) to get it to work correctly across all operating systems. 

If I was smarter, I would probably make an `environment.yml` just for generating the figures... pull requests welcome if you feel so inclined =)

znicholls:
@gidden I think this is good to go (no idea what stickler is complaining about anymore...) thoughts?

gidden:
hey @znicholls maybe just open a new PR on the same branch? that should show you what stickler is complaining about. (

znicholls:
@gidden this notebook (https://github.com/IAMconsortium/pyam/blob/59da398348ac05a2930e8d119b9e58ba712bb7c6/doc/source/tutorials/plot_aggregated_variables.ipynb) is the important one. Cell 7 of this (https://github.com/IAMconsortium/pyam/blob/f394fc21052a1d203816d039e4b8ee8178f0b8b5/doc/source/tutorials/plot_aggregated_variables.ipynb) shows the problem with the current implementation.

znicholls:
ping @gidden

gidden:
trying to get to this today, sorry for the delays!

pyam/plotting.py
```diff
@@ -375,19 +375,103 @@ def stack_plot(df, x='year', y='value', stack='variable',
...
+    # then cast back to datetime at the end..?
+    _df.index = _df.index.astype(float)
+
+    time_original = _df.index.values
+    first_zero_times = pd.DataFrame(index=["first_zero_time"])
+    both_positive_and_negative = _df.apply(
+        lambda x: (x >= 0).any() and (x < 0).any()
+    )
+
+    for col in _df.loc[:, both_positive_and_negative]:
```
gidden: I get what this is doing, but it still may be worth adding a small comment at the top explaining it.

I think this could be a little tighter using `np.argmax` or, for example

```
positive = values[values >= 0]
negative = values[values <= 0]
crosses = positive[::-1] & negative[1::]
```

and then you can calculate the time for each index where crosses is true

pyam/plotting.py
```diff
@@ -375,19 +375,103 @@ def stack_plot(df, x='year', y='value', stack='variable',
...
+    _df = _df.reindex(sorted(_df.index)).interpolate(method="values")
+
+    # Sort lines so that negative timeseries are on the right,
+    # positive timeseries are on the left and timeseries which
+    # go from positive to negative are ordered such that the
+    # timeseries which goes negative first is on the right
+    # (case of timeseries which go from negative to positive
+    # is an edge case we haven't thought about as it's unlikely
+    # to apply to us).
+    col_order = [
```
gidden: I think this can be cleaned up a bit too, e.g.,

```
col_order = [c for c in _df if (_df[c] > 0).all()] + first_zero_times.columns[::-1] + [c for c in _df if (_df[col] < 0).all()] 
```

pyam/plotting.py
```diff
@@ -375,19 +375,103 @@ def stack_plot(df, x='year', y='value', stack='variable',
...
+        col_order[::-1]
+    ].cumsum(axis=1)[
+        col_order
+    ]
+
+    total_kwargs = kwargs.pop("total_kwargs", {})
+    plt_total = kwargs.pop("total", False)
+    time = _df.index.values
+    upper = positive_only_cumulative.iloc[:, 0].values
+    for j, col in enumerate(_df):
```
gidden: I'm having a little trouble understanding what is going on here, could you explain and I can continue the review?

znicholls: Ah ok yep this bit is tricky, mainly because making sure the top and bottom are in the right spot is hard (by default matplotlib doesn't put them in the right spot for this application).

In line 454 we set our upper limit for the first segment to plot. That is simply the sum of the positive only cumulative values  i.e. your upper limit should be either zero, if there are no positive contributions, or more, if there are positive values (cumulative is probably misleading here as it's really the sum of timeseries rather than a cumulative sum over time for a given variable...).

As we loop over the columns of `_df`, which has been sorted from positive to negative, we are basically plotting from the top line (the sum of all positive contributions) down. Hence we next need to find where the lower line for this particular section of the stack should be. 

The starting point for that is `poc_nr` i.e. the sum of all positive contributions up until the next row (maybe we should swap this order in the code). However, if this row includes some negative values, then we need to make sure they show up too. That's what line 462 does: it ensures that if this row has negative values, the lower line will go to those negative values rather than getting stuck at the lower limit of the sum of positive values only i.e. zero. 

Then we have both our upper and lower limits so we can use `fill_between` before finally setting `upper = lower` so that the next segment to plot will appear in the right place. 

Hopefully that helps a bit.

gidden: Thanks! I think I get it now

znicholls:
any other comments @gidden or shall I sit down with @rossursino and have a go at tweaking?

gidden:
I can try to look through again once more, but probably no time today

gidden:
Hey @znicholls let me know when the CI stuff is out of this PR and I can look at the pure plotting implementation and offer feedback.

gidden:
Hey @znicholls and @rossursino where should I be looking for a final review? Is this rebased on current (or reasonably current) master now?

znicholls:
> where should I be looking for a final review?

Here. For some reason github isn't updating the PR, it should actually look like https://github.com/IAMconsortium/pyam/pull/223/files...

gidden:
Ok sounds good. Just let me know when to take a look =)

znicholls:
> Here. For some reason github isn't updating the PR, it should actually look like https://github.com/IAMconsortium/pyam/pull/223/files...

Alright all good now, just changing the base back and forth seems to have done it

gidden:
Ok, this looks spectacular! Thanks so much @rossursino and @znicholls for seeing this one through.

PR #199, opened 2019-02-20
Initialize an `IamDataFrame` from `pd.DataFrame` with formatting specs

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR allows initialising an `IamDataFrame` from a file or `pd.DataFrame` and filling out missing required columns efficiently.

See the unit tests in `tests/test_cast_to_iamc.py` for possible use cases.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -26,6 +26,7 @@ from pyam.utils import (
     sort_data,
     to_int,
     find_depth,
+    concat_with_pipe,
     pattern_match,
     years_match,
     month_match,
@@ -1524,11 +1525,17 @@ def df_to_pyam(df, **kwargs):
         the data to be cast to an `IamDataFrame`
     value: str or list, must be columns of `df`
         use data in these columns as `value`, use column name as `variable`
+    kwargs
+        mapping of column required for an `IamDataFrame` to
+        - one column in `df`
+        - multiple columns, which will be concatenated by pipe
+        - a string to be used as value for this column
     """
     # ensure that only either `value` or `variable` custom setting is used
     if all([i in kwargs or i in df.columns for i in ['value', 'variable']]):
         raise ValueError('using both `value` and `variable` is not valid!')
 
+    # if `value` arg is given, melt columns and use column name as `variable`
     if 'value' in kwargs:
         value = kwargs.pop('value')
         idx = set(df.columns) & (set(IAMC_IDX) | set(['year', 'time']))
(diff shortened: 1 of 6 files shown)


danielhuppmann:
@znicholls, this PR tries to make it easier to initialize an `IamDataFrame` which doesn't quite have the right format, for example when you have a `pd.DataFrame` in a `scen`-file format with values in multiple columns and the column name being the identifier. See also the discussion here https://github.com/IAMconsortium/pyam/pull/193#issuecomment-464634324.

Can you check whether that works with your thinking for the `ScmDataFrame`?

znicholls:
will leave final review and merge to @gidden nice tweak @danielhuppmann

pyam/core.py
```diff
@@ -1501,13 +1500,13 @@ def compare(left, right, left_label='left', right_label='right',
 
 def concat(dfs):
     """Concatenate a series of `pyam.IamDataFrame`-like objects together"""
-    if not hasattr(dfs, '__iter__'):
+    if isstr(dfs) or not hasattr(dfs, '__iter__'):
         raise TypeError('Input data must be iterable (e.g., list or tuple)')
```
gidden: can you add here "non-string iterable" just for consistency..

pyam/utils.py
```diff
@@ -194,18 +235,18 @@ def format_data(df):
...
-def style_df(df, style='heatmap'):
-    if style == 'highlight_not_max':
-        return df.style.apply(lambda s: ['' if v else 'background-color: yellow' for v in s == s.max()])
-    if style == 'heatmap':
-        cm = sns.light_palette("green", as_cmap=True)
-        return df.style.background_gradient(cmap=cm)
+
+def sort_data(data, cols):
+    """Sort `data` rows and order columns"""
+    return data.sort_values(cols)[cols + ['value']].reset_index(drop=True)
```
gidden: what happens if "value" is already in `cols`? not sure if this is a corner case, but could do `list(set(cols + ['value']))`

gidden:
ok, small changes requested. looks great!

pyam/utils.py
danielhuppmann: Usig `list(set(cols + ['value']))` would defeat the purpose of having a consistent ordering of columns, which makes the unit tests a lot more straightforward by using `pd.testing.assert_frame_equal()`.

I tried migrating that function to `core.py`, but then this function can't be used in the unit tests, which is opening up another can of worms.

pyam/utils.py
```diff
@@ -395,3 +436,16 @@ def to_int(x, index=False):
         return x
     else:
         return _x
+
+
+def concat_with_pipe(x, cols=None):
+    """Concatenate a `pd.Series` separated by `|`, drop `None` or `np.nan`"""
+    return '|'.join([x[i] for i in cols or x.index
```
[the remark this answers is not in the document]
danielhuppmann: done

pyam/core.py

gidden:
thanks @danielhuppmann !

PR #201, opened 2019-02-26
Outside legends

gidden:

- [x] Tests Added
- [x] Documentation Added

Add tutorial and native support for legends outside of plots


gidden:
@znicholls @rossursino here is an example where I added a new figure test

znicholls:
> here is an example where I added a new figure test

thank you!

gidden:
Don't disagree, but matplotlib has lots of helpful messages about what
arguments are available, and while we could do something similar, it is a
much larger amount of work than I can put in for this PR.

PR #202, opened 2019-02-26
Add a validation check to `rename()`

### LINKED ISSUE #182: Overly ambitious `groupby().sum()´ in `rename()`
## Description of the problem

We identified a silly mistake in the `IAMC 1.5°C Scenario Data` due to the over-ambitious use of `groupby().sum()` in the `rename()` function implementation. See https://github.com/iiasa/ipcc_sr15_scenario_analysis/issues/2 for details.

There were a number of variables with a typo, and when using `rename()` to correct that, I failed to check whether any variable without the typo also existed. This was the case for two variables, so the values are now double compared to what they should be.

## Proposed solution

I would like to add a kwarg `check_duplicate (default True)` to the function, which raises an error message if any rename mapping leads to a conflict with the existing `model-scenario-region-variable-unit-year` index in the data. A user would need to consciously override the check to merge renamed and existing values as part of the rename. Idea: fail with a bang rather than return unexpected results.

The merging of data renamed to the same `variable` name should obviously always work.

@gidden, @OFR-IIASA, any comments?

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a validation check to `df.rename()` to prevent accidentally renaming and aggregating existing and renamed variables.

In terms of the new unit test `test_rename_duplicates()`, the user may not be (actively) aware that a variable `test_3` exists in the `IamDataFrame` and not actually want to aggregate `test_1` and `test_3`. With the new feature, the user will receive an error message as default behaviour and has to actively override the validation step.

closes #182

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -611,8 +611,9 @@ class IamDataFrame(object):
                         _data.loc[~rows, self._LONG_IDX].drop_duplicates())
             )
             if any(merged.duplicated()):
-                msg = 'duplicates between original and renamed data!'
-                raise ValueError(msg)
+                msg = 'Duplicated rows between original and renamed data!\n{}'
+                conflict_rows = merged.loc[merged.duplicated(), self._LONG_IDX]
+                raise ValueError(msg.format(conflict_rows.drop_duplicates()))
 
         # merge using `groupby().sum()`
         ret.data = _data.groupby(ret._LONG_IDX).sum().reset_index()
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -608,7 +608,7 @@ class IamDataFrame(object):
         if check_duplicates:
             merged = (
                 _data.loc[rows, self._LONG_IDX].drop_duplicates().append(
-                        _data.loc[~rows, self._LONG_IDX].drop_duplicates())
+                    _data.loc[~rows, self._LONG_IDX].drop_duplicates())
             )
             if any(merged.duplicated()):
                 msg = 'Duplicated rows between original and renamed data!\n{}'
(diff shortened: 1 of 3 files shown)


danielhuppmann:
@gidden, ready for review if you have a minute...

pyam/core.py
```diff
@@ -580,6 +589,7 @@ def rename(self, mapping=None, inplace=False, append=False, **kwargs):
         # renaming is only applied where a filter matches for all given columns
         rows = ret._apply_filters(filters)
         idx = ret.meta.index.isin(_make_index(ret.data[rows]))
+        _data = ret.data.copy()
```
gidden: why do we need to make a copy here?

danielhuppmann: In case `check_duplicates=True` and a `ValueError` is raised, we don't want any changes to the `IamDataFrame` to persist.

But I guess `_data = ret.data.copy() if check_duplicates else ret.data` would be slightly more efficient, right?

gidden: Agreed. Can you add a warning in the docstring stating that this makes a copy as well? Also, are you sure we want that to be `True` by default? What are the pro/cons?

danielhuppmann: Added the more elaborate only-make-interim-copy-if-`check_duplicates` and an inline comment to explain that the interim copy is kept until after the validation check.

danielhuppmann: > Also, are you sure we want that to be `True` by default? What are the pro/cons?

I think that the more careful behaviour (`check_duplicates == True`) should be the default. Users should explicitly have to override the sanity check when (as in the unit test) renaming timeseries `test_1` to `test_3` *and aggregating* with existing timeseries data of `test_3`.

gidden:
looks good, thanks @danielhuppmann !

PR #203, opened 2019-02-27
extend use case for `format_data()` to take `variable` and `value`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Working on the `message_ix.Reporting` class, I realized that we would need a way to cast an `IamDataFrame` with both a `variable` and a `value` arg passed to the "fill-by-value/rename/concat" part of the `format_data()` function,


gidden:
thanks @danielhuppmann !

PR #207, opened 2019-03-07
Add a function to aggregate a variable to a region from subregions

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR refactors the `check_aggregate_regions()` function into a separate `aggregate_region()` function and a `check_...()` function to use it in scenario postprocessing (calculating regional values rather than just checking that the aggregation is correct).

The important issue here is that any variables components that are only defined at the `region` (e.g., 'World') should be added to the regional total. Say that CO2 emissions from air travel are only accounted for at the global level, the following output should be returned:

**Input**

| region  | variable                             | unit      | 2020 |
| --------|---------------------------|---------|-------|
| Europe | Emissions\|CO2                 | GtCO2 |  5      |
| Africa   | Emissions\|CO2                 | GtCO2 |  3      |
| World  | Emissions\|CO2\|Air Travel | GtCO2 |  1      |

**Expected output**

| region  | variable                             | unit      | 2020 |
| --------|---------------------------|---------|-------|
| World  | Emissions\|CO2                 | GtCO2 |  9      |

## Refactoring as part of this PR

While implementing, there were a number of issues that I believed should be improved even though they break the API.

@znicholls, you implemented the first version of this, can you mark the items below if you agree with the change?

 - [x] all `[check_]aggregate[_regions]()` functions: refactor `units` to `unit` for closer resemblance to `df.filter()`
 - [x] the `check_aggregate...()` functions return the index columns in the standard IAMC-order
   (i.e, `[model, scenario, region, variable, unit]`)
 - [x] the function `check_aggregate_regions()` was renamed to `check_aggregate_region()` because it only ever checks one region at a time
 - [x] the function `check_aggregate_regions()` takes a kwarg `subregions` for an optional list of regions to aggregate and a kwarg `components` for variable components to be included at the `region` level
   (before, it was not possible to select custom variable components and `components` referred to the custom subregions, differing from the use in `check_aggregate()`
 - [x] `_apply_filters()` interprets `col=None` as no filter applied, i.e, all `True` (before, it would return all `False`). This streamlines passing the unit filter from the `[check_]aggregate[_regions]()` and is in line with how pandas treats `slice(None)`.
 - [x] `_apply_filters()` takes `kwargs`, not a `dict`


gidden:
Hey @danielhuppmann quick first question. I think we already have an `aggregate_regions` style function in [`map_region`](https://github.com/IAMconsortium/pyam/blob/master/pyam/core.py#L1205). Is there overlap here? Happy to use any suggestions you have to improve/change it. Plus I see the docstring is definitely wrong in its summary...

pyam/core.py
```diff
@@ -722,118 +716,140 @@ def check_aggregate(self, variable, components=None, units=None,
...
+        # default subregions to all regions other than `region`
+        if subregions is None:
+            rows = self._apply_filters(variable=variable)
+            subregions = set(self.data[rows].region) - set([region])
 
-    def check_aggregate_regions(self, variable, region='World',
-                                components=None, units=None,
-                                exclude_on_fail=False, **kwargs):
+        if not len(subregions):
+            msg = 'cannot aggregate variable `{}` to `{}` because it does not'\
```
znicholls: is this message right @danielhuppmann ?

znicholls:
>     * `_apply_filters()` takes `kwargs`, not a `dict`

Probably need to ask @gidden about this one I think?

Otherwise looks very nice to me

pyam/core.py
danielhuppmann: I'd say yes: if the variable `Emissions|CO2` is not defined in any subregion, the function should not make any assumption that this is zero at the `World` level. And if you want to add up subcategories of `Emissions|CO2|*` at the World level, you should use `aggregate()` rather than `aggregate_region()`.

danielhuppmann:
> Hey @danielhuppmann quick first question. I think we already have an `aggregate_regions` style function in [`map_region`](https://github.com/IAMconsortium/pyam/blob/master/pyam/core.py#L1205). Is there overlap here? Happy to use any suggestions you have to improve/change it. Plus I see the docstring is definitely wrong in its summary...

1) why do you think that the docstring is wrong?

2) well, I didn't have `map_regions()` on my radar... dug into it a bit now:

    **main distinction**
    - `map_regions()`
        - implicitly assumes that the variable tree is complete
        - supports different mappings of subregions to regions per model
        - returns an average (?) over each subregion, see footnote [1]
    - `aggregate_region()`
        - intended to build up a region "tree" similar to a variable tree (calling it recursively over a variable tree and region hierarchy from bottom up)
        - implicitly assumes a common (or at least not conflicting) region names
        - includes variable components that are only defined at the `region` level (not the subregions)

    **similarities**
    One can use one function to perform the duty of the other, I think, see footnote [2]


[1] doing `IamDataFrame(REG_DF).map_regions('r5_region').timeseries()`,
where `REG_DF` is from `conftest.py` does NOT return the `World` timeseries

[2] e.g., for `df = IamDataFrame(REG_DF).filter(model='IMAGE')`
```
df.aggregate_region(variable='Primary Energy', region='R5MAF', subregions=['NAF', 'ME'])
```
is similar to (except for the "average" issue in [1])
```
df.map_regions('r5_region').timeseries()
```
where `region='R5MAF', subregions=['NAF', 'ME']` could also be read dynamically from `run_control()` as is implemented in `map_regions()`.

gidden:
I meant `map_regions` docstring is wrong, sorry! =)

gidden:
Ok, just getting back to this now. I agree that there are differences between the two, but my suspicion here is that we should harmonize them into one function (or two, but different). 

The original goal of `map_regions` was to be able to take *either* a many-to-one or one-to-many mapping and plot them. Perhaps it would be better to break this into two different functions, one to aggregate, and one to "paint" (for lack of a better word). What do you think? In principle, it should be relatively easy to update the tests/tutorials that go in "that direction" (many-to-one).

What do you think?

danielhuppmann:
Agree that the two could, maybe should, be refactored into a data-manipulation and a “paint” function. But...

`aggregate_regions()` has the extra feature to automatically search for and add subcategories of the one variable and one region that it operates on that are not defined in the subregions.

`map_regions()` just aggregates all variables without consideration of the variable tree and whatever region mapping it gets - or it loads a default region mapping from file and does some model-dependent best-guessing how to apply the mapping.

Merging these two features will require a lot kwargs...

I did go down that rabbit hole to harmonise them and I did not see a light after an hour or two, so I abandoned the effort. We can venture into it again together next week...

gidden:
Per in-person discussions, we have tentitavely agreed to break this into two functions:

1) `aggregate_regions()` for the many-to-one mapping
2) `downscale_regions()` for the one-to-many mapping

danielhuppmann:
@gidden, merge conflicts resolved, should be good to be merged once the CI passes

pyam/core.py
```diff
@@ -733,118 +727,140 @@ def check_aggregate(self, variable, components=None, units=None,
...
+            msg = '`{}` - {} of {} rows are not aggregates of components'
             logger().info(msg.format(variable, len(diff), len(df_variable)))
 
             if exclude_on_fail:
                 self._exclude_on_fail(diff.index.droplevel([2, 3, 4]))
 
-            diff = pd.concat([diff], keys=[variable], names=['variable'])
+            return IamDataFrame(diff, variable=variable).timeseries()
+
+    def aggregate_region(self, variable, region='World', subregions=None,
```
gidden: A few thoughts here:

1) Instead of a region/subregion argument, there should be something like `mapping` which is a dict, dataframe, etc. with key-value pairs of existing-region to aggregate region
2) `unit` here seems to be a filtering argument, I think that should not be here but used explicitly with `filter` before hand
3) my gut feeling is that `component` here is extraneous (a user should instead do df.aggregate_region().aggregate_variable()`, but I think based on our discussion this is unavoidable - let me know what you think

pyam/core.py
```diff
@@ -733,118 +727,140 @@ def check_aggregate(self, variable, components=None, units=None,
...
+            rows = region_df._apply_filters(variable=components)
+            _data = _data.add(_aggregate(region_df.data[rows], cols),
+                              fill_value=0)
+
+        if append is True:
+            self.append(_data, region=region, variable=variable, inplace=True)
+        else:
+            return _data
+
+    def check_aggregate_region(self, variable, region='World', subregions=None,
```
gidden: here I would actually make the same comment on the `mapping` arg, let me know if you disagree. @znicholls may also have an opinion here..

pyam/core.py
```diff
@@ -733,118 +727,140 @@ def check_aggregate(self, variable, components=None, units=None,
...
                 self._exclude_on_fail(diff.index.droplevel([2, 3]))
 
-            diff = pd.concat([diff], keys=[region], names=['region'])
-            diff.index = diff.index.swaplevel(i=-1, j=-2)
-            return diff.unstack().rename_axis(None, axis=1)
+            col_args = dict(region=region, variable=variable)
+            return IamDataFrame(diff, **col_args).timeseries()
+
+    def _variable_components(self, variable):
+        """Get all components of a variable"""
```
gidden: I think it maybe worth showing an example in the docstring

pyam/core.py
danielhuppmann: Re 1: seems to me a question of personal preference whether to pass a dict or explicit args. Question would be how to ensure that the dict-way is meaningful (allow overlapping region aggregation?, order of aggregation if you have multiple levels of region hierarchy?).

danielhuppmann: Re 3: note that this function adds variable components that only exist at the aggregate-region Level to the new timeseries, so this has to be part of this function.

pyam/core.py
danielhuppmann: See response above.

pyam/core.py
znicholls: > 1. Instead of a region/subregion argument, there should be something like `mapping` which is a dict, dataframe, etc. with key-value pairs of existing-region to aggregate region

This would allow you to do multiple region aggregations at once right? If so, perhaps that could be in a helper method (i.e. this method would stay, and would always just do one region at a time, and we might have another, `aggregate_regions` which would allow you to define a mapping in order to do many regions, all to different 'targets', at once).

> 2\. `unit` here seems to be a filtering argument, I think that should not be here but used explicitly with `filter` before hand

+1

> 3\. my gut feeling is that `component` here is extraneous (a user should instead do df.aggregate_region().aggregate_variable()`, but I think based on our discussion this is unavoidable - let me know what you think

Thinking about it more, I'm not sure that this argument is actually wise. The only case I can think of where you'd want it is if you wanted to introduce an inconsistency into your dataset (e.g. you wanted 'CO2|Air' not to be included in 'CO2' at the world level). I'm not sure if we should allow this. 

If you just drop this keyword, then you ensure that the data stays internally consistent and if users want to do something else, that's up to them.

znicholls:
Just catching up on this now

> Per in-person discussions, we have tentitavely agreed to break this into two functions:
> 
> 1. `aggregate_regions()` for the many-to-one mapping
> 2. `downscale_regions()` for the one-to-many mapping

So in this PR you add `aggregate_regions()`, in a future PR we will add `downscale_regions()`?

pyam/core.py
danielhuppmann: Re 1: we could refactor `map_regions()` to take a mapping, do some consistency checks (e.g., regions exist?), and then call `aggregate_region()` iteratively or (in the future `downscale_region()`), and return the concatenated `IamDataFrame` of the new region mapping (or append to `self`). That would make for multiple shorter functions, rather than one beast of a function...

danielhuppmann: Re @znicholls and 3: unfortunately, we cannot force IAM users to be consistent in their variable tree hierarchy, so we will have variables like `Primary Energy|Coal` and `Primary Energy|Fossil` in one `IamDataFrame` until at least IPCC AR8 (*). So the function needs to have a somewhat decent way to handle such cases - and requiring the user to provide a custom list of `components` seems to me the easiest way.

(*) Yes, one could start `Primary Energy|Fossil|Coal`. Not gonna happen...

danielhuppmann:
> Just catching up on this now
> 
> > Per in-person discussions, we have tentitavely agreed to break this into two functions:
> > 
> > 1. `aggregate_regions()` for the many-to-one mapping
> > 2. `downscale_regions()` for the one-to-many mapping
> 
> So in this PR you add `aggregate_regions()`, in a future PR we will add `downscale_regions()`?

Yes (but renaming the functions to `aggregate_region()` and `check_aggregate_region()` because it is only one region being treated).

And as suggested in the inline discussion above, we could refactor the existing `map_regions()` to take a region-mapping and then call the functions iteratively as required.

pyam/core.py
znicholls: > unfortunately, we cannot force IAM users to be consistent in their variable tree hierarchy

ah yes I forgot about this. Keeping `components` sounds good then

pyam/core.py

pyam/core.py
gidden: On 1, I would think it would be pretty straightforward to use [`pd.Series.map`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html) to change region names as desired, then `groupby().sum()`. My suggestion here is that we do our best to let pandas take care of error messaging, etc. for us. I really think we should not try to reinvent the wheel, and pass as much as we can directly to pandas methods.

gidden:
hey @danielhuppmann. it turns out the geopandas dep is more complicated than initially thought - it seems some underlying datasets have changed (perhaps ISO names? I would need to dig further)... For the moment, can you try updating the CI conda installs to `geopandas<0.5.0` and I can make an issue

appveyor.yml
```diff
@@ -26,7 +26,7 @@ install:
   - conda --version
   - activate testing
   - conda install -y numpy pandas pyyaml xlrd xlsxwriter matplotlib==3.0.3 seaborn==0.9.0 six requests jupyter nbconvert
-  - conda install -y -c conda-forge libiconv gdal fiona geopandas cartopy
+  - conda install -y -c conda-forge libiconv gdal fiona geopandas=0.4.1 cartopy
```
gidden: `conda install -y -c conda-forge libiconv gdal fiona "geopandas<0.5.0" cartopy`

.travis.yml
```diff
@@ -28,6 +28,7 @@ before_install:
 
 install:
   - make install
+  - conda update -y -c conda-forge geopandas=0.4.1
```
gidden: this should go in pyam/ci/environment-conda-forge.txt (where you should be able to change the entry to `geopandas<0.5.0`.

danielhuppmann: done

danielhuppmann:
force-pushed to get rid of my own very sorry efforts at implementing @gidden's suggestion

cherrypicked from #226

danielhuppmann:
Thanks for the assist with `geopandas`, @gidden!

I made an issue for the refactoring of `map_regions()` (#225) as discussed. So should be good to go!

gidden:
Woo, thanks so much @danielhuppmann. can you please add that conversation to the `map_region()` issue?

gidden:
hah, beat me to it =)

pyam/core.py
znicholls: what was the resolution here? Use a dict map as a refactoring in future?

danielhuppmann: For now, we will refactor `map_regions()` to take a mapping ad call `aggregate_region()` or `downscale_region()´ iteratively, see #225.

Later on, extending `aggregate_region()` to take a `mapping` instead of the string args `region` and `subregions` may be practical for speedup.

PR #212, opened 2019-04-02
R style columns

gidden:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Who else has gotten very useful data from very useful colleages in that very pesky R-style data format, where year columns look like `X2015` instead of just `2015`? Worry no more, `pyam` has your back!

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -148,7 +148,7 @@ def format_data(df, **kwargs):
             # not a string/iterable/etc, fall down to final return statement
             pass
         return c
-    df.columns = map(convert_r_columns, df.columns)
+    df.columns = df.columns.map(convert_r_columns)
 
     # if `value` is given but not `variable`,
     # melt value columns and use column name as `variable`
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -137,15 +137,13 @@ def format_data(df, **kwargs):
         try:
             first = c[0]
             second = c[1:]
-            if first != 'X':
-                # not in the X2015 R-style, fall down to final return statement
-                pass
-            try:
-                #  bingo! was X2015 R-style, return the integer
-                return int(second)
-            except:
-                # nope, not an int, fall down to final return statement
-                pass
+            if first == 'X':
+                try:
+                    #  bingo! was X2015 R-style, return the integer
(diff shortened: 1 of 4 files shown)


gidden:
Mind taking a look @znicholls? I want to give @danielhuppmann a bit of a break from reviews (at least for one!).

pyam/utils.py
```diff
@@ -132,6 +132,24 @@ def format_data(df, **kwargs):
...
+                    #  bingo! was X2015 R-style, return the integer
+                    return int(second)
+                except:
+                    # nope, not an int, fall down to final return statement
+                    pass
+        except:
+            # not a string/iterable/etc, fall down to final return statement
+            pass
+        return c
+    df.columns = map(convert_r_columns, df.columns)
```
znicholls:
```suggestion
    df.columns = df.columns.map(convert_r_columns)
```

may be more pandas like?

gidden:
Ok, can do. Errors in CI are not related, FYI.

znicholls:
> Errors in CI are not related, FYI.

I guessed as much. Do you want me to split my fix from #198 out into a separate PR to see if that helps?

znicholls:
> Do you want me to split my fix from #198 out into a separate PR to see if that helps?

Ignore that, just looked at the errors properly

pyam/utils.py
```diff
@@ -132,6 +132,24 @@ def format_data(df, **kwargs):
     if isinstance(df, pd.Series):
         df = df.to_frame()
 
+    # Check for R-style year columns, converting where necessary
```
danielhuppmann: suggest to move that function out of `format_data()` - the function is already too long...

gidden: But it is also the catch-all function for saying "here is an input data set, make it into a `pyam.IamDataFrame`. I agree we could break `format_data` into multiple functions that all get called, but I don't think that's in the scope of this PR. Happy to make an issue though!

gidden:
tests will fail again because our backends are down from public access..

gidden:
should be good to go now

gidden:
once issue discussion resolved with @danielhuppmann

PR #214, opened 2019-04-08
Tidy up requirements

znicholls:

- [x] Tests Added (N/A)
- [ ] Decide how to clean up multiple requirements files
- [ ] Decide how to document
- [ ] Documentation Added 

```
```

# Description of PR

Tidies up requirements to ensure it is clear what is required to use pyam, what is required for the CI etc.


znicholls:
@gidden @danielhuppmann at the moment it looks like there's requirements in multiple different places (`ci/`, `requirements.txt`, `requirements.py`). Is it worth trying to put these all in one place? If so, do you have a preference of where?

gidden:
No strong preference, as long as it is documented. Thanks for helping us
tidy up! =)

gidden:
But yes, to address your other statement, I believe it is worth consolidating all of this in one spot. The one thing we may need to have a copy of is a specific environment to replicate unit tests on CI (e.g., where to update the matplotlib requirement to get a certain freetype library in).

znicholls:
Ok cool. I think I'll be able to consolidate everything in one spot. The one complication is how to test the different install options (conda + pip or pip only). At the moment we're only testing conda + pip, which I am quite happy with, but if we continue down that route I'd make the docs a touch clearer so people are aware that pip only installs might not play nice with some plotting routines. 

If you're happy, the steps I'd take would be:

1. consolidate requirements in one spot, maintaining the status quo of having tests only run under a conda install then pip install type setup
1. update docs with installation instructions and warnings about e.g. installing cartopy with pip (has never worked for anyone I've tried it with...)

gidden:
Yes, sounds good. `cartopy` and other geospatial libraries have hard requirements on `gdal` and `proj4` which are quite finicky c-level libraries that require either complicated installs or install via `conda-forge`. So a note in the docs would be quite appreciated..

znicholls:
I think the best I can get it is this (do we have dev docs you want notes like this to go in?):

- all requirements go in `setup.py`, with the geospatial libraries being specified as an optional extra
- in `./ci/`, you have the environment files which specify how conda should install things for the tests
    - there is still the awkward thing here where you have to uninstall matplotlib with conda and then reinstall with pip to get the right freetype, I don't know how to fix that and put it all in one place
    - the problem is that you have to change the matplotlib version in `.travis.yml`, not in the ci environment files but there it is (one option would be to update all the plots to remove this pip uninstall reinstall business)
- you can use the environment files in `./ci/` to make sure that the requirements specified in the [feedstock](https://github.com/conda-forge/pyam-feedstock/blob/master/recipe/meta.yaml) are up to date
    - I don't see how to make the feedstock automatically keep track with the environment files in `./ci/`, maybe it's impossible and this step always has to be done by hand (given it's a 3 minute job, probably not an issue)

gidden:
Hey @znicholls, just tried `make virtual-environment` but got 

```
$ make virtual-environment
make: Nothing to be done for 'virtual-environment'.
```

This was after trying

```
$ make test
/home/gidden/miniconda3/envs/pyam/bin/pytest tests --mpl --cov -rfsxEX --cov-report term-missing
ImportError while loading conftest '/home/gidden/work/iiasa/pyam/tests/conftest.py'.
tests/conftest.py:2: in <module>
    import matplotlib
E   ModuleNotFoundError: No module named 'matplotlib'
```

I made a fresh conda env, so assumed I should then use the environment make target. Is there something else I should be doing to check this on my machine?

znicholls:
Nope.. I'll have to try tomorrow with a fresh venv...

Can you post/email me the output of pip list and conda list with the venv
active?

Makefile
```diff
@@ -33,18 +62,37 @@ publish-on-pypi: venv
...
+	cd doc; make html
 
-venv:
-	[ -d ./venv ] || python3 -m venv venv
-	./venv/bin/pip install --upgrade pip
-	./venv/bin/pip install -e .[tests,docs,deploy]
-	touch venv
+.PHONY: virtual-environment
+virtual-environment: $(VENV_DIR)  ## make virtual environment for development
+$(VENV_DIR): setup.py ci/environment-conda-default.txt ci/environment-conda-forge.txt
```
gidden: My Make-foo is a bit rusty. I am not sure what this line is doing, but its removal leads to successfully installing an env.

gidden:
Fix (for me at least) is inline. If you have time, can you hop on slack real quick? If not no problem.

gidden:
Ok @znicholls, everything works on my machine with that line removed. I also just added a `make install` target so we get the "normal" workflow. With your permission, I will commit those changes and merge this.

gidden:
closing in favor of #216, thanks again for your effort here @znicholls. CI and test infrastructure is even better now!!

PR #216, opened 2019-04-09
[WIP] Trying to use Makefile on Travis

gidden:
**do not merge**


gidden:
Note that coverage increase is due to now seeing the correct `_version.py` file

PR #221, opened 2019-04-18
rewriting the docs of the IPCC color palette notebook

danielhuppmann:
Some minor rewording of the docstrings in the color palette notebook


gidden:
looks like timeout errors in CI, but PR looks great! I restarted CI, but if it fails again only due to timeouts, then I will merge.

PR #226, opened 2019-04-29
try to fix ci

gidden:
@danielhuppmann I think this is what we need to do to fix CI


danielhuppmann:
thanks @gidden for the assist! cherry-picked into #207 (https://github.com/IAMconsortium/pyam/pull/207/commits/e38afc543209dd1b28454150c233d620f46aefc5)

PR #229, opened 2019-04-30
[WIP] initial inclusion of operations on dataframes

gidden:

- [ ] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

This PR provides one example of how to do operations on dataframes. It is quite limited in scope, but still perhaps useful. I'd like to use it as a basis for discussion, so @danielhuppmann and @znicholls please chime in!!

The basic idea here is to be able to do basic operations, i.e., addition, subtraction, multiplication such as:

```
df = df1 - df2 # all dfs are pyam.IamDataFrames
```

I have thus far implemented a `__sub__` method. At the moment, it assumes that both dataframes included in the operation have the same model/scenario/etc. data. One operand **must** have a single variable. This is one of the pain points - we probably want to actually *intuit* which is the value being "subtracted". At the moment, I assume it is some number of variables minus some other variable. But it could also be models, or scenarios, etc.

So, as I said, this is simply a starting point for implementation and discussion. Let me know what you think!

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -115,6 +115,29 @@ class IamDataFrame(object):
     def __len__(self):
         return self.data.__len__()
 
+    def __sub__(self, other):
+        # confirm data formulated as expected
+        single_var_self = len(self.variables()) == 1
+        single_var_other = len(other.variables()) == 1
+        if  not single_var_self and not single_var_other:
+            raise ValueError('Can only subtract if one operand has a '
+                             'single variable type')
+
+        # get new variable column data
+        left = self.variables()[0] if single_var_self else self['variable']
+        right = other['variable'] if single_var_self else other.variables()[0]
+        varcol = left + ' - ' + right
+
+        # perform subtraction and return
+        df = self.copy()
+        x = df.data.set_index(GROUP_IDX).drop('variable', axis=1)
+        y = other.data.set_index(GROUP_IDX).drop('variable', axis=1)
+
+        df.data.value = (x.value - y.value).values
+        df.data.variable = varcol.values
+        return df
+        
+    
     def _execute_run_control(self):
         for module_block in run_control()['exec']:
             fname = module_block['file']
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -1,0 +1,13 @@
+import pandas as pd
+
+def test_subtract(meta_df):
+    x = meta_df.filter(variable='Primary Energy', scenario='scen_a')
+    y = meta_df.filter(variable='Primary Energy|Coal', scenario='scen_a')
+    obs = (x-y).data
+
+    exp = x.copy().data
+    exp['variable'] = 'Primary Energy - Primary Energy|Coal'
+    exp['value'] = [0.5, 3.0]
+
+    pd.testing.assert_frame_equal(obs, exp)
+


gidden:
Another option would be to have an interface like:

```
df1.subtract(df2, basis='variable')
```

But I honestly much prefer the nice mathiness of `df1 - df2` and always avoid the `subtract()`-style functions in pandas..

znicholls:
+1, my only concern would be catching all the possible edge cases if you want this to always intuit what to operate on.

danielhuppmann:
sorry for being the party-pooper, but I don't think that the elegant `df1  - df2` will fly, even if each `IamDataFrame` has only one variable - because what would the variable-name be in the new `IamDataFrame`?

So I think we will need something like `df1.subtract(df2, axis='variable', label='new_variable_name')` (`axis` being the equivalent arg from [pandas.DataFrame.substract](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.subtract.html))

gidden:
This is a good point @danielhuppmann. In the current implementation it allows for a many-to-one, one-to-many, or one-to-one variable listing based on the order of operands. The variables are then renamed to 'var1 - var2'.  I think this is the best we could do in an "intuition" style naming.

To note, we would have the same problem with `df1.subtract())`, if we want to support subtracting multiple variables from one another.

Another option - we could try to do something fancy with indexing structure + shape intuition. I.e., figure out which index shape makes sense to subtract then do the intuition that way. We still have the naming convention issue (which gets even worse if we want to support models/scenarios since they are in the meta field!).

One final thought - maybe we just forgo the naming issue altogether? If the user wants to do math, maybe they should keep track of the appropriate names? 

I don't know, I'm stuck because I want to do math with these things all the time, and it's really not obvious how best to do it. Perhaps there are other libraries we could look to for inspiration?

znicholls:
> I don't know, I'm stuck because I want to do math with these things all the time, and it's really not obvious how best to do it.

Definitely +1 for this. I always find myself filtering, pulling out dataframes, renaming variables, grouping by, operating (adding/subtracting etc.), then putting back into a new `IamDataFrame` instance. The only time I've tried doing this is in `ScmDataFrame` where I wrote `process_over` (see line 1086 of https://github.com/lewisjared/openscm/blob/scmcallib-rework/openscm/scmdataframe/base.py)

> Perhaps there are other libraries we could look to for inspiration?

This would be great, I have no idea what thought.

znicholls:
> If the user wants to do math, maybe they should keep track of the appropriate names?

I thought about this more and I think this could be the most sensible thing. If you had a method e.g. `.subtract` (which copies the pandas API) then you could do something like `.subtract(other, sub_column, new_name)` where `sub_column` is the column to subtract over (e.g. `variable`) and `new_name` is the value of `sub_column` in the output `IamDataFrame` (e.g. `Emissions|CO2|All - Fossil`). If we also added a `groupby` feature to an `IamDataFrame` then you could do stuff like 

```python
# add total - variable for all variables in a dataframe
key_var = "Emissions|CO2"
for variable, vdf in df.filter(variable=key_var, keep=False).groupby("variable"):
    df.append(
        df
        .filter(variable=key_var)
        .subtract(vdf, "variable", "{}|Total - {}".format(key_var, variable)),
        inplace=True
    )

# you could even wrap the code above into a higher level method if it was really
# that helpful
```

Or you could do something like

```python
# Find the difference between models
key_model = "MESSAGE-GLOBIOM"
for model, mdf in df.filter(model=key_model, keep=False).groupby("model"):
    df.append(
        df
        .filter(model=key_model)
        .subtract(mdf, "model", "{} - {}".format(key_model, model)),
        inplace=True
    )
```

I'm not sure that intuiting is maintainable in future because there's just too many combinations. However a combination of groupby, operations and append gives you the ability to do split-apply-combine type operations which should make most things much easier than they are now?

gidden:
Thanks for picking this back up @znicholls. At this point I'm more than willing to concede the interface discussion - if users want to use `df1 - df2` they can do it the current "hard way" (i.e., do it in pandas and re-import to pyam).

@znicholls would it be worth trying to further flesh out a concept here on a short call (maybe we can piggy back off of the AR6 call today)?

gidden:
Closing in favor of a future PR by @znicholls

PR #231, opened 2019-05-06
Remove reference to obsolete plotting.ipynb

KatharinaGruber:
plotting.ipynb was removed in #115.

PR #232, opened 2019-05-09
Remove plotting

gidden:
Following on from #231


gidden:
Thanks again @KatharinaGruber !

PR #236, opened 2019-05-14
Extend meta_df tests to include datetime

### LINKED ISSUE #222: Aggregate fails with time column
**Describe the bug**

If you make a dataframe with datetime times and call its `aggregate` method, you receive an error. This should work the same as a dataframe with year columns. 

```python
>>> from datetime import datetime
>>> 
>>> import pandas as pd
>>> 
>>> from pyam import IamDataFrame
>>> 
>>> eg_df = pd.DataFrame([
...     ['model_a', 'scen_a', 'World', 'Primary Energy|Gas', 'EJ/y', 1, 6.],
...     ['model_a', 'scen_a', 'World', 'Primary Energy|Coal', 'EJ/y', 0.5, 3],
...     ['model_a', 'scen_b', 'World', 'Primary Energy|Gas', 'EJ/y', 2, 7],
... ],
...     columns=['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010],
... )
>>> year_df = IamDataFrame(data=eg_df)
>>> year_df.timeseries()
                                                  2005  2010
model   scenario region variable            unit            
model_a scen_a   World  Primary Energy|Coal EJ/y   0.5   3.0
                        Primary Energy|Gas  EJ/y   1.0   6.0
        scen_b   World  Primary Energy|Gas  EJ/y   2.0   7.0
>>> 
>>> year_df.aggregate('Primary Energy', append=True)
>>> year_df.timeseries()
                                                  2005  2010
model   scenario region variable            unit            
model_a scen_a   World  Primary Energy      EJ/y   1.5   9.0
                        Primary Energy|Coal EJ/y   0.5   3.0
                        Primary Energy|Gas  EJ/y   1.0   6.0
        scen_b   World  Primary Energy      EJ/y   2.0   7.0
                        Primary Energy|Gas  EJ/y   2.0   7.0
>>> 
>>> dt_df = IamDataFrame(data=eg_df.rename(
...     {2005: datetime(2005, 1, 1), 2010: datetime(2010, 6, 1)},
...     axis="columns"
... ))
>>> dt_df.timeseries()
                                                  2005-01-01  2010-06-01
model   scenario region variable            unit                        
model_a scen_a   World  Primary Energy|Coal EJ/y         0.5         3.0
                        Primary Energy|Gas  EJ/y         1.0         6.0
        scen_b   World  Primary Energy|Gas  EJ/y         2.0         7.0
>>> 
>>> dt_df.aggregate('Primary Energy', append=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/zebedeenicholls/Documents/AGCEC/Misc/pyam-zn-fork/pyam/core.py", line 700, in aggregate
    df_components = _aggregate_by_variables(self.data, components, units)
  File "/Users/zebedeenicholls/Documents/AGCEC/Misc/pyam-zn-fork/pyam/core.py", line 1319, in _aggregate_by_variables
    return df.groupby(GROUP_IDX).sum()['value']
  File "/Users/zebedeenicholls/miniconda3/envs/pyam-zn-fork/lib/python3.6/site-packages/pandas/core/generic.py", line 7632, in groupby
    observed=observed, **kwargs)
  File "/Users/zebedeenicholls/miniconda3/envs/pyam-zn-fork/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 2110, in groupby
    return klass(obj, by, **kwds)
  File "/Users/zebedeenicholls/miniconda3/envs/pyam-zn-fork/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 360, in __init__
    mutated=self.mutated)
  File "/Users/zebedeenicholls/miniconda3/envs/pyam-zn-fork/lib/python3.6/site-packages/pandas/core/groupby/grouper.py", line 578, in _get_grouper
    raise KeyError(gpr)
KeyError: 'year'
```

**Failing Test**

Change `meta_df` to `test_df` in all tests in `tests/test_feature_aggregate.py`.

**System (please complete the following information):**

 - OS: macOS
 - Python and pyam version: Python 3.6.7, 0.1.2+53.gb0e6562.dirty

**Additional context**

The failure occurs because the definition of `GROUP_IDX` is static. Hence line 1319 of `pyam/core.py` looks for a `year` column and doesn't find it, hence the failure. I think there are two possible solutions here:

1. Like was done with `LONG_IDX`, move `GROUP_IDX` to being an attribute of `IamDataFrame`. It would make sense to also move `YEAR_IDX`, `REGION_IDX` and `SORT_IDX` at the same time in order to avoid these failures in other places. (I would suggest moving all indexes to being attributes of `IamDataFrame` for consistency.)
1. Drop `year` support and always use `time` columns. Given earlier discussions I think this one is unlikely to happen, but this instance is another place where it would be a sensible solution. Just for completeness, I think dropping `year` support has a couple of extra arguments in its favour. It's always possible to recover year information from a datetime. It would make the meaning of the `time` column uniform (always being a datetime) and hence force users with other `time` information to put it in more well named columns (e.g. `Summer evening` might have to go in a `season-time-of-day` column instead to be more explicit, it would be annoying to try and filter e.g. `time` based on year only to have it explode because `time` is full of strings).

znicholls:

- [x] Tests Added

```
```

# Description of PR

Closes #222. Good news is this was already sorted in v0.2.0 so all this does is make sure it stays that way by expanding the tests a bit.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -323,7 +323,7 @@ class IamDataFrame(object):
             "time" is not a column of `self.data`
         """
         if "time" not in self.data:
-            raise ValueError("`time` must be in `self.data` to use this method")
+            raise ValueError("time column must be datetime to use this method")
 
         ret = copy.deepcopy(self) if not inplace else self
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -309,6 +309,31 @@ class IamDataFrame(object):
         fill_values['year'] = year
         self.data = self.data.append(fill_values, ignore_index=True)
 
+    def swap_time_for_year(self, inplace=False):
+        """Convert the `time` column to `year`.
+
+        Parameters
+        ----------
+        inplace: bool, default False
+            if True, do operation inplace and return None
+
+        Raises
+        ------
+        ValueError
+            "time" is not a column of `self.data`
+        """
+        if "time" not in self.data:
+            raise ValueError("`time` must be in `self.data` to use this method")
+
+        ret = copy.deepcopy(self) if not inplace else self
+
+        ret.data["year"] = ret.data["time"].apply(lambda x: x.year)
+        ret.data = ret.data.drop("time", axis="columns")
+        ret._LONG_IDX = [v if v != "time" else "year" for v in ret._LONG_IDX]
+
+        if not inplace:
+            return ret
+
     def as_pandas(self, with_metadata=False):
         """Return this as a pd.DataFrame
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -331,6 +331,11 @@ class IamDataFrame(object):
         ret.data = ret.data.drop("time", axis="columns")
         ret._LONG_IDX = [v if v != "time" else "year" for v in ret._LONG_IDX]
 
+        if any(ret.data[ret._LONG_IDX].duplicated()):
+            error_msg = ('swapping time for year will result in duplicate '
+                         'rows in `data`!')
+            raise ValueError(error_msg)
+
         if not inplace:
             return ret
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -325,7 +325,7 @@ class IamDataFrame(object):
         if "time" not in self.data:
             raise ValueError("time column must be datetime to use this method")
 
-        ret = copy.deepcopy(self) if not inplace else self
+        ret = self.copy() if not inplace else self
 
         ret.data["year"] = ret.data["time"].apply(lambda x: x.year)
         ret.data = ret.data.drop("time", axis="columns")
@@ -1388,8 +1388,8 @@ def _check_rows(rows, check, in_range=True, return_test='any'):
         msg = 'Unknown checking type: {}'
         raise ValueError(msg.format(check.keys() - valid_checks))
 
-    rows = rows.copy()
     if 'year' not in rows:
+        rows = rows.copy()
         rows['year'] = rows['time'].apply(lambda x: x.year)
 
     where_idx = set(rows.index[rows['year'] == check['year']]) \
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1383,6 +1383,10 @@ def _check_rows(rows, check, in_range=True, return_test='any'):
         msg = 'Unknown checking type: {}'
         raise ValueError(msg.format(check.keys() - valid_checks))
 
+    rows = rows.copy()
+    if 'year' not in rows:
+        rows['year'] = rows['time'].apply(lambda x: x.year)
+
     where_idx = set(rows.index[rows['year'] == check['year']]) \
         if 'year' in check else set(rows.index)
     rows = rows.loc[list(where_idx)]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -167,15 +167,15 @@ TEST_STACKPLOT_DF['model'] = 'IMG'
 TEST_STACKPLOT_DF['scenario'] = 'a_scen'
 
 
-TIME_AXES = [
-    [2005, 2010],
-    [datetime(2005, 6, 17), datetime(2010, 7, 21)],
-    ['2005-06-17', '2010-07-21'],
-    ['2005-06-17 00:00:00', '2010-07-21 12:00:00']
-]
-
-
-@pytest.fixture(scope="function", params=TIME_AXES)
+@pytest.fixture(
+    scope="function",
+    params=[
+        [2005, 2010],
+        [datetime(2005, 6, 17), datetime(2010, 7, 21)],
+        ['2005-06-17', '2010-07-21'],
+        ['2005-06-17 00:00:00', '2010-07-21 12:00:00']
+    ]
+)
 def test_df(request):
     tdf = TEST_DF.iloc[:2]
     tdf = tdf.rename({2005: request.param[0], 2010: request.param[1]},
@@ -195,9 +195,17 @@ def test_pd_df():
     yield TEST_DF.copy()
 
 
-@pytest.fixture(scope="function")
-def meta_df():
-    df = IamDataFrame(data=TEST_DF)
+@pytest.fixture(
+    scope="function",
+    params=[
+        [2005, 2010],
+        [datetime(2005, 6, 17), datetime(2010, 7, 21)],
+    ]
+)
+def meta_df(request):
+    mdf = TEST_DF.rename({2005: request.param[0], 2010: request.param[1]},
+                         axis="columns")
+    df = IamDataFrame(data=mdf)
     yield df
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -167,11 +167,14 @@ TEST_STACKPLOT_DF['model'] = 'IMG'
 TEST_STACKPLOT_DF['scenario'] = 'a_scen'
 
 
+TEST_YEARS = [2005, 2010]
+TEST_DTS = [datetime(2005, 6, 17), datetime(2010, 7, 21)]
+
 @pytest.fixture(
     scope="function",
     params=[
-        [2005, 2010],
-        [datetime(2005, 6, 17), datetime(2010, 7, 21)],
+        TEST_YEARS,
+        TEST_DTS,
         ['2005-06-17', '2010-07-21'],
         ['2005-06-17 00:00:00', '2010-07-21 12:00:00']
     ]
@@ -198,8 +201,8 @@ def test_pd_df():
 @pytest.fixture(
     scope="function",
     params=[
-        [2005, 2010],
-        [datetime(2005, 6, 17), datetime(2010, 7, 21)],
+        TEST_YEARS,
+        TEST_DTS,
     ]
 )
 def meta_df(request):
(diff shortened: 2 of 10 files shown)


pyam/core.py
```diff
@@ -309,6 +309,31 @@ def interpolate(self, year):
...
+        ------
+        ValueError
+            "time" is not a column of `self.data`
+        """
+        if "time" not in self.data:
+            raise ValueError("time column must be datetime to use this method")
+
+        ret = copy.deepcopy(self) if not inplace else self
+
+        ret.data["year"] = ret.data["time"].apply(lambda x: x.year)
```
danielhuppmann: what would happen here if you have two `time` entries for the same year - if I understand this correctly, the resulting `IamDataFrame` would then have a non-unique `year` in `data` for the same model/scenario/variable/region?

This does not cause an issue in the new unit tests, but maybe it makes sense to add a sanity check here similar to https://github.com/IAMconsortium/pyam/blob/25aa7f09fb49ed64569255dd2bf260fb8e62f717/pyam/utils.py#L252

tests/test_feature_append_rename_convert.py
```diff
@@ -160,15 +166,20 @@ def test_rename_append(meta_df):
     obs = meta_df.rename(mapping, append=True)
 
     # test data changes
+    dts = [dt.datetime(2005, 6, 17), dt.datetime(2010, 7, 21)]
+    times = [2005, 2010] if "year" in meta_df.data else dts
```
danielhuppmann: why double-quotes?

tests/conftest.py
```diff
@@ -195,9 +195,17 @@ def test_pd_df():
...
 
 
-@pytest.fixture(scope="function")
-def meta_df():
-    df = IamDataFrame(data=TEST_DF)
+@pytest.fixture(
+    scope="function",
+    params=[
+        [2005, 2010],
+        [datetime(2005, 6, 17), datetime(2010, 7, 21)],
```
danielhuppmann: this list of `datetimes` is used quite a lot now throughout the tests - maybe make that available as static object and reuse throughout?

tests/test_cast_to_iamc.py
```diff
@@ -58,17 +67,21 @@ def test_cast_with_model_arg(meta_df):
 
 
 def test_cast_by_column_concat(meta_df):
+    dts = [dt.datetime(2005, 6, 17), dt.datetime(2010, 7, 21)]
```
danielhuppmann: reuse from `test/conftest.py`

tests/test_core.py
```diff
@@ -518,22 +518,39 @@ def test_validate_up(meta_df):
...
 
 def test_validate_both(meta_df):
     obs = meta_df.validate({'Primary Energy': {'up': 6.5, 'lo': 2.0}})
     assert len(obs) == 2
-    assert list(obs['year'].values) == [2005, 2010]
+    if 'year' in meta_df.data:
+        assert list(obs['year'].values) == [2005, 2010]
+    else:
+        exp_time = pd.to_datetime([
+            datetime.datetime(2005, 6, 17),
```
danielhuppmann: reuse from `test/conftest.py`

tests/test_feature_aggregate.py
```diff
@@ -9,13 +12,19 @@ def test_do_aggregate_append(meta_df):
     meta_df.aggregate('Primary Energy', append=True)
     obs = meta_df.filter(variable='Primary Energy').timeseries()
 
+    dts = [dt.datetime(2005, 6, 17), dt.datetime(2010, 7, 21)]
```
danielhuppmann: reuse from `test/conftest.py`

tests/test_feature_append_rename_convert.py
```diff
@@ -136,13 +137,18 @@ def test_rename_index(meta_df):
     obs = meta_df.rename(mapping, scenario={'scen_a': 'scen_c'})
 
     # test data changes
+    dts = [dt.datetime(2005, 6, 17), dt.datetime(2010, 7, 21)]
```
danielhuppmann: reuse from `test/conftest.py`

tests/test_feature_compare.py
```diff
@@ -13,17 +16,23 @@ def test_compare(meta_df):
...
-        columns=['variable', 'unit', 'year', 'meta_df', 'clone'],
+        columns=['variable', 'unit', 'time', 'meta_df', 'clone'],
     )
     exp['model'] = 'model_a'
     exp['scenario'] = 'scen_a'
     exp['region'] = 'World'
-    exp = exp.set_index(IAMC_IDX + ['year'])
+    time_col = 'time'
+    if "year" in meta_df.data.columns:
+        exp["year"] = exp["time"].apply(lambda x: x.year)
```
danielhuppmann: use the new function `swap_time_for_year()`?

pyam/core.py
znicholls: Nice will do

tests/test_feature_append_rename_convert.py
znicholls: habit, have changed

tests/test_feature_compare.py
znicholls: can't do that here as `exp` has nan's so can't be cast to `IamDataFrame`

tests/conftest.py
znicholls: nice will do!

tests/test_feature_compare.py
danielhuppmann: got it, thanks!

znicholls:
cheers @danielhuppmann, that should all be addressed now!

znicholls:
not sure what's going on with CI, read through https://github.com/pyproj4/pyproj/issues/134 but am none the wiser...

danielhuppmann:
sorry about this - @gidden and I already discussed that the region-plotting-tests are quite unstable. How about deactivating those for Windows machines?

gidden:
Hi folks - indeed these are our most finicky tests. One suggestion before turning them off for windows.

In pyproj4/pyproj#134, someone notes that the issue goes away for pyproj>2.1.

On appveyor, I see that the version is:
```
pyproj: 1.9.6
```

So perhaps we can try adding the minimum version to our appveyor install? 

If that still fails, then I'm ok with just turning them off on windows..

gidden:
Hey @znicholls, I'm now digging through more geopandas/cartopy/pyproj issues. I have two thoughts.

First is related to geopandas/geopandas#830, which suggests a specific pyproj version for windows. To attempt this, can you please revert the update in ` ci/environment-conda-forge.txt` and try pinning *only* the appveyor install to `"pyproj=1.9.4"`?

Second, I'm looking through our last successful CI on windows: https://ci.appveyor.com/project/gidden/pyam/builds/24523793/job/2idsdm0hhgvg2tfw

In reviewing the package versions at runtime, there is no difference between that run and 60ccc9a. In reviewing the installed conda packages, however, I see the following difference:

Successful run:

```
proj4                     5.2.0                ha925a31_1 
```

Failed run:
```
proj4                     5.2.0             h6538335_1002    conda-forge
```

Thus this implies one potential fix is to install `proj4` via the first `conda install` (without `-c conda-forge`).

I know this is annoying, but if you would be willing to try those two changes, that would be amazing. Otherwise, let's just turn this stuff off for windows.

gidden:
I don't know whether to be happy or sad, but it looks fixed now!

pyam/core.py
```diff
@@ -309,6 +309,36 @@ def interpolate(self, year):
...
+
+        Raises
+        ------
+        ValueError
+            "time" is not a column of `self.data`
+        """
+        if "time" not in self.data:
+            raise ValueError("time column must be datetime to use this method")
+
+        ret = copy.deepcopy(self) if not inplace else self
```
gidden: should this now simply be `self.copy()` since we have that method?

pyam/core.py
```diff
@@ -1358,6 +1388,10 @@ def _check_rows(rows, check, in_range=True, return_test='any'):
         msg = 'Unknown checking type: {}'
         raise ValueError(msg.format(check.keys() - valid_checks))
 
+    rows = rows.copy()
```
gidden: should this copy move into the `if`statement? otherwise we might be doing a copy unnecessarily?

gidden:
A few inline comments, then this is g2g for me. @danielhuppmann you still have this blocking in review, so let us know if it passes muster!

pyam/core.py
znicholls: It's to ensure you don't accidentally add  a `year` column to your dataframe

gidden: But wouldn't that only happen if you get inside the `if` statement? Suggestion would be:

```
    if 'year' not in rows:
        rows = rows.copy()
        rows['year'] = rows['time'].apply(lambda x: x.year)
```

gidden: Or perhaps I am confused. Sorry if that's the case!

znicholls: ah yep I get it, nice!

gidden:
Looks great! I'll let @danielhuppmann pull this one in.

Issue #238, opened 2019-05-28
Add method to set meta indicators from data


danielhuppmann:
It would be useful to have a generic `df.set_meta_from_data()` function, e.g.,
```
df.set_meta_from_data(name, method, column='value', **kwargs)
```
where 
 - `kwargs` can be additional filters (for example apply only where `variable='Temperature'` or `region='World'`)
 - `method` can be min, max, mean, median, and maybe even other `pyam` functions like `pyam.timeseries.cross_threshold()` or `cumulative()`
 - `column` would be what to set as meta - the actual `value` (default) or the `year`/`region`/... column where the value is reached (e.g., year of peak temperature)
 - `name` is the column name in `meta` (similar to the use in `df.set_meta()`)

So the use case suggested by @znicholls for setting the peak temperature (see #237) could be implemented as
```
df.set_meta_from_data(variable='Temperature', name='Peak temperature', method='max')
```

danielhuppmann:
closed via #241

Issue #237, opened 2019-05-23
Adding peak temperature scenario metadata


znicholls:
@gidden and @danielhuppmann is there a function that allows the user to easily add peak temperature for each scenario as a metadata column of an `IamDataFrame`? I'm thinking something like, `df.add_peak_temperature_meta()`.  

Maybe it's currently a two-line solution, something like the below?

```python
mdf = df.data.groupby(["model", "scenario"]).max("Temperature")
df.set_meta(mdf)
```

Sorry if this is already in there and should be obvious, I couldn't work it out. If it's not in there, I'm happy for you to assign this one to me (I won't get to it straight away but will eventually).

danielhuppmann:
Indeed, this use case already popped up in the SR15, in parictular notebook `sr15_2.0_categories_indicators.ipynb` at https://github.com/iiasa/ipcc_sr15_scenario_analysis.

Below, a few lines of code copy-pasted from that notebook - a bit more elaborate than your suggestion because it did a lot more things, like also adding a meta-column in which year peak temperature was reached, and doing that for multiple climate models.

```
median_warming = 'AR5 climate diagnostics|Temperature|Global Mean|MAGICC6|MED'

def peak_warming(x, return_year=False):
    peak = x[x == x.max()]
    if return_year:
        return peak.index[0]
    else:
        return float(max(peak))

median_temperature = sr1p5.filter(variable=median_warming).timeseries()
name = 'median warming at peak (MAGICC6)'
sr1p5.set_meta(median_temperature.apply(peak_warming, raw=False, axis=1), name)
```

I'm a bit hesitant about the very concise `df.add_peak_temperature_meta()`, because the exact variable string from which to derive that info will be application-specific - or which climate model to use.

But a more generic `df.set_meta_from_data()` might be useful, e.g., 
```
df.set_meta_from_data(name, method, column='value', **kwargs)
```
where 
 - `kwargs` can be additional filters (for example apply only where `variable='Temperature'` or `region='World'`)
 - `method` can be min, max, mean, median, and maybe even other `pyam` functions like `pyam.timeseries.cross_threshold()` or `cumulative()`
 - `column` would be what to set as meta - the actual `value` (default) or the `year`/`region`/... column where the value is reached (e.g., year of peak temperature)
 - `name` is the column name in `meta` (similar to the use in `df.set_meta()`)

znicholls:
Nice that looks very elegant!

gidden:
[...]

PR #241, opened 2019-07-04
Add `set_meta_from_data()` feature

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR implements the feature requested by @znicholls (see #238 and #237) to derive meta indicators directly from the `data` dataframe, with a `method` arg passed for aggregation (e.g., return the year where the temperature maximum is reached).

# Discussion item

It might be more user-friendly to allow strings as `method` arg in addition to functions, i.e., `method='max'` rather than `method=`np.max`. Any thoughts?

Closes 238


danielhuppmann:
@znicholls, can I please get a hug for writing a test that also ensures it works with `datetime`? ;-)

danielhuppmann:
unit tests are failing because of issues with the region-plotting, not the new feature and related tests

for the review, it makes most sense to only look at commit https://github.com/IAMconsortium/pyam/pull/241/commits/f5151892cb760c79eb6b7e6848a225600546ec7c, the commits before that are just clean-up and moving all tests related to setting meta indicators to an own file

znicholls:
I send you all the hugs! I’ll do a proper review Monday, for now my only thought is whether applying `method` with `pd.groupby.apply(method)` rather than `method(pd.groupby)` would provide more flexibility.

pyam/core.py
```diff
@@ -458,6 +458,27 @@ def set_meta(self, meta, name=None, index=None):
...
+            unique values (e.g., `numpy.max()`)
+        column: str, optional
+            the column from `data` to be used for the meta indicator
+        kwargs: passed to :meth:`IamDataFrame.filter()` for downselected `data`
+        """
+        _data = self.filter(**kwargs).data
+        if method is None:
+            meta = _data.set_index(META_IDX)[column]
+        else:
+            meta = method(_data.groupby(META_IDX))[column]
```
```suggestion
            meta = _data.groupby(META_IDX).apply(method)[column]
```

alternate way which might provide more flexibility

gidden: Perhaps even `meta = _data.groupby(META_IDX)[column].apply(method)`?

gidden:
hey folks, trying out now a fix for CI (hopefully permanent..). Regarding how to pass the `method` arg, I'm a bit torn here, as I can see arguments in both cases (passing a function or string). One thing that is nice about using a string is that it plays well with more on-disk usage (e.g., writing a config file, being able to pickle/serialize objects, etc.).  However, it seems like the work is done and is quite straightforward for using `apply`.

Perhaps we can leave the question open and potentially support both in the future. In that case, I would change the kwarg from `method` to `apply`. This provides clarity to the user (i.e., `np.max` will be used by `df.apply`). It also leaves us the option of in the future providing a more generic `method` arg, which would simply take a dataframe as argument.

gidden:
Hi folks - CI has been updated on the master branch. Please rebase this one on top!

gidden:
Restarted the build, should bring in updates from master

pyam/core.py

danielhuppmann:
rebased and refactored to `data.apply(method)` as suggested by @znicholls (though this does still not allow to use `np.mean()` as method because this doesn't work on `pandas.groupby`)

znicholls:
> though this does still not allow to use `np.mean()` as method because this doesn't work on `pandas.groupby`

this test passes ?

```python
def test_set_meta_from_data_mean(meta_df):
    meta_df.set_meta_from_data('pe_mean', variable='Primary Energy', method=np.mean)
    exp = pd.Series(data=[3.5, 4.5], index=EXP_IDX, name='pe_mean')
    pd.testing.assert_series_equal(meta_df['pe_mean'], exp)
```

danielhuppmann:
> this test passes ?

Seems to pass now - no idea why it failed yesterday when I tried it...

znicholls:
@gidden lgtm so over to you for merging

gidden:
LGTM, should we add this feature in a tutorial somewhere?

danielhuppmann:
added a section in the 'first-steps' tutorial per suggestion by @gidden

danielhuppmann:
reopening to get pandas 0.25.0 hotfix after merging #246

gidden:
Thanks @danielhuppmann !

PR #243, opened 2019-07-08
Conn auth

gidden:

- [x] Tests Added
- [x] Documentation Added

This PR adds the ability to connect to non-public databases by providing a username and password.

TODO: 

- [x] this currently also requires knowledge of an "application" name. For example, for the SR1.5 explorer, the "short name" is `iamc15` and the "application name" is `IXSE_SR15`.  We should decide which to go with in the end. To note, utilizing the "short name" will require an additional mapping in the python code (to the best of my knowledge), which was seen as desirable to avoid. In any case, I think we should limit the user to needing to know **only** one name.
- [x] decide on how to pass over credentials. Options: 1) leave as is and it is up to the user to store them properly. 2) support reading/writing from a config file. 3) support passing over env variables. 4) both 2 & 3
- [x] add functionality to check success of requests (likely as a context manager)


gidden:
@byers if you want to get started working on AR6, you can use this branch to do so. You'll need to coordinate the naming conventions with @zikolach and @fonfon

pyam/iiasa.py
```diff
@@ -37,31 +47,65 @@ def valid_connection_names():
...
+            headers = {'Accept': 'application/json',
+                       'Content-Type': 'application/json'}
+            app = application or _LOGIN_APP_NAMES[name]
+            data = {'username': user, 'password': pw, 'application': app}
+            response = requests.post(
+                _LOGIN_URL, headers=headers, data=json.dumps(data)
+            )
+            self._auth = response.json()
+            if not response.ok:  # auth failed, provide user a nice message
+                msg = 'Login to {} failed with given credentials: {}'
```
fonfon: I wouldn't output the credentials, at least not the password

fonfon:
For the application names I'd prefer to fetch them from the auth-server and see how that goes. If we hard-code a lookup dict here it will become tedious (or forgotten) to keep it up to date, even if the application names do not change very often.
Users would then have to provide the application name though, like listed on https://data.ene.iiasa.ac.at/auth/#/apps

pyam/iiasa.py
gidden: Ah here, the creds aren't printed, only the app name and error log.

gidden: Where `self._auth` is the result of a failed `requests.post()`

pyam/iiasa.py
```diff
@@ -37,31 +47,65 @@ def valid_connection_names():
...
+            self._auth = requests.get(url).json()
+        else:  # read credentials from kwargs
+            try:
+                if isinstance(creds, collections.Mapping):
+                    user, pw = creds['username'], creds['password']
+                else:
+                    user, pw = creds
+            except Exception as e:
+                msg = 'Could not read credentials: {}\n{}'.format(
+                    creds, str(e))
```
gidden: @fonfon, here we do print `creds`. However, this is done because the input is malformed (i.e., not a dict with appropriate keys or a 2-tuple). What's the tradeoff here between presenting the user with an exact description of the problem and potentially providing back a password as plaintext?

gidden:
Ok @fonfon, happy to make this more amenable to the existing REST API features. Let's discuss when we're both in the office/on slack.

gidden:
Here is the example how users can currently interact with the `pyam.iiasa.Connection` class: https://pyam-iamc.readthedocs.io/en/stable/tutorials/iiasa_dbs.html

byersiiasa:
> @byers if you want to get started working on AR6, you can use this branch to do so. You'll need to coordinate the naming conventions with @zikolach and @fonfon

thanks I'll try this today - for some reason I didn't get a notification

gidden:
Note that this needs iiasa/auth_server#53 to continue further

gidden:
Hi folks, from my perspective this PR is now complete and ready for final review. @fonfon @byersiiasa and/or @danielhuppmann would you mind taking a final look?

pyam/iiasa.py
```diff
@@ -17,51 +19,110 @@
...
-    'iamc15': 'D. Huppmann, E. Kriegler, V. Krey, K. Riahi, '
-    'J. Rogelj, S.K. Rose, J. Weyant, et al., '
-    'IAMC 1.5C Scenario Explorer and Data hosted by IIASA. '
-    'IIASA & IAMC, 2018. '
-    'doi: 10.22022/SR15/08-2018.15429, '
-    'url: data.ene.iiasa.ac.at/iamc-1.5c-explorer'
-}
 
+def _get_token(creds):
+    # get anonymous auth if creds
```
danielhuppmann: should this be "if creds is not defined"

pyam/iiasa.py
```diff
@@ -17,51 +19,110 @@
...
+        self._connected = None
+        if name:
+            self.connect(name)
+
+    @property
+    @lru_cache()
+    def valid_connections(self):
+        url = '/'.join([_BASE_URL, 'applications'])
+        headers = {'Authorization': 'Bearer {}'.format(self._token)}
+        response = requests.get(url, headers=headers).json()
```
zikolach: would it be better to check response status here?

pyam/iiasa.py
```diff
@@ -17,51 +19,110 @@
...
+            raise ValueError(msg.format(name, valid))
 
-        logger().info(
-            'You are connected to the {} {}. Please cite as:\n\n{}'
-            .format(name, 'scenario explorer', _CITATIONS[name])
-        )
+        url = '/'.join([_BASE_URL, 'applications', name, 'config'])
+        headers = {'Authorization': 'Bearer {}'.format(self._token)}
+        response = requests.get(url, headers=headers).json()
+        idxs = {x['path']: i for i, x in enumerate(response)}
```
zikolach: Maybe I misunderstand how it works, but the response should contain json similar to following one
```
[
    {
        "name": "IXSE_SETNAV",
        "scheme": "IXMP Scenario Explorer SPA UI",
        "config": [
            {
                "path": "productName",
                "type": "STRING",
                "value": "SET-Nav Scenario Explorer"
            },
            {
                "path": "uiUrl",
                "type": "STRING",
                "value": "https://data.ene.iiasa.ac.at/set-nav"
            }
        ]
    },
    {
        "name": "IXSE_SR15",
        "scheme": "IXMP Scenario Explorer SPA UI",
        "config": [
            {
                "path": "productName",
                "type": "STRING",
                "value": "IAMC 1.5°C Scenario Explorer hosted by IIASA"
            },
            {
                "path": "uiUrl",
                "type": "STRING",
                "value": "https://data.ene.iiasa.ac.at/iamc-1.5c-explorer"
            }
        ]
    },
    {
        "name": "IXSE_SR15_QA",
        "scheme": "IXMP Scenario Explorer SPA UI",
        "config": [
            {
                "path": "productName",
                "type": "STRING",
                "value": "IAMC 1.5°C Scenario Explorer hosted by IIASA (QA)"
            },
            {
                "path": "uiUrl",
                "type": "STRING",
                "value": "https://data.ene.iiasa.ac.at/IXMPtest/scenario-explorer-qa"
            }
        ]
    },
    {
        "name": "IXSE_TEST_PUBLIC",
        "scheme": "IXMP Scenario Explorer SPA UI",
        "config": [
            {
                "path": "productName",
                "type": "STRING",
                "value": "Scenario Explorer (sandbox)"
            },
            {
                "path": "uiUrl",
                "type": "STRING",
                "value": "https://data.ene.iiasa.ac.at/ixmp-explorer-sandbox"
            }
        ]
    }
]
```
How does it select specific app?

zikolach: Sorry, I overlook name in `        url = '/'.join([_BASE_URL, 'applications', name, 'config'])`

pyam/iiasa.py
gidden: Yup agreed, will update

pyam/iiasa.py
gidden: Yes, probably so. I can write a convenience function to do this for all requests.

pyam/iiasa.py
gidden: Yeah, this queries information after `connect()` is called, which aligns it to a specific application

gidden:
@danielhuppmann @zikolach @fonfon (and @khaeru if you're interested)

I have listed a set of TODOs in the PR summary. Could you please provide input/opinions on items 1 and 2?

Namely, for credentials, I see a lot of potential problems here. E.g., if we want to support environment variables explicitly, once they are set, a user can no longer get explicitly to public dbs as "anonymous", because any sort of default/fall-back will always see them. Config files are less problematic, but is it our job in this library to support this? A user can always write their own file and read it in however they wish...

On the naming, I have no super strong opinions and would be happy if y'all could come to a consensus that I can implement.

gidden:
Hi all, in bea472e I have added the ability to pass a credentials file with a warning if a non-file based option is used. I think that should cover the issues here.

gidden:
Ok all, to my eye this PR is now complete. Please let me know the resolution of point 1 between yourselves.

byersiiasa:
Hi
I have just tested (limited) and seems to work well. In particular - tested it that when credentials are supplied, then more instances are available... nice!

```
# Connection with instance / creds
Connection(creds=('xxxx','XXXXX')).valid_connections
```
returns
`['IXSE_AR6', 'IXSE_SETNAV', 'IXSE_SR15', 'IXSE_SR15_QA', 'IXSE_TEST_PUBLIC']`

```
# Connection without instance / creds
 Connection()
```
returns
`['IXSE_SETNAV', 'IXSE_SR15', 'IXSE_SR15_QA', 'IXSE_TEST_PUBLIC']`


Thanks all!

gidden:
Hi @danielhuppmann @zikolach and @fonfon would be great to try to come to some resolution on the point above so this can be merged. Thanks!

gidden:
Hi @danielhuppmann @zikolach and @fonfon. This has been an open for ~2 weeks now. It would be great to get a response on point 1 above. If there is no resolution in a reasonable timeframe, I suggest we pull this as-is unless someone wants to object and review!

zikolach:
@gidden fyi I have no access to mark this PR as accepted despite the fact I included in reviewers list

gidden:
Just added you as a collaborator!

zikolach:
> Just added you as a collaborator!

@gidden half way there - now I can squash and merge but not accept/reject ;)

byers:
Hello,

I think someone included Github id reference @byers instead of @byersiiasa
in a pull request comment. This might explain why he's not getting
notifications of comments, etc. since I am ;-).

It looks like an interesting project, but i'm pretty busy with a startup
and other projects, so please remove me from your list (@byersiiasa instead
of @byers)

https://github.com/byersiiasa

Thanks for your time,

Byers

danielhuppmann:
Thanks @gidden and @zikolach for the effort!

Let's use the "application names" and maybe add support for "short names" on the authentication API at a later stage.

PR #244, opened 2019-07-18
[WIP] Region aggregation mip feature

peterkolp:

- [x] Tests Added
- [ ] Documentation Added

```
- (#244)[https://github.com/IAMconsortium/pyam/pull/244] Added feature which does something
```

# Description of PR

Enhance  **aggregate_region** - add  (min, max, avg) aggregation methods.
Add **weighted_average_region** 
```
    def aggregate_region(self, variable, region='World', subregions=None,
                         components=None, append=False, method='sum'):

   def weighted_average_region(self, variable, weight, region='World',
                                subregions=None, append=False):

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,7 @@
 
 ## Individual Updates
 
+- [#244](https://github.com/IAMconsortium/pyam/pull/244) add aggregation methods (min/max/avg) to aggregate_ragion and add eighted_average_region function
 - [#228](https://github.com/IAMconsortium/pyam/pull/228) Update development environment creation instructions and make pandas requirement more specific
 - [#219](https://github.com/IAMconsortium/pyam/pull/219) Add ability to query metadata from iiasa data sources
 - [#214](https://github.com/IAMconsortium/pyam/pull/214) Tidy up requirements specifications a little
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -783,30 +783,22 @@ class IamDataFrame(object):
 
             return IamDataFrame(diff, variable=variable).timeseries()
 
-    def aggregate_region_by_weight(self, variable, region='World',
-                                   subregions=None, components=None,
-                                   append=False, weight_var=None):
-        """Compute the aggregate of timeseries over a number of regions
-         using weighted average (by weight_var) method
+    def aggregate_region_by_weight(self, variable, weight, region='World',
+                                   subregions=None, append=False):
+        """Compute weighted average of a timeseries over multiple regions
          weight_var: str,
             weighting variable to be used for calculating weighted average
             is case aggregator is 'w.avg'
         """
         subregion_df = self.filter(region=subregions)
         cols = ['region', 'variable']
-        if weight_var:
-            weightvar_df = subregion_df.filter(variable=weight_var).data
-        else:
-            weightvar_df = None
-        #
-        if not weight_var:
-            msg = 'cannot aggregate because no weight_var arg provided'
-            logger().error(msg)
-            return
-        elif weightvar_df.empty:
+        
+        weightvar_df = subregion_df.filter(variable=weight).data
+
+        if weightvar_df.empty:
             msg = "cannot aggregate because no data found "\
                 "or weight_var='{}'"
-            logger().error(msg.format(weight_var))
+            logger().error(msg.format(weight))
             return
         var_df = subregion_df.filter(variable=variable).data
         # _data = _aggregate(var_df, cols,
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -713,7 +713,7 @@ class IamDataFrame(object):
         if not inplace:
             return ret
 
-    def aggregate(self, variable, components=None, append=False):
+    def aggregate(self, variable, components=None, method='sum', append=False):
         """Compute the aggregate of timeseries components or sub-categories
 
         Parameters
@@ -722,6 +722,8 @@ class IamDataFrame(object):
             variable for which the aggregate should be computed
         components: list of str, default None
             list of variables, defaults to all sub-categories of `variable`
+        method: func or str
+            method to use for aggregation
         append: bool, default False
             append the aggregate timeseries to `data` and return None,
             else return aggregate timeseries
@@ -736,15 +738,15 @@ class IamDataFrame(object):
             return
 
         rows = self._apply_filters(variable=components)
-        _data = _aggregate(self.data[rows], 'variable')
+        _data = _aggregate(self.data[rows], 'variable', method)
 
         if append is True:
             self.append(_data, variable=variable, inplace=True)
         else:
             return _data
 
-    def check_aggregate(self, variable, components=None, exclude_on_fail=False,
-                        multiplier=1, **kwargs):
+    def check_aggregate(self, variable, components=None, method='sum',
+                        exclude_on_fail=False, multiplier=1, **kwargs):
         """Check whether a timeseries matches the aggregation of its components
 
         Parameters
@@ -753,6 +755,8 @@ class IamDataFrame(object):
             variable to be checked for matching aggregation of sub-categories
         components: list of str, default None
             list of variables, defaults to all sub-categories of `variable`
+        method: func or str
+            method to use for aggregation
         exclude_on_fail: boolean, default False
             flag scenarios failing validation as `exclude: True`
         multiplier: number, default 1
@@ -767,7 +771,8 @@ class IamDataFrame(object):
         # filter and groupby data, use `pd.Series.align` for matching index
         rows = self._apply_filters(variable=variable)
         df_variable, df_components = (
-            _aggregate(self.data[rows], 'variable').align(df_components)
+            _aggregate(self.data[rows], 'variable', method)
+            .align(df_components)
         )
 
         # use `np.isclose` for checking match
@@ -822,8 +827,7 @@ class IamDataFrame(object):
             return _data
 
     def aggregate_region(self, variable, region='World', subregions=None,
-                         components=None, append=False,
-                         method='sum'):
+                         components=None, append=False):
         """Compute the aggregate of timeseries over a number of regions
         including variable components only defined at the `region` level
         Parameters
@@ -840,8 +844,6 @@ class IamDataFrame(object):
         append: bool, default False
             append the aggregate timeseries to `data` and return None,
             else return aggregate timeseries
-        method: str, default 'sum'
-            aggregator function. One of 'sum', 'avg', 'min', 'max'
         """
         # default subregions to all regions other than `region`
         subregions = subregions or self._all_other_regions(region, variable)
@@ -852,17 +854,11 @@ class IamDataFrame(object):
             logger().info(msg.format(variable, region))
 
             return
-        known_methods = ['sum', 'avg', 'min', 'max']
-        if method not in known_methods:
-            msg = 'cannot aggregate because method=`{}` is unknown'\
-                  ' i.e. not in `{}`'
-            logger().error(msg.format(method, known_methods))
-            return
+
         # compute aggregate over all subregions
         subregion_df = self.filter(region=subregions)
         cols = ['region', 'variable']
-        _data = _aggregate(subregion_df.filter(variable=variable).data, cols,
-                           method)
+        _data = _aggregate(subregion_df.filter(variable=variable).data, cols)
 
         # add components at the `region` level, defaults to all variables one
         # level below `variable` that are only present in `region`
@@ -1412,7 +1408,7 @@ def _meta_idx(data):
     return data[META_IDX].drop_duplicates().set_index(META_IDX).index
 
 
-def _aggregate(df, by, method):
+def _aggregate(df, by, method=np.sum):
     """Aggregate `df` by specified column(s), return indexed `pd.Series`"""
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -791,9 +791,20 @@ class IamDataFrame(object):
     def aggregate_region_by_weight(self, variable, weight, region='World',
                                    subregions=None, append=False):
         """Compute weighted average of a timeseries over multiple regions
-         weight_var: str,
+
+        Parameters
+        ----------
+        variable: str
+            variable for which the aggregate should be computed
+         weight: str
             weighting variable to be used for calculating weighted average
-            is case aggregator is 'w.avg'
+        region: str, default 'World'
+            dimension
+        subregions: list of str
+            list of subregions, defaults to all regions other than `region`
+        append: bool, default False
+            append the aggregate timeseries to `data` and return None,
+            else return aggregate timeseries
         """
         # default subregions to all regions other than `region`
         subregions = subregions or self._all_other_regions(region, variable)
@@ -829,7 +840,9 @@ class IamDataFrame(object):
     def aggregate_region(self, variable, region='World', subregions=None,
                          components=None, append=False):
         """Compute the aggregate of timeseries over a number of regions
-        including variable components only defined at the `region` level
+        
+        This function adds `components` only defined at the `region` level
+
         Parameters
         ----------
         variable: str
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -38,6 +38,7 @@ from pyam.utils import (
     YEAR_IDX,
     IAMC_IDX,
     SORT_IDX,
+    KNOWN_FUNCS
 )
 from pyam.read_ixmp import read_ix
 from pyam.timeseries import fill_series
@@ -1422,18 +1423,16 @@ def _aggregate(df, by, method='sum', weightvar_df=None):
 
 
 def _get_method_func(method):
-    if method == 'min':
-        _agg_func = np.min
-    elif method == 'max':
-        _agg_func = np.max
-    elif method == 'avg':
-        _agg_func = np.average
-    elif method == 'sum':
-        _agg_func = np.sum
-    else:
-        _agg_func = method
-    return _agg_func
+    """Translate a string to a known method"""
+    if not isstr(method):
+        return method
 
+    if method in KNOWN_FUNCS:
+        return KNOWN_FUNCS[method] 
+
+    # raise error if `method` is a string but not in dictionary of known methods
+    raise ValueError('method `{}` is not a known aggregator'.format(method))
+    
 
 def _raise_filter_error(col):
     raise ValueError('filter by `{}` not supported'.format(col))
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1413,13 +1413,12 @@ def _meta_idx(data):
     return data[META_IDX].drop_duplicates().set_index(META_IDX).index
 
 
-def _aggregate(df, by, method='sum', weightvar_df=None):
+def _aggregate(df, by, method):
     """Aggregate `df` by specified column(s), return indexed `pd.Series`"""
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
     # pick aggregator func (default: sum)
-    _agg_func = _get_method_func(method)
-    return df.groupby(cols)['value'].agg(_agg_func)
+    return df.groupby(cols)['value'].agg(_get_method_func(method))
 
 
 def _get_method_func(method):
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: pyam/core.py [hunk 1]
```diff
@@ -801,6 +802,12 @@ def aggregate_region(self, variable, region='World', subregions=None,
         append: bool, default False
             append the aggregate timeseries to `data` and return None,
             else return aggregate timeseries
+        aggregator: str, default 'sum'
+            aggregator function. One of 'sum', 'avg', 'min', 'max', 'w.avg'
+            w.avg requires weight_var to be provided as well
+        weight_var: str,
```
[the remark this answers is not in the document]
peterkolp: fixed

pyam/core.py
```diff
@@ -813,11 +820,21 @@ def aggregate_region(self, variable, region='World', subregions=None,
             logger().info(msg.format(variable, region))
 
             return
-
+        known_aggregators = ['sum','avg','min','max','w.avg']
```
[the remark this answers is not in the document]
peterkolp: added.

pyam/core.py
```diff
@@ -813,11 +820,21 @@ def aggregate_region(self, variable, region='World', subregions=None,
...
+        if not aggregator in known_aggregators:
```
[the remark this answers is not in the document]
peterkolp: changed.

danielhuppmann:
Thanks @peterkolp for the draft of this feature extension!

One concern, though, is that we are overloading the `aggregate_region()` function, which already has one "extra" feature, namely to look for `components` only defined at the `region` level and include those values in the aggregation.

My suggestion:
 - [ ] extract "translation" of `str` to `function` in auxiliary function so that it can be re-used later (e.g, by #241)
 - [ ] rename argument `aggregator` in `aggregate_region()` to `method`
 - [ ] refactor the "weighting" feature into own function `aggregate_region_by_weight()`, which calls `aggregate_region(components=[], ...)` to cleanly separate the two features (and reduce scope for user error)

pyam/core.py
```diff
@@ -1361,11 +1378,31 @@ def _meta_idx(data):
...
 
 
-def _aggregate(df, by):
+def _aggregate(df, by, aggregator='sum', weightvar_df=None):
     """Aggregate `df` by specified column(s), return indexed `pd.Series`"""
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
-    return df.groupby(cols).sum()['value']
+    # pick aggregator func (default: sum)
+    if aggregator == 'w.avg' and not weightvar_df.empty:
```
zikolach: it won't work if you pass `w.avg` and empty weightvar_df as expected because it will then silently fall back to `sum`

danielhuppmann: agree, one more reason for splitting the weighted-average into another function

peterkolp: OK. Agree that silently switching to method='sum' is bad behaviour.
What's your take on "expected behaviour" in that case (method='w.avg' and no weighting var)?

1. - just report ERROR (no aggregation)
2. - silently switch to avg
3. - report problem (WARN/ERROR) and switch to avg
4. - ...

same behaviour in case of weighting var (string) provided - but there's no data?

pyam/core.py
```diff
@@ -1361,11 +1378,31 @@ def _meta_idx(data):
...
+        calcdf = pd.merge(df, weightvar_df,
+                          how='left', suffixes=('', '_weight'),
```
zikolach: just out of curiosity - why `left`? is it similar to SQL's join type? what witl it do with records from left frame having no corresponding records in right one?

peterkolp: good question. we could use it to generate error messages about missing weights. but I'm not sure whether that will make a difference. 
@danielhuppmann any hints how to deal with partially missing weights?

pyam/core.py
zikolach: I would report error in both cases as it seems to be inconsistent input

pyam/core.py
```diff
@@ -1361,11 +1412,27 @@ def _meta_idx(data):
...
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
-    return df.groupby(cols).sum()['value']
+    # pick aggregator func (default: sum)
+    _agg_func = _get_method_func(method)
+    return df.groupby(cols)['value'].agg(_agg_func)
+
+
+def _get_method_func(method):
+    if method == 'min':
```
zikolach: Would it make sense to check type of input parameter and use something like `getattr(np, method)` to collapse multiple if branches?

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,179 @@
...
+            'R11_MEA',
+            'R11_NAM', 'R11_PAO', 'R11_PAS', 'R11_SAS', 'R11_WEU']
+r5_asiaregs = ['R11_CPA', 'R11_SAS']
+r5_oecdregs = ['R11_NAM', 'R11_PAO', 'R11_WEU']
+
+
+def read_file(filename):
+    df = pd.read_excel(filename, None)
+    pdf = pd.concat(
+           [df.get(sheet) for sheet in df.keys()
```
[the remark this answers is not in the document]
peterkolp: removed

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,179 @@
...
+               if sheet.startswith('data')])
+    return pyam.IamDataFrame(pdf)
+
+
+# idf = read_file(input_filename)
+
+TEST_DF = pd.DataFrame([
+   ['model_a', 'scen_a', 'region_a', 'Primary Energy', 'EJ/y', 1, 6.],
```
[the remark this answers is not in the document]
peterkolp: outdated!

pyam/core.py
```diff
@@ -817,7 +870,8 @@ def aggregate_region(self, variable, region='World', subregions=None,
         # compute aggregate over all subregions
         subregion_df = self.filter(region=subregions)
         cols = ['region', 'variable']
-        _data = _aggregate(subregion_df.filter(variable=variable).data, cols)
+        _data = _aggregate(subregion_df.filter(variable=variable).data,
+                           cols, method=method)
 
         # add components at the `region` level, defaults to all variables one
```
danielhuppmann: please add an `if` statement that the `components` defined only at `region` level are exclusively added when using `method in [`'sum', np.sum]`

pyam/core.py
```diff
@@ -721,6 +721,8 @@ def aggregate(self, variable, components=None, append=False):
             variable for which the aggregate should be computed
         components: list of str, default None
             list of variables, defaults to all sub-categories of `variable`
+        method: func or str
+            method to use for aggregation
```
danielhuppmann: please add `, e.g., ['mean', np.min]`

peterkolp: ok.

pyam/core.py
peterkolp: ok.

RELEASE_NOTES.md
```diff
@@ -16,6 +16,7 @@
 
 ## Individual Updates
 
+- [#244](https://github.com/IAMconsortium/pyam/pull/244) add aggregation methods (min/max/avg) to aggregate_ragion and add eighted_average_region function
```
fonfon: typos: aggregate_region; weighted_average_region

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
+import pyam
```
fonfon: typo in filename; how about just `test_region_aggregation`?

fonfon:
Some tests fail on my machine, and also on the travis builds

pyam/core.py
```diff
@@ -1361,11 +1423,24 @@ def _meta_idx(data):
...
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
-    return df.groupby(cols).sum()['value']
+    # pick aggregator func (default: sum)
+    return df.groupby(cols)['value'].agg(_get_method_func(method))
+
+
+def _get_method_func(method):
+    """Translate a string to a known method"""
+    if not isstr(method):
```
fonfon: A `if callable(method)` would be a stricter check

pyam/core.py
```diff
@@ -782,10 +788,60 @@ def check_aggregate(self, variable, components=None, exclude_on_fail=False,
...
+
+        weightvar_df = subregion_df.filter(variable=weight).data
+
+        if weightvar_df.empty:
+            msg = "cannot aggregate because no data found "\
+                "or weight_var='{}'"
+            logger().error(msg.format(weight))
+            return
+        var_df = subregion_df.filter(variable=variable).data
+        # _data = _aggregate(var_df, cols,
```
fonfon: I'd prefer to omit unused code, or is there a reason to still keep this?

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+                               subregions=['CPA'],
+                               append=True, weight='POP')
+    exp_l = [v for v in list(df.filter(region='CPA',
+             variable='FE').data.value) if not math.isnan(v)]
+    obj_l = [v for v in list(df.filter(region='A1_CPA').data.value)
+             if not math.isnan(v)]
+    np.testing.assert_allclose(obj_l, exp_l)
+
+
+# @pytest.mark.skip(reason="w.avg does not work with zero or None values")
```
fonfon: There is some commented code in this file -- I'd either add a TODO with a short explanation if you plan to still update/use it, or remove the code otherwise

pyam/core.py
```diff
@@ -721,6 +721,9 @@ def aggregate(self, variable, components=None, append=False):
             variable for which the aggregate should be computed
         components: list of str, default None
             list of variables, defaults to all sub-categories of `variable`
+        method: func or str
+            method to use for aggregation
+            e.g. np.mean, np.sum, ... 'min','max', ...
```
danielhuppmann: missing space between `min` and `max`

pyam/core.py
```diff
@@ -782,10 +788,60 @@ def check_aggregate(self, variable, components=None, exclude_on_fail=False,
...
+        subregions = subregions or self._all_other_regions(region, variable)
+
+        subregion_df = self.filter(region=subregions)
+        cols = ['region', 'variable']
+
+        weightvar_df = subregion_df.filter(variable=weight).data
+
+        if weightvar_df.empty:
+            msg = "cannot aggregate because no data found "\
+                "or weight_var='{}'"
```
danielhuppmann: suggest to rephrase to "cannot aggregate because no weights found"
(no need to specify "or weight_var='{}'"

pyam/core.py
```diff
@@ -782,10 +788,60 @@ def check_aggregate(self, variable, components=None, exclude_on_fail=False,
...
+            logger().error(msg.format(weight))
+            return
+        var_df = subregion_df.filter(variable=variable).data
+        # _data = _aggregate(var_df, cols,
+        #                    method='w.avg', weightvar_df=weightvar_df)
+        cols = [c for c in list(var_df.columns) if c not in ['value'] + cols]
```
danielhuppmann: it's partly personal preference, but I think that the PR https://github.com/danielhuppmann/pyam/pull/12 is a far more pythonic and **readable** implementation

pyam/core.py
```diff
@@ -782,10 +788,60 @@ def check_aggregate(self, variable, components=None, exclude_on_fail=False,
...
+        else:
+            return _data
+
     def aggregate_region(self, variable, region='World', subregions=None,
-                         components=None, append=False):
+                         components=None, append=False, method='sum'):
         """Compute the aggregate of timeseries over a number of regions
-        including variable components only defined at the `region` level
+
+        This function adds `components` only defined at the `region` level
```
danielhuppmann: I'd add "This function aggregates the variable across subregions. When the specified method is `sum`, it also adds `components` that are only defined at the `region` level."

pyam/core.py
```diff
@@ -817,25 +871,33 @@ def aggregate_region(self, variable, region='World', subregions=None,
...
-        components = components or (
-            set(region_df._variable_components(variable)).difference(
-                subregion_df._variable_components(variable)))
-
-        if len(components):
-            rows = region_df._apply_filters(variable=components)
-            _data = _data.add(_aggregate(region_df.data[rows], cols),
-                              fill_value=0)
+        # ONLY in case method is sum/np.sum
+        if method in ['sum', np.sum]:
```
danielhuppmann: please add a `ValueError` when method is not summation and components is not `None` (to warn against ignored input)

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+], columns=['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010])
+
+
+@pytest.fixture(scope="function")
+def wavg_df():
+    df = pyam.IamDataFrame(data=WAVG_DF)
+    yield df
+
+
+TEST2_DF = pd.DataFrame([
```
danielhuppmann: not sure why we even need such a detailed test dataframe - all behaviours should be checkable with `WAVG_DF`.

If really necessary, please follow the naming convention of other test dataframes.

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+# assert ' because it does not exist in any subregion' in caplog.text
+# assert caplog.records[0].levelname == 'ERROR' # actually INFO
+# assert len(idf.data.columns) == 7
+# pd.testing.assert_frame_equal(TEST_DF, df.data)
+# @pytest.mark.xfail(reason="xxx")
+
+
+def test_single_region_agg(reg_df):
+    for m in ['min', 'max', 'avg', 'sum']:
+        my_df = pyam.IamDataFrame(reg_df.data)
```
danielhuppmann: please refactor `my_df` to `df` (or `_df`). And in any case, using `df = reg_df.copy()` is an easier way to get the same result.

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+        my_df.aggregate_region('Primary Energy', region='A1_MEA', method=m,
+                               subregions=['MEA'], append=True)
+        np.testing.assert_allclose(
+            list(my_df.filter(region='A1_MEA').data.value),
+            list(my_df.filter(region='MEA',
+                              variable='Primary Energy').data.value))
+        assert list(my_df.filter(region='A1_MEA').data.value) == \
```
danielhuppmann: why test for both `assert_allclose` and `==`?

in any case, `df.data.value` returns to a `pd.Series`, so the `pandas.testing` should be easier to use. And if this only tests one region, please use the smaller test dataframe.

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+# assert 'cannot aggregate variable' in caplog.text
+# assert ' because it does not exist in any subregion' in caplog.text
+# assert caplog.records[0].levelname == 'ERROR' # actually INFO
+# assert len(idf.data.columns) == 7
+# pd.testing.assert_frame_equal(TEST_DF, df.data)
+# @pytest.mark.xfail(reason="xxx")
+
+
+def test_single_region_agg(reg_df):
+    for m in ['min', 'max', 'avg', 'sum']:
```
danielhuppmann: please add a short comment what this test is trying to do

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+            list(my_df.filter(region='A1_MEA').data.value),
+            list(my_df.filter(region='MEA',
+                              variable='Primary Energy').data.value))
+        assert list(my_df.filter(region='A1_MEA').data.value) == \
+            list(my_df.filter(region='MEA',
+                              variable='Primary Energy').data.value)
+
+
+def test_single_region_agg_nan_value(reg_df):
+    for m in ['min', 'max', 'avg']:
```
danielhuppmann: Same comments as above. Please check explicitly that the function returns an `np.nan` rather than using `dropna()`, and add a docstring that this is how the functions work.

tests/test_agrregation_region_aggrfuncs.py
```diff
@@ -0,0 +1,241 @@
...
+        my_df.aggregate_region('Primary Energy', region='A1_MEA', method=m,
+                               subregions=['MEA'], append=True)
+        assert list(my_df.filter(region='A1_MEA').data.value) == \
+            list(my_df.filter(region='MEA',
+                              variable='Primary Energy').data.dropna()[
+                'value'])
+
+
+def test_single_region_sum_nan_value(reg_df):
+    for m in ['sum']:
```
danielhuppmann: Same comments as above. Please add a docstring that `sum` drops `np.nan` and returns zero if no data is given.

pyam/core.py
```diff
@@ -752,6 +754,8 @@ def check_aggregate(self, variable, components=None, exclude_on_fail=False,
             variable to be checked for matching aggregation of sub-categories
         components: list of str, default None
             list of variables, defaults to all sub-categories of `variable`
+        method: func or str
+            method to use for aggregation
```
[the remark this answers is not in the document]
peterkolp: done.

pyam/core.py
peterkolp: done.

pyam/core.py
peterkolp: rephrased.

pyam/core.py
peterkolp: proposed text added.

pyam/core.py
peterkolp: just a left over. removed.

tests/test_agrregation_region_aggrfuncs.py
peterkolp: @danielhuppmann 

- what pandas.testing do you have in mind?
 I've had no luck with assert_series_equal nor Series.equal. Those do complain about `series.index values being different` ...

tests/test_agrregation_region_aggrfuncs.py
peterkolp: Hm. What actually happens is that for mn/max/avg there is no "data point" returned for the year having np.nan in input. (in case of sum we get sum(Nan) = 0. for that one)

tests/test_agrregation_region_aggrfuncs.py
danielhuppmann: see [pd.testing.assert_series_equal](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_series_equal.html), this allows to ignore different index types

pyam/core.py
```diff
@@ -817,25 +871,33 @@ def aggregate_region(self, variable, region='World', subregions=None,
...
+        if method in ['sum', np.sum]:
+            region_df = self.filter(region=region)
+            components = components or (
+                set(region_df._variable_components(variable)).difference(
+                    subregion_df._variable_components(variable)))
+
+            if len(components):
+                rows = region_df._apply_filters(variable=components)
+                _data = _data.add(_aggregate(region_df.data[rows], cols),
+                                  fill_value=0)
```
danielhuppmann: maybe we need a `_data.dropna(inplace=True)` here?

tests/test_agrregation_region_aggrfuncs.py
peterkolp: refactored (--> df)

pyam/utils.py
```diff
@@ -32,6 +32,8 @@
                           + ['{}{}'.format(i, j) for i, j in itertools.product(
                               string.ascii_uppercase, string.ascii_uppercase)]))
 
+KNOWN_FUNCS = {'min': np.min, 'max': np.max, 'avg': np.mean, 'sum': np.sum}
```
zikolach: @peterkolp does it still missing `w.avg` key? or I misunderstand how to call `aggregate_region` method

tests/test_agrregation_region_aggrfuncs.py
peterkolp: w.avg - weighted averages are handled by `weighted_average_region` func.
e.g.:
`    df.weighted_average_region('Price', append=True, weight='Primary Energy')`

danielhuppmann:
closing in favour of #305

PR #246, opened 2019-07-19
hotfix for pandas 0.25.0 update, closes #245

### LINKED ISSUE #245: Pandas 0.25 breaks plotting
Pandas 0.25 has removed `pandas.plotting._style`, it's now in `pandas.plotting._matplotlib.style`. Is there a way to avoid this reliance on a private module of pandas so this break doesn't happen in future (my brief look says no but just in case you have better ideas)?

gidden:
Per the discussions on the `pandas-dev` list, we simply incorporate the private module file we need in our code base.

### DIFF
--- a/pyam/_style.py
+++ b/pyam/_style.py
@@ -1,0 +1,92 @@
+# being a bit too dynamic
+import warnings
+
+import matplotlib.cm as cm
+import matplotlib.colors
+import numpy as np
+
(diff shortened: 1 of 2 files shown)


znicholls:
@danielhuppmann over to you

danielhuppmann:
thanks @gidden for the quick fix

PR #247, opened 2019-07-22
Fix required version as tests are failing with pandas v0.23.4 installed

zikolach:


gidden:
hi @zikolach out of curiosity, what is the issue with 0.23.4? Also, if we want to exclude a specific version, better to do so explicitly than require a new one, e.g.,: `pandas!=0.23.4`

zikolach:
@gidden probably that's just thing you addressed in one of the latest hotfixes, just not tested against older versions of pandas.

```
________________________________________________________________________________________ test_set_meta_from_data_method_other_column[meta_df0] ________________________________________________________________________________________

meta_df = <pyam.core.IamDataFrame object at 0x113f017b8>

    def test_set_meta_from_data_method_other_column(meta_df):
        if 'year' in meta_df.data.columns:
            col, value = 'year', 2010
        else:
            col, value = 'time', pd.to_datetime('2010-07-21T00:00:00.0')
        meta_df.set_meta_from_data('pe_max_yr', variable='Primary Energy',
                                   method=np.max, column=col)
        exp = pd.Series(data=[value] * 2, index=EXP_IDX, name='pe_max_yr')
>       pd.testing.assert_series_equal(meta_df['pe_max_yr'], exp)

tests/test_feature_set_meta.py:137: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.7/site-packages/pandas/util/testing.py:1210: in assert_series_equal
    assert_attr_equal('dtype', left, right)
venv/lib/python3.7/site-packages/pandas/util/testing.py:941: in assert_attr_equal
    raise_assert_detail(obj, msg, left_attr, right_attr)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = 'Attributes', message = 'Attribute "dtype" are different', left = dtype('float64'), right = dtype('int64'), diff = None

    def raise_assert_detail(obj, message, left, right, diff=None):
        if isinstance(left, np.ndarray):
            left = pprint_thing(left)
        elif is_categorical_dtype(left):
            left = repr(left)
    
        if PY2 and isinstance(left, string_types):
            # left needs to be printable in native text type in python2
            left = left.encode('utf-8')
    
        if isinstance(right, np.ndarray):
            right = pprint_thing(right)
        elif is_categorical_dtype(right):
            right = repr(right)
    
        if PY2 and isinstance(right, string_types):
            # right needs to be printable in native text type in python2
            right = right.encode('utf-8')
    
        msg = """{obj} are different
    
    {message}
    [left]:  {left}
    [right]: {right}""".format(obj=obj, message=message, left=left, right=right)
    
        if diff is not None:
            msg += "\n[diff]: {diff}".format(diff=diff)
    
>       raise AssertionError(msg)
E       AssertionError: Attributes are different
E       
E       Attribute "dtype" are different
E       [left]:  float64
E       [right]: int64

venv/lib/python3.7/site-packages/pandas/util/testing.py:1035: AssertionError

```

setup.py
```diff
@@ -22,7 +22,7 @@
     'argparse',
     'numpy',
     'requests',
-    'pandas>=0.21.0,!=0.24.0',
+    'pandas>=0.25.0',
```
gidden: HI @zikolach, can you please do two things here:

First, make this line `pandas==0.23.4` so we can see the test failure explicitly

Second, update this line to `pandas>=0.21.0,!=0.24.0,!=0.23.4`

zikolach:
@gidden thanks for suggestion. Just set `pandas==0.23.4`, waiting for tests results

zikolach:
@gidden now builds failed, see e.g. [similar issue I mentioned above](https://ci.appveyor.com/project/gidden/pyam/builds/26306314/job/lhy0p0hol4nnkt92#L1585). Should we try another `0.23.*` versions?

gidden:
Hey @zikolach, I think this is up to your personal preference. You're welcome to walk the version list to try to find all incompatible versions. Another option is just to say that we require `pandas>0.23` which I think will address your issue while also minimizing our dependency restrictions.

gidden:
Apologies, this line should instead be `pandas>0.24.0`

zikolach:
> Apologies, this line should instead be `pandas>0.24.0`

I wanted to check compatibility with exact version `0.24.1` (missed second `=`)

zikolach:
@gidden could you please check it again. @zikolach thanks for suggestion with versions and PEP-440 reference.

gidden:
Ok, lgtm!

PR #248, opened 2019-07-22
skip plotting tests if geopandas cannot be imported

danielhuppmann:
# Description of PR

When setting up a new machine, I was not unable to get `geopandas` to install correctly because of package-channel-inconcistencies. This made all region-plotting tests fail locally, making debugging quite cumbersome.

This PR adds skipping of region-plotting tests if `import geopandas` fails.


gidden:
lgtm, thanks @danielhuppmann !

PR #249, opened 2019-07-30
Make log warning "if-filter()-returns-empty" optional

danielhuppmann:
# Description of PR

The function `filter()` prints a warning to the log if a returned `IamDataFrame` is empty. This makes sense in most instance (and should remain the default behaviour), but it produces verbose and potentially misleading output in some cases and larger scripts.

One such case (fixed in this PR) is the `aggregate_region()` function when the new `region` does not have any data yet - which is not a problem and should not raise a warning.

# Alternative implementations

An alternative to circumvent the warning would be:
```
new = IamDataFrame(df.data[df._apply_filters(**filter_kwargs)])
```
This is more verbose but computationally more efficient (first downselect, then cast as `IamDataFrame`, ignore meta) compared to `filter()` (make a full copy including `meta`, then downselect). Still, in our limited-size use cases, concise code wins over speed, I would think...?

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -842,7 +842,7 @@ class IamDataFrame(object):
 
         # add components at the `region` level, defaults to all variables one
         # level below `variable` that are only present in `region`
-        region_df = self.filter(region=region)
+        region_df = self.filter(region=region, warn_if_empty=False)
         components = components or (
             set(region_df._variable_components(variable)).difference(
                 subregion_df._variable_components(variable)))
@@ -951,15 +951,17 @@ class IamDataFrame(object):
         logger().info('{} non-valid scenario{} will be excluded'
                       .format(len(idx), '' if len(idx) == 1 else 's'))
 
-    def filter(self, keep=True, inplace=False, **kwargs):
+    def filter(self, keep=True, inplace=False, warn_if_empty=True, **kwargs):
         """Return a filtered IamDataFrame (i.e., a subset of current data)
 
         Parameters
         ----------
-        keep: bool, default True
+        keep: bool
             keep all scenarios satisfying the filters (if True) or the inverse
-        inplace: bool, default False
+        inplace: bool
             if True, do operation inplace and return None
+        warn_if_empty: bool
+            warn in the log ifthe returned `IamDataFrame` is empty
         filters by kwargs:
             The following columns are available for filtering:
              - metadata columns: filter by category assignment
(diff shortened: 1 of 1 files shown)


pyam/core.py
```diff
@@ -951,15 +951,17 @@ def _exclude_on_fail(self, df):
         logger().info('{} non-valid scenario{} will be excluded'
                       .format(len(idx), '' if len(idx) == 1 else 's'))
 
-    def filter(self, keep=True, inplace=False, **kwargs):
+    def filter(self, keep=True, inplace=False, warn_if_empty=True, **kwargs):
```
zikolach: not sure, but shouldn't it be configurable via setting log level?

danielhuppmann: the point is that in some particular cases such as the one where the new option is used, there should *never*  be a log message

gidden: I think it can be a bit cleaner to explicitly catch expected warnings a la [`catch_warnings`](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)

danielhuppmann:
@gidden, I tried to implement the behaviour using `with warnings.catch_warnings():` instead of the additional kwarg, but this requires to use `warnings.warn()` in filter()` instead of `logger().warning()` - and as a consequence, the logging between the different levels `WARN`, `INFO`, `ERROR` is not consistent.

I looked at https://docs.python.org/3/library/logging.html#logging.captureWarnings, but this does not have the expected effect (if `True`, warnings are not logged at all).

So I would still go with the approach as implemented in this PR.

gidden:
Hey @danielhuppmann, I went through a bit of the link you sent and related pages and found this very useful advice: https://docs.python.org/2/howto/logging.html#when-to-use-logging

The basic gist, applied to our case, is that we should probably use `warnings.warn()` instead of `logger.warn()` (that's probably also true in lots of other places... we can make an issue if that's useful). 

This provides two benefits:
1) an empty result of a filtered dataframe is avoidable, and users should really be warned if it's happening so they can do something about it (i.e. `if the issue is avoidable and the client application should be modified to eliminate the warning`)
2) if the user *really* wants to still do this and they don't want the message, they can use the native capabilities to suppress it (a la [`capture_warnings`](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)

danielhuppmann:
It may be better to use `warnings.warn()` in theory, but this has some practical implications which I did not manage to solve:

1) using `pyam.logger().setLevel('ERROR')` would not affect warnings (because they are not controlled by `logging` anymore)
2) the root-logger styling is lost on warnings, so we might get output like:
    ```
    INFO:root:Reading `<some file>`
    Filtered IamDataFrame is empty!
    INFO:root:All scenarios have the required variable `<some variable>`
    ```
    So the warnings would not be marked as such in the log file - not ideal...

It may be possible to fix those issues - but it's beyond my Python skills...

danielhuppmann:
see #258 for an alternative solution approach

gidden:
I think #258 is really elegant, so will close in favor of that option1

PR #251, opened 2019-08-07
Region ag hotfix

gidden:
This is a hotfix for region aggregation as observed by @zikolach. It 

a) checks that regions actually exist, throwing an error if not
b) checks that `pattern_match` is provided actual data, throwing an error if not

@zikolach can you please test this and make sure it catches your use case? We can iterate on this branch as needed.

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -309,9 +309,14 @@ def pattern_match(data, values, level=None, regexp=False, has_nan=True):
     matching of model/scenario names, variables, regions, and meta columns to
     pseudo-regex (if `regexp == False`) for filtering (str, int, bool)
     """
-    matches = np.array([False] * len(data))
+    # coerce values to a list
     if not isinstance(values, collections.Iterable) or isstr(values):
         values = [values]
+
+    if len(data) == 0 or len(values) == 0:
+        return []
+
+    matches = np.array([False] * len(data))
 
     # issue (#40) with string-to-nan comparison, replace nan by empty string
     _data = data.copy()
--- a/tests/test_feature_aggregate.py
+++ b/tests/test_feature_aggregate.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pandas as pd
-from pyam import check_aggregate, IAMC_IDX
+from pyam import check_aggregate, IAMC_IDX, IamDataFrame
 
 from conftest import TEST_DTS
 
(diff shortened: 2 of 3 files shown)


zikolach:
As discussed the solution won't work in the case when aggregated region does not exist in source data (which is the case as I need to aggregate R11 regions to non-existing yer but to-be-created R5 regions).
Regarding `pattern_match` function I would argue for returning empty array as the result for empty data as similar behavior is specific to filter/map functions

gidden:
Yes, apologies, the region check was erroneous. My worry with returning and empty list is whether existing functions work as expected in that case. @zikolach is it possible to provide a test that exhibits the behavior you're experiencing? If so, I can make all relevant updates to confirm the new behavior is as expected.

gidden:
Hi all - I have now added a test that shows that even with returning empty lists, we still do not get expected behavior. Given that this wasn't tested before, I am not sure exactly what to implement to get to expected behavior. @danielhuppmann if/when you have time, can you please update this branch with the corrected implementation?

pyam/utils.py
```diff
@@ -309,10 +309,15 @@ def pattern_match(data, values, level=None, regexp=False, has_nan=True):
     matching of model/scenario names, variables, regions, and meta columns to
     pseudo-regex (if `regexp == False`) for filtering (str, int, bool)
     """
-    matches = np.array([False] * len(data))
+    # coerce values to a list
     if not isinstance(values, collections.Iterable) or isstr(values):
         values = [values]
 
+    if len(data) == 0 or len(values) == 0:
```
zikolach: @gidden you return here a "normal" python array - would you expect it can provide same contacts are np.ndarray?

tests/test_feature_aggregate.py
```diff
@@ -4,6 +4,13 @@
 
 from conftest import TEST_DTS
 
+
+def test_missing_region(check_aggregate_df):
+    print(check_aggregate_df['region'])
+    obs = check_aggregate_df.aggregate_region('Primary Energy', region='foo')
```
zikolach: what is the expected behavior if no subregions provided?

gidden: It should agg all regions I think. @danielhuppmann ?

gidden:
closing in favor of #252

PR #252, opened 2019-08-07
Region ag hotfix

gidden:
With @zikolach, we finally got the region aggregation back online. However, to do so we decided to make the API contracts for `aggregate_region` more consistent. Now if you `append=False`, you get back an `IamDataFrame`. @danielhuppmann would you mind helping with the test passage? Sadly, I don't have time to go through them all and convert. In the meantime, @zikolach can use this branch to continue his work.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -844,8 +844,11 @@ class IamDataFrame(object):
         # level below `variable` that are only present in `region`
         region_df = self.filter(region=region)
         components = components or (
-            set(region_df._variable_components(variable)).difference(
-                subregion_df._variable_components(variable)))
+            set(region_df._variable_components(variable))
+            .difference(
+                subregion_df._variable_components(variable)
+            )
+        )
 
         if len(components):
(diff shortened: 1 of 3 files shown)


danielhuppmann:
@gidden, it's not obvious to me which issue this fix addresses. I had offline discussions with @zikolach yesterday about two questions - but neither seems related to this PR.

Also, if I understand you correctly, tests fail because the return-class of `aggregate_region()` changed - so it would be good to re-implement this PR in two steps:
1) solve the issue (and explain it in the PR description), add unit tests, without modifying the return class - existing and new tests should pass
2) change the return class - existing tests will fail. I can then tackle this at some point.

In any case, the changed return type should then also be applied to `df.aggregate()`.

danielhuppmann:
also, see related PR #249 which improved the log output of `aggregate()` and `aggregate_region()` and should be merged prior to more refactoring

gidden:
As discussed @danielhuppmann, closing in favor of #254

PR #254, opened 2019-08-08
hotfix for pyam aggregation and filter in the wild

gidden:
@zikolach has run into a few issues when trying to use pyam in the wild with unruly datasets. This hotfix patches some issues seen when certain functions are called with missing data. Specifically:

- `df.aggregate_region()` now works as expected when the region-to-aggregate-to DNE in the dataframe already
- `df.filter()` now does not error if `df` is empty

There is also an example test that can be implemented when `aggregate_region()` is updated as referenced in #255

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -182,6 +182,13 @@ def test_variable_unit(test_df):
     npt.assert_array_equal(test_df.variables(include_units=True), exp)
 
 
+def test_filter_empty_df():
+    cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
+    data = pd.DataFrame([], columns=cols)
+    df = IamDataFrame(data=data)
+    df.filter(variable='foo')
+
+
 def test_filter_variable_and_depth(test_df):
     obs = list(test_df.filter(variable='*rimary*C*', level=0).variables())
     exp = ['Primary Energy|Coal']
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -184,7 +184,7 @@ def test_variable_unit(test_df):
 
 def test_filter_empty_df():
     # test for issue seen in #254
-    cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
+    cols = IAMC_IDX + [2005, 2010]
     data = pd.DataFrame([], columns=cols)
     df = IamDataFrame(data=data)
     obs = df.filter(variable='foo')
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -183,10 +183,12 @@ def test_variable_unit(test_df):
 
 
 def test_filter_empty_df():
+    # test for issue seen in #254
     cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
     data = pd.DataFrame([], columns=cols)
     df = IamDataFrame(data=data)
-    df.filter(variable='foo')
+    obs = df.filter(variable='foo')
+    assert len(obs) == 0
 
 
 def test_filter_variable_and_depth(test_df):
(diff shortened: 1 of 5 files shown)


gidden:
Interestingly, this passes on my laptop, but not on my desktop...

danielhuppmann:
the new test fails on my laptop with
```
        pd.testing.assert_frame_equal(obs.reset_index(drop=True),
>                                     exp.reset_index(drop=True))
E       AttributeError: 'memoryview' object has no attribute 'reset_index'
```
which is different from the error seen in CI. The offending line in CI passes.

gidden:
As requested @danielhuppmann, each commit is self contained:

- fa93e4b shows the original failure seen by @zikolach 
- 2e835b8 fixes that error
- 6a812a7 fixes another related issue seen by @zikolach in the wild when trying to filter empty dataframes
- c6685a7 provides two additional tests to confirm that behavior is now as @zikolach expects when doing region aggregation

tests/test_core.py
```diff
@@ -182,6 +182,15 @@ def test_variable_unit(test_df):
     npt.assert_array_equal(test_df.variables(include_units=True), exp)
 
 
+def test_filter_empty_df():
+    # test for issue seen in #254
+    cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
```
danielhuppmann: suggest to use `pyam.IAMC_IDX + [2005, 2010]`

tests/test_feature_aggregate.py
```diff
@@ -1,9 +1,56 @@
...
+    # obs = check_aggregate_df.filter(region='foo').data
+    # assert len(exp) > 0
+    # pd.testing.assert_frame_equal(obs.reset_index(drop=True),
+    #                               exp.reset_index(drop=True))
+
+
+def test_aggregate_region_extra_subregion():
+    cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
+    data = pd.DataFrame([
+        ['TEST', 'scen', 'China', 'Primary Energy', 'EJ/y', 1, 6],
```
danielhuppmann: please change model and scenario for consistency with `pyam/tests/conftest:TEST_DF`

tests/test_feature_aggregate.py
```diff
@@ -1,9 +1,56 @@
...
+                              region='R5ASIA',
+                              subregions=['China', 'Vietnam', 'Japan'],
+                              components=[], append=False)
+    assert len(obs) == 2
+
+
+def test_aggregate_region_missing_all_subregions():
+    cols = ['model', 'scenario', 'region', 'variable', 'unit', 2005, 2010]
+    data = pd.DataFrame([
+        ['TEST', 'scen', 'foo', 'Primary Energy', 'EJ/y', 1, 6],
```
danielhuppmann: as above, please change model and scenario for consistency with `pyam/tests/conftest:TEST_DF`

zikolach:
@gidden can you please update PR according to @danielhuppmann? Or push it to origin and let me know so I can do it

gidden:
Hi @zikolach, feel free to just PR into my remote. Once I pull it will be updated here as well.

tests/test_core.py
zikolach: fixed

tests/test_feature_aggregate.py
zikolach: fixed

tests/test_feature_aggregate.py
zikolach: fixed

zikolach:
@gidden I pushed a commit implementing @danielhuppmann suggestions. On my machine tests pass. Please rebase this PR as it has (trivial) conflicting changes in README.

gidden:
Ok, fixed the merge conflict here. @danielhuppmann if this meets your approval, I think it should be g2g after CI.

PR #258, opened 2019-08-09
add a fixture `logger.ignoreWarnings()`, use in `aggregate_region()`

danielhuppmann:

- [x] Documentation Added
- [x] Tests added

# Description of PR

The function `filter()` prints a warning to the log if a returned `IamDataFrame` is empty. This makes sense in most instance (and should remain the default behaviour), but it produces verbose and potentially misleading output in some cases and larger scripts.

One such case (fixed in this PR) is the `aggregate_region()` function when the new `region` does not have any data yet - which is not a problem and should not raise a warning.

This PR adds a fixture `logger.adjust_log_level()` (see [logging.captureWarnings](https://docs.python.org/2/library/logging.html#logging.captureWarnings) for comparison) as short-hand to change the logging level around a block of code.

# Expected use

```
with logger.adjust_log_level():
    df.filter(model='foo')
```

### DIFF
--- a/pyam/logger.py
+++ b/pyam/logger.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 import logging
 
 # globally accessible logger
@@ -14,14 +15,10 @@ def logger():
     return _LOGGER
 
 
-class ignoreWarnings():
-    """Fixture to ignore logging messages below `level`"""
-    def __init__(self, level='ERROR'):
-        self.level = _LOGGER.getEffectiveLevel()
-        _LOGGER.setLevel(level)
-
-    def __enter__(self):
-        pass
-
-    def __exit__(self, exception_type, exception_value, traceback):
-        _LOGGER.setLevel(self.level)
+@contextmanager
+def adjust_log_level(level='ERROR'):
+    """Context manager to change log level"""
+    old_level = _LOGGER.getEffectiveLevel()
+    _LOGGER.setLevel(level)
+    yield
+    _LOGGER.setLevel(old_level)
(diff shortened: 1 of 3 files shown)


danielhuppmann:
@gidden, building the docs fails, all other tests pass - can you please review, and maybe take a look at why the docs fails? thanks!

gidden:
hey @danielhuppmann, nice! it may be easier to implement this as a [`context_manager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) so you don't need to worry about the overhead that comes with making it a class. That would look something like:

```
@context_manager
def adjust_log_level(level='ERROR'):
    old_level = _LOGGER.getEffectiveLevel()
    _LOGGER.setLevel(level)
    yield
    _LOGGER.setLevel(old_level)
```

then use it the same way:

```
with adjust_log_level():
    region_df = self.filter(region=region)
```

I think a test here would be as simple as:

```
assert _LOGGER.getEffectiveLevel() == 'WARNING'
with adjust_log_level():
    assert _LOGGER.getEffectiveLevel() == 'ERROR'
assert _LOGGER.getEffectiveLevel() == 'WARNING'
```

gidden:
lgtm, I'm not sure there is a need to add a release note unless you want to. I will merge (unless you say not to) after CI finishes.

danielhuppmann:
yeah, I don't think this needs a release note... going too far down the rabbit hole for our standard users.

gidden:
Thanks @danielhuppmann !!

pyam/logger.py
```diff
@@ -12,3 +13,12 @@ def logger():
         _LOGGER = logging.getLogger()
         _LOGGER.setLevel('INFO')
     return _LOGGER
+
+
+@contextmanager
+def adjust_log_level(level='ERROR'):
```
zikolach: @danielhuppmann what would happen if `adjust_log_level` is called before `logger`? Is that valid use case? Currently tests in #262 are failing because of that

Issue #227, opened 2019-04-29
region plotting breaks in geopandas 0.5.0


gidden:
It appears region plotting is broken with the latest geopandas release, see attached MPL images (we lose France, among other countries). We should decide:

- do we support this new version and make us backwards incompatible?
- do we ask geopandas how best to adapt?
- do we make geopandas<0.5.0 a hard install requirement?

any thoughts @danielhuppmann or @znicholls 
![test_region-failed-diff](https://user-images.githubusercontent.com/1392657/56891605-59b4fb00-6a7d-11e9-8dc1-f4a517502f4b.png)
![baseline-test_region](https://user-images.githubusercontent.com/1392657/56891606-59b4fb00-6a7d-11e9-899f-ce8103fc62b3.png)
![test_region](https://user-images.githubusercontent.com/1392657/56891607-5a4d9180-6a7d-11e9-908b-0eb9cb3a9fdd.png)

znicholls:
> do we make geopandas<0.5.0 a hard install requirement?

I would recommend this for current releases (and master). Then when you can be bothered, add geopandas>0.5.0 and make a new (backwards incompatible) release. I don't use mapping so I don't really mind and would prefer to see efforts focussed elsewhere (but if mapping is a priority for you guys, go for it).

gidden:
Yes, I am +1 this as well for the moment. The problem with making these requirements is that you start to quickly become incompatible with other packages (that in the future have a ">0.5.0" req).

znicholls:
I think at some point you (we) need to hit pause on development and upgrade both pandas and geopandas to be compatible with new versions. There's already conflicts with things that require pandas>0.24.0 and pyam. (I would also suggest streamlining the release process so they can be made more frequently at the same time, but let's see how things go.)

gidden:
agreed. I think this is on @danielhuppmann's radar too

danielhuppmann:
yes, it's on my radar - afaik, we are compatible with `pandas` 0.24.1 (they resolved the issue for `combine_first()` which caused unit tests to fail in 0.24.0). But the amount of `pandas` warnings on the latest version (replacing `labels` with `codes` etc.) makes looking at unit test output quite painful.

danielhuppmann:
This is causing more issues with dependencies bumping their version numbers, see #260 for a quickfix - this PR needs to be reverted once we move to `geopandas==0.5.0` .

Also, readthedocs still claims to be failing to build on Travis, even though it passes...?

[...]

PR #260, opened 2019-08-16
[WIP] fix error not finding shared file `lipoppler.so.76` when building the docs on CI

danielhuppmann:
This is a hotfix for a follow-up problem when building the docs on CI because of the current restriction `geopandas<0.5.0`, see #227 

This fix **does not fix** the issues with building the docs on readthedocs.


gidden:
lgtm, thanks @danielhuppmann!

PR #261, opened 2019-08-17
add check that `keep` in `filter()` must be boolean

### LINKED ISSUE #253: `filter()` assumes non-starred `dict` as `keep` kwarg
# Observed (user) error

Just ran into a stupid error where I used `df.filter(dict(foo='bar')` instead of `df.filter(**dict(foo='bar')`, so rather than applying the expected `foo='bar'` rule, the function interprets the dict as the `keep` kwarg.

# Suggested solution

Add a check that `keep` is a boolean, raise error otherwise.

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

A non-starred `dict` in `df.filter()` is mis-interpreted as the `keep` kwarg, returning a copy of the full (rather than downselected) `IamDataFrame`. This PR adds a check the `keep` must be boolean.

closes #253

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -206,6 +206,11 @@ def test_filter_error_illegal_column(test_df):
     pytest.raises(ValueError, test_df.filter, foo='test')
 
 
+def test_filter_error_keep(test_df):
+    # non-starred dict is mis-interpreted as `keep` kwarg, see #253
+    pytest.raises(ValueError, test_df.filter, dict(model='foo'))
+
+
 def test_filter_year(test_df):
     obs = test_df.filter(year=2005)
     if "year" in test_df.data.columns:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -208,7 +208,7 @@ def test_filter_error_illegal_column(test_df):
 
 def test_filter_error_keep(test_df):
     # string or non-starred dict was mis-interpreted as `keep` kwarg, see #253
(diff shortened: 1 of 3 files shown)


gidden:
Looks good @danielhuppmann. One request: can you add an explicit test for a non-bool'd `keep`? I think the last test is actually doing this, but it would be nice to have something explicit (e.g., `keep='foo'`).

tests/test_core.py
```diff
@@ -201,8 +201,15 @@ def test_variable_depth_raises(test_df):
...
-def test_filter_error(test_df):
-    pytest.raises(ValueError, test_df.filter, foo='foo')
+def test_filter_error_illegal_column(test_df):
+    # filtering by column `foo` is not valid
+    pytest.raises(ValueError, test_df.filter, foo='test')
+
+
+def test_filter_error_keep(test_df):
+    # string or non-starred dict was mis-interpreted as `keep` kwarg, see #253
+    pytest.raises(ValueError, test_df.filter, model='foo')
```
gidden: I think this needs to be: `pytest.raises(ValueError, test_df.filter, model='foo', keep=1)` or something similar to explicitly test the boolean-ness.

gidden:
ok, lgtm! will merge when CI passes.

danielhuppmann:
thanks @gidden for the review!

PR #263, opened 2019-08-20
Fix adjusting log level called with uninitialized logger

zikolach:


gidden:
Hi @zikolach, happy to merge this as is. Is it possible to write a test that fails under the current master? We obviously don't have this edge case in our current test suite (or else it would have failed).

danielhuppmann:
not sure how easy it is to write a test that is guaranteed to be called first for all `pyam`...

pyam/logger.py
```diff
@@ -18,7 +18,8 @@ def logger():
 @contextmanager
 def adjust_log_level(level='ERROR'):
     """Context manager to change log level"""
-    old_level = _LOGGER.getEffectiveLevel()
-    _LOGGER.setLevel(level)
+    _logger = logger()
```
danielhuppmann: it should be enough to replace only the first line by `old_level = logger().getEffectiveLevel()` - this will ensure that the internal variable ´_LOGGER` is initialized later in the function (and avoids calling the `if-then` logic three times)

gidden:
> 
> 
> not sure how easy it is to write a test that is guaranteed to be called first for all `pyam`...

I thought so as well, but a quick test:

```
def test_foo():
    with adjust_log_level():
	print('foo')

```

```
$ pytest tests/test_logger.py 
============================================================================== test session starts ===============================================================================
platform linux -- Python 3.7.3, pytest-4.4.0, py-1.8.0, pluggy-0.9.0
Matplotlib: 3.0.3
Freetype: 2.9.1
rootdir: /home/gidden/work/iiasa/pyam
plugins: mpl-0.10, cov-2.6.1
collected 2 items                                                                                                                                                                

tests/test_logger.py F.                                                                                                                                                    [100%]

==================================================================================== FAILURES ====================================================================================
____________________________________________________________________________________ test_foo ____________________________________________________________________________________

    def test_foo():
>       with adjust_log_level():

tests/test_logger.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../miniconda3/envs/pyam37/lib/python3.7/contextlib.py:112: in __enter__
    return next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

level = 'ERROR'

    @contextmanager
    def adjust_log_level(level='ERROR'):
        """Context manager to change log level"""
>       old_level = _LOGGER.getEffectiveLevel()
E       AttributeError: 'NoneType' object has no attribute 'getEffectiveLevel'

../../../.local/lib/python3.7/site-packages/pyam_iamc-0.2.0+13.gd2fa051-py3.7.egg/pyam/logger.py:21: AttributeError
======================================================================= 1 failed, 1 passed in 0.26 seconds =======================================================================
```

pyam/logger.py
zikolach: I think calling `logger()` and using its result makes clearer consent that `_LOGGER` may be not initialized (= should not be used directly). Having `logger().getEffectiveLevel()`and `_LOGGER.setLevel(level)` looks messy from my point of view. But if you think your variant more correct I can change it

PR #264, opened 2019-08-26
update docs and readme to point to `groups.io` and `RTD`

danielhuppmann:
# Description of PR

This PR updates the README and docs to point to `groups.io` and `RTD` (rather than googlegroups and https://software.ene.iiasa.ac.at/pyam).

It also rephrases to "scenario" rather than "model" in the `pyam` docs teaser (because this is closer to what the package does). This change is not inconsistent with the published manuscript.


gidden:
thanks @danielhuppmann !

PR #267, opened 2019-09-20
Region aggregation if region missing

znicholls:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

Updates pyam so behaviour of `check_aggregate_region` is sensible even if the region your checking is missing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 # Next Release
 
 - [#268](https://github.com/IAMconsortium/pyam/pull/268) Update `aggregate_region` so it can find variables below sub-cateogories too
+- [#267](https://github.com/IAMconsortium/pyam/pull/267) Make clear error message if variable-region pair is missing when `check_aggregate_region` is called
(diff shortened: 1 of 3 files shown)


znicholls:
@danielhuppmann @gidden thoughts would be great here. The new tests currently fail in a cryptic way. The question is, if you call `check_aggregate_region` and the region you're checking isn't there, should an error be raised, a warning or just nothing and `None` is returned (cause a region can't be wrong if it's not there...)?

danielhuppmann:
From a quick look, it seems that the command tries to return the object as a pivoted `pd.DataFrame` (via `timeseries()`), which doesn't work if the dataframe is empty.

On the question "how do we expect this to behave", I would follow the behaviour in `aggregate_region()` (log message and empty return): https://github.com/IAMconsortium/pyam/blob/127680a8f412659c440655c4238c33cb45b528da/pyam/core.py#L831

znicholls:
@danielhuppmann this should be good to go

PR #268, opened 2019-09-23
Extended consistency checking

znicholls:

- [x] Tests Added
- [x] Documentation Added
- [x] Add this feature to aggregation tutorial

```
```

# Description of PR

Extends the checking of internal consistency of data. The use case is this: a variable is reported at the World level only, but is more than one level deeper in the hierarchy than the variable being considered. Previously, our check_aggregate functions would raise an error, now they find this variable and correctly realise that the data is internally consistent.


danielhuppmann:
thanks @znicholls, but I'm not sure that this change makes sense - do we really want to "jump" over levels in the variable tree? Wouldn't it make more sense to require that the variable `foo|bar` is also defined at the region level?

pyam/core.py
```diff
@@ -906,12 +907,15 @@ def check_aggregate_region(self, variable, region='World', subregions=None,
...
+    def _variable_components(self, variable, all=False):
+        """Get all components (sub-categories) of a variable for a given level
 
-        For `variable='foo'`, return `['foo|bar']`, but don't include
-        `'foo|bar|baz'`, which is a sub-sub-category"""
+        If `all`, for `variable='foo'`, return `['foo|bar']`, but don't include
+        `'foo|bar|baz'`, which is a sub-sub-category. Otherwise return all
+        variables below `variable` in the hierarchy"""
         var_list = pd.Series(self.data.variable.unique())
+        if all:
```
danielhuppmann: How about the following - looks more readable to me...?
```
_level = None if all is True else 0
return var_list[pattern_match(var_list, '{}|*'.format(variable), _level)]
```

pyam/core.py
```diff
@@ -906,12 +907,15 @@ def check_aggregate_region(self, variable, region='World', subregions=None,
...
             return IamDataFrame(diff, **col_args).timeseries()
 
-    def _variable_components(self, variable):
-        """Get all components (sub-categories) of a variable
+    def _variable_components(self, variable, all=False):
+        """Get all components (sub-categories) of a variable for a given level
 
-        For `variable='foo'`, return `['foo|bar']`, but don't include
-        `'foo|bar|baz'`, which is a sub-sub-category"""
+        If `all`, for `variable='foo'`, return `['foo|bar']`, but don't include
```
danielhuppmann: Isn't this exactly the other way round? `all is True` returns all sub-sub-...-categories...

znicholls:
> Wouldn't it make more sense to require that the variable `foo|bar` is also defined at the region level?

I think it would but I also think that ship has sailed. For example, aviation and shipping are often reported at the World level only and you need to take account of that when checking regional sums. Pyam can already do that, this PR just makes it so it can take account variables which are deeper in the hierarchy. That is required in some cases (the current one for me being sorting variables into MAGICC's fossil and afolu boxes: I need to put aviation below fossil but make sure aviation is counted when making aggregates/checking consistency).

znicholls:
Not sure why windows tests are failing, will try rebasing...

danielhuppmann:
@znicholls, still not quite convinced...

> For example, aviation and shipping are often reported at the World level only and you need to take account of that when checking regional sums. 

That, I agree with - but if you have a structure like
 - `Emissions|CO2|<sectors>` defined for all regions
 - `Emissions|CO2|Aviation|<sub-categories>` defined at the `World` level
I would still argue that it should be required that `Emissions|CO2|Aviation` is defined also at the `World` level - in which case the current implementation works (i.e., checks that `Aviation` is the sum of sub-categories and includes it in the validation of global CO2 emissions).

znicholls:
> That, I agree with - but if you have a structure like

Yes, but if you have a structure like

- `Emissions|CO2` (at the world and regional level)
- `Emissions|CO2|Fossil` (at the world and regional level)
- `Emissions|CO2|Fossil|sub-sectors` (where sub-sectors doesn't include `Aviation`)
- `Emissions|CO2|Fossil|Aviation` (at the world level only)

then the current setup will fail when checking the aggregation of `Emissions|CO2` because it will add all the `Emissions|CO2|Fossil` variables at the regional level and won't find the required `Emissions|CO2|Fossil|Aviation` at the world level

gidden:
I'm +1 with @znicholls on this one. Bunkers are a consistent pain the in side for all emissions-related calculations, and I think first class support for them in `pyam` operations is an important feature to include.

That being said, would it make sense to add a few more tests here @znicholls? At the moment, I see that you added a shipping sector in the test data, but I don't think any tests are being applied explicitly for this feature (correct me if I'm wrong)..

Also, would it be useful to add this new capability to the existing aggregation tutorial? I guarantee that folks who don't know it exists a prior will never find it buried either in the code or this PR.

gidden:
And ho-boy, it looks like windows tests are failing in our tutorial tests with `ModuleNotFoundError: No module named 'win32api'`. That looks like a doozy... I can try to look into it soon.

gidden:
@zicholls, can you try adding `conda install pywin32` to `appveyor.yml`? I have no idea why this is needed now. Maybe Appveyor changed some default configs...

znicholls:
> At the moment, I see that you added a shipping sector in the test data, but I don't think any tests are being applied explicitly for this feature (correct me if I'm wrong)..

Adding the shipping sector made the test suite fail (before the change) so I think that's testing for this feature in some way. Having said that, happy to add whatever tests you think are sensible (I just couldn't think of other obvious ones beyond what we already have). 

> Also, would it be useful to add this new capability to the existing aggregation tutorial?

Will put it on the todos.

gidden:
> > At the moment, I see that you added a shipping sector in the test data, but I don't think any tests are being applied explicitly for this feature (correct me if I'm wrong)..
> 
> Adding the shipping sector made the test suite fail (before the change) so I think that's testing for this feature in some way. Having said that, happy to add whatever tests you think are sensible (I just couldn't think of other obvious ones beyond what we already have).
> 

Ok, music to my ears, but also not obvious! I guess I was thinking of just including a consistency check of the sector total, but maybe not needed? I leave the decision in your expert hands.

znicholls:
> Ok, music to my ears, but also not obvious! I guess I was thinking of just including a consistency check of the sector total, but maybe not needed? I leave the decision in your expert hands.

Added a couple more which hopefully make things more obvious. Over to you

pyam/core.py
```diff
@@ -843,9 +844,13 @@ def aggregate_region(self, variable, region='World', subregions=None,
         # level below `variable` that are only present in `region`
         with adjust_log_level():
             region_df = self.filter(region=region)
+
+        rdf_components = region_df._variable_components(variable, level=None)
```
danielhuppmann: Maybe rename these internal variables to have them in one line each?

danielhuppmann:
@gidden, over to you for a final review and merge

znicholls:
> I approve the changes, but I'm still not convinced that this is a good path forward for the `check_consistency()` function - it only works on an `IamDataFrame`that has

I agree that this only works for a specific use case. However that's the use case I currently have (pulling out the CMIP6 data, doing some operations on it, making sure I didn't cock anything up along the way). I think the limitations you point out are great things to go for next, I just have no idea how to implement them. Maybe we make a new issue for this?

gidden:
Agreed that there are limitations here, but I see us building up a clean interface. When more features are needed, we can dig into the details, but the community writ large also needs to decide on those details (how do we define sub-sub regions, e.g.). 

@danielhuppmann you know that a pet peeve of mine is the hierarchical tree anyway, so something that forces us more in that direction is not a negative at least to me =)

In any case, regarding that, I do think we can come up with some intelligent "cleaning" operations that basically takes a non-strict hierarchy, makes it strict, does an op, and then deals with additional data after. But I think that is a fight for another time..

gidden:
thanks all for the robust discussion!

Issue #271, opened 2019-10-02
download a whole database snapshot


byersiiasa:
Currently I don't find a way to pull from the server a whole database.

In a situation where all models report the same complete set of variables, then one could iterate through a list of models. But in cases where different models have reported different or incomplete sets, this isn't possible - 
Perhaps a whole new function would be best, like `conn.query(model='*', variable='*', region='*')`

A little similar to #137

byersiiasa:
I will try a PR to edit the `read_iiasa` function, but for non-public dbs requires adding the credentials.

byersiiasa:
Can someone (perhaps @zikolach or Peter Kolp),  please check my branch.  I've added the necessary fix to the `read_iiasa` function, but it is failing on something to do with `sub-annual`

Lines 344 and 361 of https://github.com/byersiiasa/pyam/blob/fix_read_iiasa/pyam/iiasa.py
```
df = pyam.read_iiasa('IXSE_AR6', user='byers', pw='*******')
INFO:root:You are connected to the IXSE_AR6 scenario explorer hosted by IIASA. If you use this data in any published format, please cite the data as provided in the explorer guidelines: https://data.ene.iiasa.ac.at/ar6-scenario-submission/#/about.
Traceback (most recent call last):

  File "<ipython-input-2-01a089b1c36c>", line 1, in <module>
    df = pyam.read_iiasa('IXSE_AR6', user='byers', pw='iannoush')

  File "c:\users\byers\src\pyam\pyam\iiasa.py", line 358, in read_iiasa
    df = conn.query(**kwargs)

  File "c:\users\byers\src\pyam\pyam\iiasa.py", line 328, in query
    if pd.Series([i in [-1, 'year'] for i in df.subannual]).all():

  File "C:\Users\byers\Continuum\Anaconda3\envs\pyamTV\lib\site-packages\pandas\core\generic.py", line 5180, in __getattr__
    return object.__getattribute__(self, name)

AttributeError: 'DataFrame' object has no attribute 'subannual'
```

byersiiasa:
BTW same occurs for the SR15 SE also. Sometimes before this error it also throws a similar error along lines of
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\GitHub\pyam\pyam\iiasa.py", line 363, in read_iiasa
    df = conn.query(**kwargs)
  File "C:\GitHub\pyam\pyam\iiasa.py", line 325, in query
[...]

PR #273, opened 2019-10-11
[WIP] Fix requesting ixmp data over API (#271)

zikolach:
This is a bug fix. Please update release notes if needed

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

- pass credentials to request available apps
- handle missing (optional) fields in server responses (meta)
- add test for invalid credentials

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -321,7 +321,7 @@ class Connection(object):
         _check_response(r)
         # refactor returned json object to be castable to an IamDataFrame
         df = pd.read_json(r.content, orient='records')
-        df = pd.DataFrame(data = df,
+        df = pd.DataFrame(data=df,
                           columns=['meta', 'model', 'region', 'runId',
                                    'scenario', 'time', 'unit', 'value',
                                    'variable', 'version', 'year'])
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -320,11 +320,10 @@ class Connection(object):
         r = requests.post(url, headers=headers, data=data)
         _check_response(r)
         # refactor returned json object to be castable to an IamDataFrame
-        df = (
-            pd.read_json(r.content, orient='records')
-            .drop(columns='runId')
-            .rename(columns={'time': 'subannual'})
-        )
+        df = pd.read_json(r.content, orient='records')
+        if df.empty:
+            return df
+        df.drop(columns='runId').rename(columns={'time': 'subannual'})
         # check if returned dataframe has subannual disaggregation, drop if not
         if pd.Series([i in [-1, 'year'] for i in df.subannual]).all():
             df.drop(columns='subannual', inplace=True)
@@ -341,7 +340,7 @@ class Connection(object):
         return df
 
 
-def read_iiasa(name, meta=False, **kwargs):
+def read_iiasa(name, meta=False, creds=None, **kwargs):
     """
     Query an IIASA database. See Connection.query() for more documentation
 
@@ -354,9 +353,9 @@ def read_iiasa(name, meta=False, **kwargs):
     kwargs :
         Arguments for pyam.iiasa.Connection.query()
     """
-    conn = Connection(name)
+    conn = Connection(name, creds)
     # data
-    df = conn.query(**kwargs)
+    df = conn.query(creds=creds, **kwargs)
     df = IamDataFrame(df)
     # metadata
     if meta:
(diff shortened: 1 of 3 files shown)


pyam/iiasa.py
```diff
@@ -31,7 +31,7 @@
 
 def _check_response(r, msg='Trouble with request', error=RuntimeError):
     if not r.ok:
-        raise error('{}: {}'.format(msg, str(r)))
+        raise error('{}: {}'.format(msg, str(r.content)))
```
fonfon: `r.text` would be more appropriate here

byersiiasa:
So now the creds auth is  now working better but I still don't see how to extract the whole database.. An if I specify a variable or model, we get the 1000 variable error.

byersiiasa:
SO this is working - but one should not request the whole db at once, or it crashes the system

pyam/iiasa.py
```diff
@@ -320,11 +326,15 @@ def query(self, **kwargs):
...
-        df = (
-            pd.read_json(r.content, orient='records')
-            .drop(columns='runId')
-            .rename(columns={'time': 'subannual'})
-        )
+        df = pd.read_json(r.content, orient='records')
+        columns = ['model', 'scenario', 'variable', 'unit',
+                   'region', 'year', 'value', 'time', 'meta',
+                   'runId', 'version']
+        df = pd.DataFrame(data=df, columns=columns)
```
gidden: this seems a bit brittle to me - e.g., what if the rest api adds additional columns to the return obj at some future point in time?

Would suggest something more like:

```
columns = ['model', 'scenario', 'variable', 'unit',
                   'region', 'year', 'value', 'time', 'meta',
                   'version']
df = pd.read_json(r.content, orient='records')
drop = list(set(df.columns) - set(columns))
df.drop(columns=drop, inplace=True)
df.fillna({'meta': 0}, inplace=True)
df.rename(columns={'time': 'subannual'}, inplace=True)
```

pyam/iiasa.py
```diff
@@ -333,15 +343,15 @@ def query(self, **kwargs):
             df[META_IDX + ['version']].drop_duplicates()
             .groupby(META_IDX).count().version
         )
-        if max(lst) > 1:
+        if not lst.empty and max(lst) > 1:
```
gidden: can't this be done with `if len(lst) > 1`?

pyam/iiasa.py
```diff
@@ -333,15 +343,15 @@ def query(self, **kwargs):
...
             raise ValueError('multiple versions for {}'.format(
                 lst[lst > 1].index.to_list()))
         df.drop(columns='version', inplace=True)
 
         return df
 
 
-def read_iiasa(name, meta=False, **kwargs):
+def read_iiasa(name, meta=False, creds=None, **kwargs):
```
gidden: docstring for this please!

pyam/iiasa.py
```diff
@@ -354,9 +364,9 @@ def read_iiasa(name, meta=False, **kwargs):
     kwargs :
         Arguments for pyam.iiasa.Connection.query()
     """
-    conn = Connection(name)
+    conn = Connection(name, creds)
     # data
-    df = conn.query(**kwargs)
+    df = conn.query(creds=creds, **kwargs)
```
gidden: if the connection already has creds, why are they needed here as well?

gidden:
Hi all - seems like nice changes, thanks @zikolach. I made a few inline comments, but one overarching one is that this seems to be touching a few different things, but it's not clear to me that they are being tested?

pyam/iiasa.py
zikolach: I think that's the artifact separating creds from read_iiasa kwargs (just passing all the arguments as before). Will remove

pyam/iiasa.py
zikolach: Sure. Just out of curiosity - why it is better in your opinion in this particular case?

zikolach:
> Hi all - seems like nice changes, thanks @zikolach. I made a few inline comments, but one overarching one is that this seems to be touching a few different things, but it's not clear to me that they are being tested?

Yes, you are right. There are multiple issues addressed in PR and other changes (e.g. ixmp_source repo). In particular this PR should solve:
- retrieving list of apps according to provided credentials (not public ones)
- be tolerant to non-provided values (such as meta) in json response payloads

Thanks for review and valuable comments!

pyam/iiasa.py
zikolach: thanks

pyam/iiasa.py
gidden: Ah, I think I misunderstood this. You're checking if the version is `>1`? Then your previous implementation was correct - my apologies. It would be worth adding a comment saying that version has to be `>1` and why.

Issue #225, opened 2019-04-24
`map_regions()` to `downscale_regions()`


gidden:
With `aggregate_regions()` doing the work to upscale (many-to-one) regional data, `map_regions()` should be deprecated and a new `downscale_regions()` should be added. The function signature should look something like:

```
def downscale_regions(mapping, proxy=None)
```

By default, it should copy regional data into subregions. If provided a proxy (i.e., another `variable`), then it should do a pattern scaling to that variable, which must be supplied at the "sub regional" resolution.

danielhuppmann:
Summarizing discussion in #207 - rather than deprecate `map_regions()`, the function should be refactored to parse the given mapping and call `aggregate_region()` or `downscale_region()` iteratively as required.

This allows to distinguish the sanity-checking of the mapping (e.g., is a particular order required aggregating from country to region to world) from the data manipulation operation.

danielhuppmann:
there is now a dedicated `downscale_region()` function, but no work has been done on deprecating/refactoring the `map_regions()` function

PR #275, opened 2019-10-18
remove all region-plotting related features, docs, etc.

gidden:
This completely removes all region-plotting related code. I will make an issue to reinstate it.


znicholls:
What's the motivation for removing `map_regions`, that works fine doesn't it?

danielhuppmann:
> What's the motivation for removing `map_regions`, that works fine doesn't it?

See the related issue https://github.com/IAMconsortium/pyam/issues/225

znicholls:
> See the related issue #225

Ok but there's features in `map_region` that aren't in `aggregate_region` right? Particularly the inbuilt mappings? If this is a refactoring operation (which is my understanding of #225), wouldn't it make sense to leave `map_regions` in and ensure continuity of the tests during the refactoring? (I ask out of curiosity and to learn more about how to approach this sort of operation, I don't actually have any strong feelings.)

gidden:
Ok all, thanks first to @znicholls - we can and should keep `map_region` in for the moment. It's added back with tests, this should be good to go now.

PR #276, opened 2019-10-24
[WIP] Add subtraction operation

znicholls:

- [ ] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

This PR supercedes #229 and adds a subtraction operation for IamDataFrames. It is specifically very limited in scope as there are likely many kinks to iron out. Further PRs can add extra operations, ``other`` types etc.


znicholls:
@gidden this should get us started. To keep things simple, I've implemented so both ``self`` and ``other`` must have only one value in ``join_col`` (as then it's easy to do naming). Happy to take suggestions on how to make this better. The failing test puzzles me, basically I don't really understand pyam's logic about metadata and timeseries hence am not sure when we need to check for overlapping metadata and when not.

pyam/core.py
```diff
@@ -1382,6 +1382,58 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+        if len(self[join_col].unique()) > 1:
+            raise ValueError(too_many_vals_error.format("self", join_col))
+
+        if len(other[join_col].unique()) > 1:
+            raise ValueError(too_many_vals_error.format("other", join_col))
+
+        s_ts = self.timeseries()
+        o_ts = other.timeseries()
+
+        if set(s_ts.index.names) != set(o_ts.index.names):
```
danielhuppmann: I'd suggest to move the `meta`-equality assertion from `append()` to an own auxiliary function and re-use here, https://github.com/IAMconsortium/pyam/blob/a4e266c93f531f94285311fd354aaa881aa403c4/pyam/core.py#L202

tests/test_feature_operations.py
```diff
@@ -0,0 +1,84 @@
...
+#   - subtracting Series
+#   - subtracting floats
+#   - subtracting vectors
+#   - subtracting arrays
+#   - other operations e.g. multiplication, addition, division
+
+def test_subtraction_with_self(check_aggregate_df):
+    tdf = check_aggregate_df.filter(variable="Primary Energy")
+    res = tdf.subtract(tdf, "variable", "zero")
+    np.testing.assert_array_equal(res.timeseries(), 0)
```
danielhuppmann: testing would be easier on `res.data.value`

tests/test_feature_operations.py
```diff
@@ -0,0 +1,84 @@
...
+#   - what happens if indexes are different (NotImplementedError?)
+
+# list of todo's for future PRs:
+#   - subtracting Series
+#   - subtracting floats
+#   - subtracting vectors
+#   - subtracting arrays
+#   - other operations e.g. multiplication, addition, division
+
+def test_subtraction_with_self(check_aggregate_df):
```
danielhuppmann: `check_aggregate_df` is a less-than-ideal test case, because it's too large to quickly grasp what a function should do on it. Any chance you could use this opportunity to reduce it to one page of data (and harmonise how the model-and-scenario columns are defined)?

https://github.com/IAMconsortium/pyam/blob/a4e266c93f531f94285311fd354aaa881aa403c4/tests/conftest.py#L42

znicholls:
> `nan` or zero or assumption that the missing value is zero (i.e., either value of `self`or negative value of `other`)?

Surely `np.nan` right? Getting a value back when you haven't done an operation would be super weird.

pyam/core.py
```diff
@@ -1382,6 +1382,58 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+        Parameters
+        ----------
+        other : pyam.IamDataFrame
+            Object containing data to subtract
+
+        join_col : str
+            Column to use to subtract the two sets of data (e.g. ``variable``)
+
+        new_name : str
+            Name to assign to ``join_col`` in the output timeseries
```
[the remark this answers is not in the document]
znicholls: It's the name of whatever column you've just written. Say you subtract two scenarios, it's what you would write in the scenario column. I've tried updating, see what you think.

pyam/core.py
```diff
@@ -1382,6 +1382,58 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+            raise NotImplementedError
+
+        too_many_vals_error = "`{}` contains more than one entry for `{}`"
+        if len(self[join_col].unique()) > 1:
+            raise ValueError(too_many_vals_error.format("self", join_col))
+
+        if len(other[join_col].unique()) > 1:
+            raise ValueError(too_many_vals_error.format("other", join_col))
+
+        s_ts = self.timeseries()
```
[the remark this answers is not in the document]
znicholls: no reason I just find timeseries easier to work with so was using them whilst fiddling, have refactored now thanks for pointing that out

pyam/core.py
znicholls: I don't really understand what the meta checks in `append` are doing so have just taken the line you linked. Happy to refactor further if you can see how to do it.

tests/test_feature_operations.py
znicholls: It's kind of the point of it. The aggregate functions have to deal with a whole heap of edge cases so building a test case for it is hard and requires a fairly large dataframe. I agree that it's not great for a unit test, but it's the only way to do an integration test on the aggregate functions as far as I can tell. I'm happy to move the aggregation feature tests into a `tests/integration` folder and use a different test dataframe for testing the operations. However I think the operations will have to deal with a lot of edge cases too so having some integration tests on them could be worthwhile too.

danielhuppmann:
> Surely `np.nan` right? Getting a value back when you haven't done an operation would be super weird.

Well, I could imagine use cases where subtracting the value given by `other` if given or use 0 as a default (if a row is not present in `other`) would make sense. But as long as the behaviour is clear (and covered in the unit tests), dropping all rows that are not in `self` **and** `other` is ok. [Note that casting to an `IamDataFrame` removes all `nan`].

pyam/core.py
danielhuppmann: Yeah, I meant the entire section of the code, not just one line. I implemented the change for you, see https://github.com/znicholls/pyam/pull/3 - should be pretty straightforward to re-use it now in `subtract()`.

znicholls: Wicked thanks mate! Have pulled in and used now. The next question is what should happen when you're subtracting with the `join_col` being `model` or `scenario`. I guess you just get a brand new, empty `meta` table as you're resulting `model` or `scenario` won't exist in the `meta` table you started with?

danielhuppmann: Yeah, good question... I'd suggest to either:
 - Don't allow subtracting on `model` or `scenario` for this PR and tackle later.
 - Start with an empty `meta` table

pyam/core.py
```diff
@@ -1382,6 +1355,66 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
         if not inplace:
             return ret
 
+    def subtract(self, other, join_col, new_name, ignore_meta_conflict=False):
```
danielhuppmann: Looking [pd.DataFrame.subtract()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.subtract.html) for inspiration: how about renaming `join_col` to `axis`? If you think of timeseries data as multi-dimensional arrays, one could see `variable` or `region` as the axis...

pyam/core.py
```diff
@@ -1382,6 +1355,66 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+        idx_tmp = list(set(idx) - set([join_col]) - {"value"})
+
+        s_data = s_data.set_index(idx_tmp).drop(join_col, axis="columns")
+        o_data = o_data.set_index(idx_tmp).drop(join_col, axis="columns")
+
+        res = (s_data - o_data).reset_index()
+        res[join_col] = new_name
+
+        res = IamDataFrame(res)
+        res.meta = out_meta
```
danielhuppmann: the new `meta` table should be down-selected to those models/scenarios that are still in `data` after dropping all `nan`...

pyam/core.py
znicholls: Nice!

pyam/core.py
znicholls: I've implemented empty (after all, if you subtract scenarios, it's more or less a new scenario right?)

gidden:
@znicholls is it possible to rebase this? I may suggest that @gaurav-ganti implement a `divide()` method on this branch since the heavy lift has already been done.

Given that there is a lot of front/back matter here, what do you think of splitting each of these functions into something like?:

```
def subtract(...):
  s_data, o_data = _pre_op(...)
  res = (s_data - o_data).reset_index()
  _post_op(res, ..)
  return res
```

gidden:
dear all - I rebased this branch and pushed a new PR on the org remote so we can better keep track of new ops there (#333)

Issue #278, opened 2019-10-30
inconsisting requirements for building the docs locally


danielhuppmann:
The requirements for building the docs locally are not consistent across these lists.

 - `extensions` imported in `doc/source/conf.py`
 - installed packages in `doc/environment.yml`
 - `doc/README.md`
 - `doc/requirements.txt`

This issue was raised in #277 but not resolved.

danielhuppmann:
@gidden, wich list takes priority here (or can be removed)?

gidden:
> * `extensions` imported in `doc/source/conf.py`
This is basically the `import` list for extensions
> * installed packages in `doc/environment.yml`
This is the file used for building on `readthedocs`, likely should be the primary list
> * `doc/README.md`
This should point to the primary list
> * `doc/requirements.txt`
This is used in the `Makefile` and could likely use instead the `environment.yml` @znicholls any opinions here?

gidden:
FYI, I found I needed to change the name in the yml file, because otherwise it would by default install to a different environment (per the name): https://github.com/iiasa/aneris/pull/3/commits/b4d099cd134be48686bb00fa3b64a733d80b3b5b

znicholls:
> @znicholls any opinions here?

I deleted `requirements.txt` in favour of `environment.yml` in #279. I think the changes make sense but am a bit unsure about what happens on read the docs. I deleted the name from `environment.yml` as I couldn't see how else to make it behave, is read the docs name agnostic?

PR #279, opened 2019-10-30
Update to build docs locally

znicholls:
In response to #278


danielhuppmann:
thanks for trying @znicholls, but this doesn't work as expected on readthedocs, see https://pyam-iamc.readthedocs.io/en/docs-requirements-makefile/install.html

(I pushed your branch to https://github.com/IAMconsortium/pyam to have it built on RTD).

znicholls:
> I pushed your branch to https://github.com/IAMconsortium/pyam to have it built on RTD

Beauty thanks will work on that then!

PR #280, opened 2019-10-30
Update to build docs locally

### LINKED ISSUE #278: inconsisting requirements for building the docs locally

znicholls:
In response to #278 and supercedes #279 

closes #278


znicholls:
thanks @danielhuppmann, I just updated the docs requirements (I think). The `setup.py` issue seems to come from somewhere else (it's also in https://pyam-iamc.readthedocs.io/en/latest/install.html), @gidden any ideas how to fix that one?

gidden:
restarted travis, this was an iiasa conneciton issue.

And https://pyam-iamc.readthedocs.io/en/docs-requirements-makefile/ LGTM!!

gidden:
I see now the setup.py issue.. let me look

gidden:
Maybe we can confirm with something like `from setup import REQUIREMENTS`. At what point did this break? do we know?

znicholls:
I have no idea unfortunately. A fallback option would be to just point to
the setup.py for requirements rather than have them as a list for the time
being?

gidden:
Just talked to @khaeru. His suggestion is that RTD does not guarantee the same locality. We should try to change 

```
.. program-output:: python -c 'import sys; sys.path.append("../.."); import setup; print("\n".join([r for r in setup.REQUIREMENTS]))'
```

to something like (and this is getting rather long..)

```
.. program-output:: python -c 'import sys, os; sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../..")); import setup; print("\n".join([r for r in setup.REQUIREMENTS]))'
```
Basically, we need to make sure the import is relative to this file location. If that doesn't work, I'm not sure what is best to do... also trying this now on iiasa/aneris

gidden:
Hi all - this approach has failed for me (there is no `__file__` attribute for a subprocess call.  

I am testing the setup.py version on aneris: https://readthedocs.org/projects/aneris/builds/9889492/

My work around for the moment is to read directly from `ci/environment-conda-default.txt`. I have no idea why that approach works but importing `setup.py` does not..

gidden:
I am now uber confused. Reading direct from `setup.py` on RTD is a success for `aneris`: https://aneris.readthedocs.io/en/test-read-from-setup/install.html

danielhuppmann:
see this output when trying to use `from setup import REQUIREMENTS`: https://pyam-iamc.readthedocs.io/en/docs-test-rtd/install.html

Seems it's looking for `setup.py` in `/home/docs/checkouts/readthedocs.org/user_builds/pyam-iamc/conda/docs-test-rtd/lib/python3.7/site-packages/`

danielhuppmann:
I guess it's iterating over `PATH` and taking the first `setup.py` that it can find?

danielhuppmann:
got it in https://github.com/IAMconsortium/pyam/tree/docs-test-rtd (no idea why it's working differently from `aneris`).

@znicholls, as this branch is on the org repo, I took the liberty of committing the fix directly

znicholls:
> @znicholls, as this branch is on the org repo, I took the liberty of committing the fix directly

perfect that was the hope!

doc/source/install.rst
```diff
@@ -40,11 +40,11 @@ fully enumerated below.
 
 The required depedencies for |pyam| are:
 
-  .. program-output:: python -c 'import sys; sys.path.append("../.."); import setup; print("\n".join([r for r in setup.REQUIREMENTS]))'
+  .. program-output:: python -c 'import sys; sys.path.insert(0,"../.."); from setup import REQUIREMENTS; print("\n".join([r for r in REQUIREMENTS]))'
```
znicholls:
```suggestion
  .. program-output:: python -c 'import sys; sys.path.insert(0, "../.."); from setup import REQUIREMENTS; print("\n".join([r for r in REQUIREMENTS]))'
```

danielhuppmann: done

danielhuppmann:
One more question - we have a hard requirement for `matplotlib==3.0.2` in the docs, but no version specification in `setup.py`. There is some indication that unit tests fail with version `3.1.x` (but not rigorously tested). Should we add the version specification as part of this PR to be safe for the release?

gidden:
we can add `matplotlib>3.1`?

danielhuppmann:
> we can add `matplotlib>3.1`?

It would have to be `matplotlib<3.1`. See the brand-new #281.

gidden:
thanks guys!

PR #285, opened 2019-11-07
Add ability to request regions with synonyms from IXMP web API

zikolach:
Changes
- (optionally) fetch region synonyms
- fix fetching timeseries data explicitly specifying column types

---
- [x] Tests Added
- [x] Documentation Added

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 - [#295](https://github.com/IAMconsortium/pyam/pull/295) Include `meta` table when writing to or reading from `xlsx` files
 - [#292](https://github.com/IAMconsortium/pyam/pull/292) Add warning message if `data` is empty at initialization (after formatting)
 - [#288](https://github.com/IAMconsortium/pyam/pull/288) Put `pyam` logger in its own namespace (see [here](https://docs.python-guide.org/writing/logging/#logging-in-a-library>))
-
+- [#285](https://github.com/IAMconsortium/pyam/pull/285) Add ability to fetch regions with synonyms from IXMP API
 # Release v0.3.0
 
 ## Highlights
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -344,7 +344,9 @@ class Connection(object):
         r = requests.post(url, headers=headers, data=data)
         _check_response(r)
         # refactor returned json object to be castable to an IamDataFrame
-        df = pd.read_json(r.content, orient='records')
+        dtype = dict(model=str, scenario=str, variable=str, unit=str,
+                     region=str, year=int, value=float, version=int)
+        df = pd.read_json(r.content, orient='records', dtype=dtype)
         logger.debug('Response size is {0} bytes, '
                        '{1} records'.format(len(r.content), len(df)))
         columns = ['model', 'scenario', 'variable', 'unit',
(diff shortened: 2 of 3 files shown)


byersiiasa:
lgtm

gidden:
Haven't had a chance to review implementation, but we need at least one functionality test

zikolach:
@gidden @danielhuppmann I added a test (as backend used for integration testing is now up-to-date). Can you please re-review?

danielhuppmann:
thanks @zikolach - I was hoping that @gidden or @byersiiasa would review this, because I'm not sure how this is supposed to work...

Can you please rebase to resolve the merge conflict (changed `logger` implementation in #288)?

More generally, I'm not sure if this return-object (in wide format) is helpful, because one would need to check across an unknown number of columns (if I understand the format with `synonym_{}` correctly?). I think a return-object in long-format (see https://github.com/iiasa/ixmp/blob/399e1b04db32b8fb6f090dba611a900fd4c421ea/ixmp/core.py#L219) would be more practical?

Also, we should be able to tag "types" to synonyms, to distinguish between standard name, ISO3, alternative spelling, ...

zikolach:
PR rebased

> More generally, I'm not sure if this return-object (in wide format) is helpful, because one would need to check across an unknown number of columns (if I understand the format with synonym_{} correctly?). I think a return-object in long-format (see https://github.com/iiasa/ixmp/blob/399e1b04db32b8fb6f090dba611a900fd4c421ea/ixmp/core.py#L219) would be more practical?

The intention was to keep it kind of backward-compatible (meaning regions method should return same (unique) list of regions). What do you think about adding/extending method parameters, e.g. instead of `include_synonyms` have `synonyms_format` with:
- default option (None) to omit synonyms
- option `columns` - current behavior
- option `rows` - long format

> Also, we should be able to tag "types" to synonyms, to distinguish between standard name, ISO3, alternative spelling, ...

Can you explain a bit more in detail what you mean? Dtypes? What is recommended way to do it?

danielhuppmann:
Thanks @zikolach!

For backward compatibility, `include_synonyms=False` can be used as default.

I don't see a practical use case for `columns` format, because an a-priori unknown number of meaningful entries would make it a bit tedious working with a wide format. So just return `rows` (long format) for `include_synonyms=True`.

> > Also, we should be able to tag "types" to synonyms, to distinguish between standard name, ISO3, alternative spelling, ...

> Can you explain a bit more in detail what you mean? Dtypes? What is recommended way to do it?

Ignore this comment, let's tackle this at some future point - I realised this opens quite a few rabbit holes.

zikolach:
> I don't see a practical use case for columns format, ...

Practical use case is to get first synonym for specific region. With long format it is a bit more difficult (as I can see from my limited pandas knowledge)

danielhuppmann:
There is no such thing as a “first”, “second”, etc. synonym in the current logic. There can be any number of them, in any order.

We may want to introduce tags for synonyms, e.g. the ISO3 synonym (what I meant earlier). But this would require a major change in ixmp, restAPI, etc., so not to be tackled now.

zikolach:
> There is no such thing as a “first”, “second”, etc.

I meant "first available synonym", but that is not important. Main constraint as mentioned was to keep region names unique in the result dataset independently on parameters. If that is no issue then it is ok from my side also to implement as you suggested. Please re-review changes.

PR #288, opened 2019-11-10
Update logging

znicholls:

- [x] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

This PR updates pyam's logging so that it no longer clashes with the root logger and is properly namespaced. At the moment, pyam uses the root logger, which means you can't only turn Pyam's logging off rather you have to adjust all logging or none (see the new tests). With this change, pyam's loggers are put in their own namespace so they can be adjusted by libraries which use them without having to touch the root logger. The more detailed explanation of why this should be changed is explained [here](https://docs.python-guide.org/writing/logging/#logging-in-a-library) ([this](https://opensource.com/article/17/9/python-logging) could also be a useful resource if we want to write something about this in the docs).


znicholls:
@gidden does this need documentation?

gidden:
@znicholls not for the moment. if a future issue comes up, I'll be sure to assign you =)

gidden:
this looks great @znicholls, thank you so much!

Issue #291, opened 2019-11-15
`timeseries()` on empty `IamDataFrame` fails very loudly


danielhuppmann:
# Description of the issue

When calling `timeseries()` on an empty `IamDataFrame`, the internally used `pandas.DataFrame.pivot_table()` fails with a long, non-intuitive error message (see below). This might be quite intimidating to non-expert users.

## Improved behaviour

Checking whether the `data` table is empty before doing the pivot operation is straightforward. 

Two questions on implementation and expected behaviour, though:

1) Should `pyam` only write a message to the log file (as is done for `filter()`, see [here](https://github.com/IAMconsortium/pyam/blob/103b60f8f2f46fddae4d0325cbc2651cb098030a/pyam/core.py#L1001), or should a (nicer) error be raised?

2) If only a warning is written to the log, should the returned object be `None` or an empty `pandas.DataFrame` (but with the correct index dimensions).

# Error message

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-48-36a29f3c3d4a> in <module>
----> 1 df.filter(variable='foo').timeseries()

~/Github/pyam/pyam/core.py in timeseries(self, iamc_index)
    382         df = (
    383             self.data
--> 384             .pivot_table(index=index, columns=self.time_col)
    385             .value  # column name
    386             .rename_axis(None, axis=1)

~/anaconda3/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5177             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5178                 return self[name]
-> 5179             return object.__getattribute__(self, name)
   5180 
   5181     def __setattr__(self, name, value):

AttributeError: 'DataFrame' object has no attribute 'value'
```

danielhuppmann:
@gidden @znicholls any thoughts?

znicholls:
I would definitely make it a loud error, no logging (if downstream users really want that they can put it in a try except block or whatever)

PR #292, opened 2019-11-15
Log message nan data in init

znicholls:

- [x] Tests Added

```
```

# Description of PR

A log message is now thrown if any null data is found on initialisation so that users know it's being dropped (I get caught out a lot being like, 'where has all my data gone'). Inspired by #291 and discussions in #286

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if nan rows are detected in init data
 - [#288](https://github.com/IAMconsortium/pyam/pull/288) Put `pyam` logger in its own namespace (see [here](https://docs.python-guide.org/writing/logging/#logging-in-a-library>))
 
 # Release v0.3.0
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,7 @@
 
 # Next Release
 
-- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if `data` is empty at initialization (after formatting)
+- [#292](https://github.com/IAMconsortium/pyam/pull/292) Add warning message if `data` is empty at initialization (after formatting)
 - [#288](https://github.com/IAMconsortium/pyam/pull/288) Put `pyam` logger in its own namespace (see [here](https://docs.python-guide.org/writing/logging/#logging-in-a-library>))
 
 # Release v0.3.0
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,7 @@
 
 # Next Release
 
-- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if nan rows are detected in init data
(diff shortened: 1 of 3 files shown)


danielhuppmann:
Thanks @znicholls for this PR. I can't test it right now because I don't see any logging output (see #293), but I am worried about confusion this may create for non-expert users.

We are aiming `pyam` (also) at a not-so-sophisticated user base, and with the proposed change, one would see a warning message (about something being dropped) if you load a snapshot (wide format) with some variables reported at a yearly resolution (e.g., temperature output from a stylized climate model) and other variables reported  at decadal resolution (e.g., IAM output). This could easily be misunderstood as the snapshot being corrupted or incomplete.

znicholls:
> This could easily be misunderstood as the snapshot being corrupted or incomplete.

That's fair (but one could argue the snapshot is somehow incomplete as not everything is on the same timestep). Happy to also close this/take another route (only throw the warning if the dataframe is actually empty once you remove nan rows?).

danielhuppmann:
> Only throw the warning if the dataframe is actually empty once you remove nan rows.

Writing to log already happens if an `IamDataFrame` is empty after `filter()`, see [here](https://github.com/IAMconsortium/pyam/blob/103b60f8f2f46fddae4d0325cbc2651cb098030a/pyam/core.py#L1001) - I think it makes sense to replicate the same in `__init__()`.

Maybe it makes sense to add an `empty` attribute similar to [pandas.DataFrame.empty](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.empty.html)?

znicholls:
> Writing to log already happens if an `IamDataFrame` is empty after `filter()`, see [here](https://github.com/IAMconsortium/pyam/blob/103b60f8f2f46fddae4d0325cbc2651cb098030a/pyam/core.py#L1001) - I think it makes sense to replicate the same in `__init__()`.

Nice will do that then

znicholls:
> Maybe it makes sense to add an `empty` attribute similar to [pandas.DataFrame.empty](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.empty.html)?

Smart

RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if nan rows are detected in init data
```
danielhuppmann:
```suggestion
- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if `data` is empty at initialization (after formatting)
```

RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- [#290](https://github.com/IAMconsortium/pyam/pull/290) Add warning message if `data` is empty at initialization (after formatting)
```
danielhuppmann: just noticed one more little issue here...

```suggestion
- [#292](https://github.com/IAMconsortium/pyam/pull/292) Add warning message if `data` is empty at initialization (after formatting)
```

danielhuppmann:
thanks @znicholls for this nice improvement - better user experience in edge-cases (or user errors...)

PR #294, opened 2019-11-16
Make default logging more verbose

### LINKED ISSUE #293: no logging output in Jupyter notebooks
Following the merge of #288, there is no logging output written in Jupyter notebooks or ipython!

Is this a feature or a bug? (If it's a feature, it's not a great one).

znicholls:

- [ ] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

Closes #293

### DIFF
--- a/pyam/__init__.py
+++ b/pyam/__init__.py
@@ -1,5 +1,4 @@
 import logging
-import warnings
 
 from pyam.core import *
 from pyam.utils import *
@@ -13,7 +12,7 @@ from pyam.iiasa import read_iiasa  # noqa: F401
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
-# in Jupyter notebooks: disable autoscroll, activate warnings, set logger to info
+# in Jupyter notebooks: disable autoscroll and set logger to info
 try:
     get_ipython().run_cell_magic(u'javascript', u'',
                                  u'IPython.OutputArea.prototype._should_scroll = function(lines) { return false; }')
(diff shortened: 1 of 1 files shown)


danielhuppmann:
thanks @znicholls, this is more in line with what I'd expect in practice, logging-wise...

Logging output is now written in Jupyter notebooks and ipython consoles, but not a python console. It's not clear to me why one would not want to always display `INFO` output per default.

danielhuppmann:
@zikolach, can you please check whether the (already merged) change in #288 and/or this PR will affect (or break) the log output for the "upload to IIASA database" workflow?

znicholls:
> It's not clear to me why one would not want to always display `INFO` output per default.

It's just the community standard, by default only WARNING and above will appear. If you want INFO, you have to turn it on yourself (by namespacing your logger you let users of your library decide what they see and don't see rather than deciding for them). 

> @zikolach, can you please check whether the (already merged) change in #288 and/or this PR will affect (or break) the log output for the "upload to IIASA database" workflow?

If you're relying on getting INFO level log output, the answer is almost certainly yes.

danielhuppmann:
Ok, colour me ignorant, but I still think that it makes sense to be more verbose in Jupyter notebooks per default...

So I'm happy with this PR - unless @gidden or @zikolach have objections, I'd say it's ready to be merged...

znicholls:
> Ok, colour me ignorant, but I still think that it makes sense to be more verbose in Jupyter notebooks per default...

We can make pyam more verbose in the console too if we want. I think as long as there's a message that lets users know what's happening, they can turn things off later if they want and it might be the right choice for pyam users.

danielhuppmann:
> We can make pyam more verbose in the console too if we want.

I don't have a preference about the default logging level in the console. @gidden, if you don't have a strong opinion here, I'll merge as is to get the notebook logging fixed.

danielhuppmann:
merging as is to fix Jupyter notebook logging

please make an issue to restart the discussion whether the default logging level should be `INFO` in the console for consistency

PR #295, opened 2019-11-18
including `meta` table when reading and writing xlsx files

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Following #286 and some offline discussions about supporting more data formats, we may want to introduce more types of content of an `IamDataFrame` (e.g., annotations at a more fine-grained level than `meta`, provenance tracking) and more IO-formats (e.g., multiple csv files in https://frictionlessdata.io).

To prepare for these new features, this PR implements writing the `meta` table as part of `to_excel()` (include by default, option to switch off with `include_meta=False`) , so that other attributes can in the future also be added in the exported file, rather than exporting each attribute individually as currently implemented, i.e.,

```
df.to_excel(<data_file>)
df.export_metadata(<meta_file>)
```

Similarly, when initialising an `IamDataFrame` from `xlsx`, it will look for a `meta` sheet and use it to populate that table if available (can be deactivated with `pyam.IamDataFrame(<file>, meta_sheet_name=False)`.

As an aside, the PR also harmonises some notation, and streamlines the IO-tests and moves them into an own test file.


danielhuppmann:
per discussion with @gidden, I merge this PR to continue efficiently with some further improvements on the tutorials

@gidden, @znicholls, if you disagree with the chosen defaults of the new feature (writing the `meta` table to an Excel spreadsheet, trying to read the `meta` table when initialising from a spreadsheet), please open an issue for discussing whether to change them

PR #297, opened 2019-11-26
add `empty` attribute, better error for `timeseries()` on empty dataframe

### LINKED ISSUE #291: `timeseries()` on empty `IamDataFrame` fails very loudly

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This PR adds an `empty` attribute and adds a better error message when doing `timeseries()` on an empty `IamDataFrame`.

Closes #291

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -389,7 +389,7 @@ class IamDataFrame(object):
         Raises
         ------
         AttributeError
-            `IamDataFrame` is empty
+            ``IamDataFrame`` is empty
         ValueError
             reducing to IAMC-index yields an index with duplicates
         """
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -388,13 +388,13 @@ class IamDataFrame(object):
 
         Raises
         ------
-        AttributeError
+        ValueError
             ``IamDataFrame`` is empty
         ValueError
             reducing to IAMC-index yields an index with duplicates
         """
         if self.empty:
-            raise AttributeError('this `IamDataFrame` is empty')
+            raise ValueError('this `IamDataFrame` is empty')
 
         index = IAMC_IDX if iamc_index else IAMC_IDX + self.extra_cols
         df = (
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -148,6 +148,11 @@ class IamDataFrame(object):
         """Identical to pd.DataFrame.tail() operating on data"""
         return self.data.tail(*args, **kwargs)
 
+    @property
+    def empty(self):
+        """Indicator whether ``IamDataFrame`` is empty"""
+        return self.data.empty
+
     def models(self):
         """Get a list of models"""
         return pd.Series(self.meta.index.levels[0])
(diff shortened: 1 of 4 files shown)


pyam/core.py
```diff
@@ -148,6 +148,11 @@ def tail(self, *args, **kwargs):
         """Identical to pd.DataFrame.tail() operating on data"""
         return self.data.tail(*args, **kwargs)
 
+    @property
+    def empty(self):
+        """Indicator whether ``IamDataFrame`` is empty"""
```
```suggestion
        """Is the ``IamDataFrame`` empty?"""
```

pyam/core.py
```diff
@@ -380,7 +385,17 @@ def timeseries(self, iamc_index=False):
...
+
+        Raises
+        ------
+        AttributeError
+            ``IamDataFrame`` is empty
+        ValueError
+            reducing to IAMC-index yields an index with duplicates
         """
+        if self.empty:
+            raise AttributeError('this `IamDataFrame` is empty')
```
znicholls: Should this be an attribute error or perhaps a value error instead?

pyam/core.py
danielhuppmann: I was using the pandas docstring as a template, see https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.empty.html

pyam/core.py

znicholls:
Oh nice then keep it as is, I was just doing what seemed to  read more
easily to me but consistency with the wider community is a better choice

danielhuppmann:
thanks @znicholls for the review!

danielhuppmann:
Merging as approved by @znicholls

Issue #299, opened 2019-12-02
Suggest to change default behaviour of `aggegate_region()` for components


danielhuppmann:
## Introduction

The current implementation of [aggegate_region()](https://pyam-iamc.readthedocs.io/en/stable/api.html#pyam.IamDataFrame.aggregate_region) has an option to add `components` (other variables), which are only defined at the (aggregate) `region` level. Per default, the function uses all sub-categories of `variable` that are only defined at the `region` level (not in any of the `subregions`).

## Problem

The current implementation is very slow when iterating over many variables, and it might cause an issue when the sequence of aggregation matters on the outcome.

## Suggested solution

Allow `variable` to be a list of variables rather than only one (string), so that the region-aggregation can be performed in bulk. But this will cause problems with the automated detection of `components`. Also, passing the components for individual variables in the list would be cumbersome... (probably require a dictionary, what if that is not complete, ...).

Therefore, I suggest the following API change:
 - `variable` is a string or a list of strings, `components` is False (default): do not add any components, perform region aggregation on single variable or entire list (per variable)
 - `variable` is a string, `components` is True: automatically detect sub-categories (current behaviour)
 - `variable` is a string, `components` is a list: use `components` for the region-only-level values to be added
 - if `variable` is a list, `components` must be False

## Required changes in existing scripts

To continue current behaviour, existing scripts would need to add `components=True`.

danielhuppmann:
@znicholls, you added this feature, in particular the automated detection of components. What are your thoughts here?

znicholls:
I think the switch to having `components` be `False` by default is wise (especially because you can do that without affecting the `check_internal_consistency` API). 

One question, if you use `aggregate_region` with a list of variables, what are you actually doing? Shouldn't you be using `aggregate` and then `aggregate_region`?

danielhuppmann:
We might want to have a similar discussion for [aggregate()](https://pyam-iamc.readthedocs.io/en/stable/api.html#pyam.IamDataFrame.aggregate) at some point, if this function should also take a list of variables (instead the current API taking only a single variable). This might be a more elaborate discussion, though...

However, for the AR6 upload-and-processing feature, we assume that the variable tree is complete, so no need to aggregate any variables from its disaggregated components; i.e., any scenario must explicitly provide `Emissions|CO2`, `Emissions|CO2|FFI`, `Emissions|CO2|AFOLU`, etc. (There are some arguments why this is a good idea).

[...]

Issue #281, opened 2019-10-31
scatter plotting tests fail for new matplotlib


jkikstra:
With `matplotlib version 3.1.1` installed, some pyam plotting tests fail. More specifically, `test_scatter`, `test_scatter_variables_with_meta_color`, `test_scatter_with_lines`, and `test_scatter_meta` fail, because of `IndexError: list index our of range, pyam\plotting.py:605` 
when:
```
pyam\core.py:1293: in scatter
    ax = plotting.scatter(df.dropna(), x, y, **kwargs)
pyam\plotting.py:605: in scatter
    handles = [handles[i] for i in idxs]
```

With `matplotlib version 3.0.2` all of the aforementioned tests pass.

khaeru:
The "quickfix" for this in 7d6b7efc0fc7e05da1fc44bb6cc2ddfc24671ef8 / #280 broke message_ix builds.

The last build for iiasa/message_ix#249, [Travis build #1167](https://travis-ci.org/iiasa/message_ix/jobs/605513801), succeeded; but [build #1180](https://travis-ci.org/iiasa/message_ix/jobs/605927570) failed just now on OS X with the following error caused by pyam:
```
___________________ ERROR collecting tests/test_reporting.py ___________________
ImportError while importing test module '/Users/travis/build/iiasa/message_ix/tests/test_reporting.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_reporting.py:15: in <module>
    import pyam
../../../miniconda/envs/testing/lib/python3.7/site-packages/pyam/__init__.py:1: in <module>
    from pyam.core import *
../../../miniconda/envs/testing/lib/python3.7/site-packages/pyam/core.py:16: in <module>
    from pyam import plotting
../../../miniconda/envs/testing/lib/python3.7/site-packages/pyam/plotting.py:4: in <module>
    import matplotlib.pyplot as plt
../../../miniconda/envs/testing/lib/python3.7/site-packages/matplotlib/pyplot.py:2374: in <module>
    switch_backend(rcParams["backend"])
../../../miniconda/envs/testing/lib/python3.7/site-packages/matplotlib/pyplot.py:207: in switch_backend
    backend_mod = importlib.import_module(backend_name)
../../../miniconda/envs/testing/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #302, opened 2019-12-10
rewrite the tutorials

### LINKED ISSUE #290: add reference in tutorials to where the source file can be found
Add a link to the notebook and data file in the rendered tutorial pages.

Suggested by Julia Steinberger via the `pyam` mailing list [(link)](https://groups.io/g/pyam/message/2)

### LINKED ISSUE #298: Add slightly more explicit example of df.filter(`exclude=False`) in tutorials
Suggestion to slightly improve the example on use of `df.validate(criteria={...}, exclude_on_fail=True) `and `df.filter (exclude=False)`.

This is just above box 20 in the tutorial:
https://pyam-iamc.readthedocs.io/en/stable/tutorials/pyam_first_steps.html#Using-metadata-indicators-for-further-scenario-characterization-&-diagnostics

An example, which I think would be a common use case is this process of 
1. set exclude tags on data
2. filter the data, dropping the excluded data in the process
3. (view table / plot the "good" data)

Example would be helpful as the implementation is slightly counterintuitive in that first we `exclude_on_fail=True`, but then to exclude that data, the argument in `df.filter` is `exclude=False`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This PR reworks most of the tutorials, in particular the "first-steps" tutorial. It updates the tutorial data source to the IAMC 1.5°C data for consistency with the IIASA-db tutorial, and improves consistency of the formatting and notation.

closes #298 
closes #290


danielhuppmann:
@znicholls, can you take please a look at the updated "checking_consistency" notebook?

I created a smaller, more specific example to better highlight what is being checked and why each validation fails. I also replaced the term "database" with "scenario ensemble" to describe the content of an `IamDataFrame` rather than where it is stored, and made some other edits, hoping make it easier for new users...

danielhuppmann:
@byersiiasa, can you take a look whether the updated "first-steps" notebook satisfies the issues #290 and #298? (if you have a minute)

jkikstra:
Hi @danielhuppmann. Thanks for reworking these tutorials! I think you have done a great job of making this first steps tutorial accessible.

Some quite minor points:

1. [filtering]: 'The feature for filtering by **model, scenario or region** are implemented using exact string matching, where ...' should probably be changed to **model, scenario, region, variable, or year** to be more complete and not confuse new users. 
2.

jkikstra:
Hi @danielhuppmann. Thanks for reworking these tutorials! I think you have done a great job of making this first steps tutorial accessible.

Some quite minor points:

1. [filtering]: 'The feature for filtering by model, scenario or region are implemented using exact string matching, where ...' should probably be changed to model, scenario, region, variable (and level), or year to be more complete and not confuse new users.
2. [filtering]: you could add a tiny example under the 'filtering by year' description, e.g `(df_selectedyears = df.filter(range(2010, 2051)))`. If you want, you can make it a box and then also use it to show it as a shorter timeseries next to the full `display_df`
3. [numpy]: you could provide a hyperlink to the numpy package when you mention it when you're dealing with `set_meta_from_data()`

In general, when using pyam in practice for processing purposes, there can be a lot of converting between different forms. It would be good to make it as clear as possible when one has a wide or long format, and when it is a pyam.IamDataFrame and when it is a pandas.DataFrame. A new user would probably assume that everything that goes in as a pyam.IamDataFrame also returns one. Thus, every time that this not the case, it should be very clear (e.g. under 'displaying timeseries data' it is a bit too easy to miss right now). (this also holds for when a pandas.Series [e.g. `overshoot`] or even a multidimensional dataframe [e.g. `df.meta`] are returned, as this can cause quite some confusion because it is not always explicit enough)
In the same way, I think it is important to also emphasize what actions return an altered copy and what actions change the df itself. 
Perhaps an overview table of some central functions with these two binary options (pyam/pandas, and copy/direct alteration) could be very useful for new users, and is more friendly then going into the API or testing it out.

francescolovat:
Hi @danielhuppmann ,

Great tutorial notebook, very illustrative and clear, even for not-very-experienced readers as myself.

Minor comments:

- In `exclude_on_fail` section, the last sentence of this paragraph is not very clear:

> Any scenario (by a particular model) failing the validation criteria is then marked as exclude=True. This "exclusion flag" is implemented in the meta table of the IamDataFrame, which can be used categorization and quantitative indicators (more below). 

I think a preposition is missing before _categorization_. Moreover, I think a clear reference to the meta table section should be included instead of the parenthesis, stating sth like: _"the second latest cell in XX shows the meta table"_ (I found it confusing talking about a meta table whose format it was not clarified until the command `df.meta.head()` at the end of the notebook).

- In the **Categorization assignment** markdown cell I'd list some of the _other possible arguments to be used by the plotting library_. So the users could experiment with them (I'd have liked to do that).
- I'd suggest to swap the order of the cells [14] and [15] (and their associated markdowns) with commands `df.filter(level=1).variables()` and `df.filter(variable='Primary Energy*', level='1-').variables()`, respectively. This is to follow the logic of the **Filtering by variables and levels** markdown cell, which is explaning the difference between filtering by both _variable_ and _level_ or with just _level._

danielhuppmann:
response to @znicholls:
> 'Note that the detailed output (i.e., where the aggregation validation fails) is not shown in a notebook when calling the function within a loop.', maybe add a pointer to a resource which explains how notebooks work for those who are confused by this

Any good reference for this?

Re "handling of bunkers" - this wasn't included in the tutorial before. I'll see if I have time to extend the tutorial when I tackle #299.

znicholls:
> Any good reference for this?

`2. Pretty Display of Variables` of https://www.dataquest.io/blog/jupyter-notebook-tips-tricks-shortcuts/ ?

> Re "handling of bunkers" - this wasn't included in the tutorial before. I'll see if I have time to extend the tutorial when I tackle #299.

Ah ok cool makes sense.

danielhuppmann:
Thank you @jkikstra and @francescolovat for these great suggestions!

Highlights of my changes:
 - added a blue-highlighted cell about the expected return type of `timeseries()`, and some more of these cells throughout the notebook
 - moved the "filter-by-year" section, add an example
 - introduce the `meta` table earlier, so that it can be referenced in the validation and indicators section
 - add a second plot with the kwarg `color='scenario'` in the plotting example

About the suggestion by @jkikstra to create an overview table of functions and their return types - I understand that this is confusing for new users, but I'm worried about adding new parts to the documentation, because this will be difficult to maintain and keep up to date...

byersiiasa:
> @byersiiasa, can you take a look whether the updated "first-steps" notebook satisfies the issues #290 and #298? (if you have a minute)

thanks - yes, nicely implemented!

gidden:
lgtm @danielhuppmann ! I tried restarting the mac CI that failed, not sure why it is failing or if this is a known issue.

danielhuppmann:
thanks @gidden!

this seems to be related to #281 - building the docs on travis installs `kealib`, which also installs `matplotlib-base-2.2.4` on linux but `matplotlib-base-3.1.2` on mac. no idea why this isn't a problem in other PRs

danielhuppmann:
no idea why Appveyor is now failing with ever more confusing error messages - maybe it's because now running on `gidden:master`...? it passed on "my" instance a few commits ago

merging to see if this resolves the issue

PR #304, opened 2019-12-13
clean-up of test suite, merge testing dataframes

danielhuppmann:
# Description of PR

This PR merges two datasets (`test_df` and `meta_df`), which were almost identical and used interchangeably throughout the test suite.

### DIFF
--- a/tests/test_cast_to_iamc.py
+++ b/tests/test_cast_to_iamc.py
@@ -4,43 +4,39 @@ from pyam import IamDataFrame, compare
 
 from conftest import TEST_DTS
 
-def test_cast_from_value_col(meta_df):
+def test_cast_from_value_col(test_df_year):
     df_with_value_cols = pd.DataFrame([
-        ['model_a', 'scen_a', 'World', 'EJ/y', TEST_DTS[0], 1, 0.5],
-        ['model_a', 'scen_a', 'World', 'EJ/y', TEST_DTS[1], 6., 3],
-        ['model_a', 'scen_b', 'World', 'EJ/y', TEST_DTS[0], 2, None],
-        ['model_a', 'scen_b', 'World', 'EJ/y', TEST_DTS[1], 7, None]
+        ['model_a', 'scen_a', 'World', 'EJ/y', 2005, 1, 0.5],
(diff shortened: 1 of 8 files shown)


danielhuppmann:
@gidden & @znicholls, before diving into developing new features, I wanted to clean up the testing dataframes to improve the long-term sustainability of this project.

how to proceed:

0. review this PR, then make another PR for the next round of clean-ups
1. continue, then review all test updates in one PR
2. whatever, no need to review this (as it doesn't change any features or test logic)

znicholls:
I’d do 0. In my opinion the tests are more important than the code and reviewing test changes is hard so should be done carefully. Of course there’s a balance and you have a better sense of the big picture so I’m happy to be overruled.

tests/test_cast_to_iamc.py
```diff
@@ -4,43 +4,40 @@
...
 
-def test_cast_from_value_col(meta_df):
+
+def test_cast_from_value_col(test_df_year):
     df_with_value_cols = pd.DataFrame([
-        ['model_a', 'scen_a', 'World', 'EJ/y', TEST_DTS[0], 1, 0.5],
-        ['model_a', 'scen_a', 'World', 'EJ/y', TEST_DTS[1], 6., 3],
-        ['model_a', 'scen_b', 'World', 'EJ/y', TEST_DTS[0], 2, None],
-        ['model_a', 'scen_b', 'World', 'EJ/y', TEST_DTS[1], 7, None]
+        ['model_a', 'scen_a', 'World', 'EJ/y', 2005, 1, 0.5],
```
gidden: Is there a reason for preferring hardcoded values to referenced variables?

gidden: Same comments below as well

danielhuppmann: the previous implementation tested the same thing twice

0. create an `IamDataFrame` with year-columns
1. test it against the test-data with year column
2. test it again against the datetime-data converted to year columns, which is exactly the same object as 1.

So I simplified it by removing step 2, and then choose the shortest notation:
 `datetime(2005, 6, 17)` > `TEST_DTS[0]` > `2005`

PR #305, opened 2019-12-17
add `method` and `weight` feature to (region) aggregation

### LINKED ISSUE #299: Suggest to change default behaviour of `aggegate_region()` for components

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR extends the functions `aggregate()`, `aggregate_region()` and the respective check-versions by a `method` and `weight` options (weights only for region aggregation). The functionality is based on the branch [peterkolp:region_aggregation_mip_feature](https://github.com/peterkolp/pyam/tree/region_aggregation_mip_feature).

The PR also changes the default of `aggregate_region()` regarding the treatment of components at the region-level as discussed in #299.
> **tl; dr**
> Before, the default behaviour of `aggregate_region()` was to add all sub-categories of `variable` that were not present in any subregion, and deactivating this addition required to set `components=[]`
> Now, the default is `components=False` (not adding sub-categories at the region-level) and `components=True` activates the automatic detection of sub-categories.
> Setting a specific list of variables to be added via `components=<list>` is unchanged.

closes #299


danielhuppmann:
This PR is directed into `IAMconsortium:tests/cleanup` to highlight the differences of the aggregation-features to that prep-work - I'll rebase after a first round of review and merging of #304.

Once this PR is merged, I'll implemented the feature that `variable` can be a list of strings rather than only one string to speed up the processing.

@byersiiasa @gidden @znicholls @peterkolp @zikolach @volker-krey - any volunteers for a review?

pyam/core.py
```diff
@@ -1444,11 +1484,45 @@ def _meta_idx(data):
...
+def _aggregate(df, by, method=np.sum):
     """Aggregate `df` by specified column(s), return indexed `pd.Series`"""
     by = [by] if isstr(by) else by
     cols = [c for c in list(df.columns) if c not in ['value'] + by]
-    return df.groupby(cols).sum()['value']
+    # pick aggregator func (default: sum)
+    return df.groupby(cols)['value'].agg(_get_method_func(method))
+
+
+def _aggregate_weight(df, weight, method):
```
byersiiasa: if this is for weighted average, then surely method is always sum/np.sum? or is this just meant as a placeholder for future methods?

Consider also instead using `np.nansum` such that if a value is missing / nan, the calculation won't return `nan`

pyam/core.py
```diff
@@ -1444,11 +1484,45 @@ def _meta_idx(data):
...
+        raise ValueError('only method `np.sum` allowed for weighted average')
+
+    _data = _get_value_col(df, YEAR_IDX)
+    _weight = _get_value_col(weight, YEAR_IDX)
+
+    if not _data.index.equals(_weight.index):
+        raise ValueError('inconsistent index between variable and weight')
+
+    cols = META_IDX + ['year']
+    return (_data * _weight).groupby(cols).sum() / _weight.groupby(cols).sum()
```
byersiiasa: would you put `.agg(method)` here instead of `.sum()`

byersiiasa: do you also want to limit inputs to the `KNOWN_FUNCS` ?

pyam/core.py
danielhuppmann: 1) The `method` arg is partly anticipating more supported functions in the future, partly for checking that the chosen method is indeed sum in that function (moving some not-so important stuff to lower-level functions, also for reusing `_aggregate_weight()` in other functions in the future and not needing to implement that check in multiple higher-level functions.

2) The `data` table cannot have `nan` (they are removed at initialisation), so this point is moot. If a data value or the weight value is missing (i.e., inconsistent series index's), a `ValueError` is raised.

pyam/core.py
danielhuppmann: only summation is allowed (for the time being) anyway, so imho no need to be more complicated...

gidden:
Restarted CI (mac py36 only)

pyam/core.py
gidden: Agree that we don't need anything more complicated here, but maybe worth putting in some comments about what would need to be changed in the future if more than sum is supported?

pyam/utils.py
```diff
@@ -33,6 +33,8 @@
                           + ['{}{}'.format(i, j) for i, j in itertools.product(
                               string.ascii_uppercase, string.ascii_uppercase)]))
 
+KNOWN_FUNCS = {'min': np.min, 'max': np.max, 'avg': np.mean, 'sum': np.sum}
```
gidden: maybe add a `'mean'` synonym as well?

tests/test_feature_aggregate.py
```diff
@@ -8,6 +8,91 @@
...
+
+    # use other method (max) both as string and passing the function
+    idx = ['model', 'scenario', 'unit', 'year']
+    exp = pd.DataFrame([
+        ['model_a', 'scen_a', 'USD/tCO2', 2005, 10.0],
+        ['model_a', 'scen_a', 'USD/tCO2', 2010, 30.0]
+    ],
+        columns=idx + ['value']
+    ).set_index(idx).value
+    obs = df.aggregate_region('Price|Carbon', method='max')
```
gidden: Maybe I don't see it, but is there a test that explicitly tests the `weight` option?

pyam/utils.py
danielhuppmann: done

tests/test_feature_aggregate.py
danielhuppmann: We have one (and the line above it tests that not adding the weight yields an unexpected answer)

`assert df.check_aggregate_region('Price|Carbon', weight=v) is None`

danielhuppmann:
Thanks @gidden for the comments! Implemented the one minor change & answered the question about the test inline.

About the "how to support other methods that sum with weights" (seems I can't answer inline) - it's not clear to me what a "weighted max" (or similar) would be expected to do, so no idea what to write as a comment. Happy to discuss implementation when an actual use case comes up.

PR #308, opened 2019-12-23
rename `logging` to `logging_utils` trying to fix win-py37 issue on CI

danielhuppmann:
This PR tries to fix an issue on Appveyor (ie Windows) only occurring with Python 3.7. 

```
> pyam\utils.py:21: in <module>
>     logger = logging.getLogger(__name__)
> E   AttributeError: 'module' object has no attribute 'getLogger'
```


danielhuppmann:
seems that the logging wasn't actually the issue, fixed (for now) by cherry-picking commits into #310

PR #314, opened 2019-12-29
Ipcc color update

gidden:

- [x] Tests Added
- [ ] Documentation Added

# Description of PR

The IPCC TSU has updated their preferred SSP/RCP color pallete for AR6. It now exists in a public repo [here](https://github.com/IPCC-WG1/colormaps). This PR updates our color mappings to match this and includes new "SSP only" colors as well. 

Probably best for at least @znicholls to take a look and confirm one or two of the translations with their own info. @danielhuppmann happy for you to give the a-ok as well if you want =)


gidden:
I considered putting in a pr, but thought that this happens so
infrequently, it may just not be worth it (other than having a single point
of reference)

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 # Next Release
 
+- [#314](https://github.com/IAMconsortium/pyam/pull/314) Update IPCC color scheme colors and add SSP-only colors
```
danielhuppmann: can you move this to the section **Individual updates** below, please?

gidden:
should be g2g @danielhuppmann

PR #315, opened 2019-12-31
clean-up of aggregation features

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR implements a number of clean-ups following PRs #305 and #312:
 - the return type of `aggregate()` and `aggregate_region()` is changed to an `IamDataFrame` instance (per suggestion by @jkikstra), previously a timeseries-dataframe
 - the return type of `check_aggregate()` and `check_aggregate_region()` is changed to a `pd.DataFrame` with both expected and actual value (previously only the expected value)
 - adds an `equals()` function (originally used to make tests easier)
 - adds an `pyam.testing.assert_frames_equal()` function (to make tests easier)
 - the tutorial for aggregation, downscaling and consistency checking is reworked (now includes description of regional components, weighted average, setting variables as list, downscaling)
 - the aggregation tests are completely refactored (because of the changed return types, it was easier to completely rewrite rather than figure out how to salvage)


danielhuppmann:
@francescolovat, @fabiosferra, @vruijven, @volker-krey, maybe you have a minute to read through the completely reworked tutorial `aggregating_downscaling_consistency` - in particular whether the behaviour of the new `downscale_region()` function is a good starting point for upcoming work?

danielhuppmann:
ping @gidden @znicholls @jkikstra @francescolovat @fabiosferra

I'd like to make a new release (including the new tutorial) in time for the openmod meeting [1], so I'd appreciate your comments by the end of this week. If there are no objections, I'll merge on Jan 13.

[1] https://wiki.openmod-initiative.org/wiki/Open_Energy_Modelling_Workshop_-_Berlin_2020

pyam/core.py
```diff
@@ -37,11 +37,11 @@
     META_IDX,
     YEAR_IDX,
     IAMC_IDX,
-    SORT_IDX,
-    KNOWN_FUNCS
+    SORT_IDX
 )
 from pyam.read_ixmp import read_ix
 from pyam.timeseries import fill_series
+from pyam.aggregate import _aggregate, _aggregate_region, _agg
```
gidden: Now that these are in their own module, they should not have preceding underscores. that is left for "private" functions that should not be called out of scope of their own module (file).

pyam/core.py
```diff
@@ -169,6 +169,16 @@ def empty(self):
         """Indicator whether ``IamDataFrame`` is empty"""
         return self.data.empty
 
+    def equals(self, other):
+        """Test if two objects contain the same data and meta indicators"""
```
gidden: this docstring should state that the object must be of type IamDataFrame (or subclasses thereof)

pyam/core.py
danielhuppmann: refactor to no-underscore, renamed `_agg` to `groupby` (that is the main purpose and it is clearly distinct from `aggregate`, even if it is not quite the same as `pandas.DataFrame.groupby`) - if you have better ideas...?

danielhuppmann:
thanks @gidden for the useful comments!
 - added a new testing-module with a function `pyam.testing.assert_frame_equal()`
 - refactored the function names in the `aggregate` module
 - added a longer docstring (copying [pandas.DataFrame.equals](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.equals.html))

Minor version will be bumped in the new release

danielhuppmann:
Thanks @jkikstra for your review and comments!

- re the keyword `components` (which is used in both `aggregate()` or `aggregate_region()`): I agree it's not great except that every alternative I can think of is worse... Also, these functions have been implemented in this way for at least a year, so any change should be a real improvement.

 - re `check_internal_consistency()`: the main reason why I didn't change that is laziness to think through what a good return type would be and write enough tests to make sure that there are no weird results in any corner cases. Maybe tackle that in a subsequent PR (cc @znicholls)?

danielhuppmann:
Follow-up to @jkikstra's comment on return type:

One option could be something like

```
> pd.concat(tutorial_df.check_internal_consistency().values(), sort=False)
```
image

jkikstra:
Hi @danielhuppmann Thanks for implementing our suggestions. 

in my eyes your suggestion for the output of `check_internal_consistency()` is a definite improvement, so I would suggest implementing that and then I have no further suggested improvements

pyam/core.py
```diff
@@ -169,6 +169,21 @@ def empty(self):
         """Indicator whether ``IamDataFrame`` is empty"""
         return self.data.empty
 
+    def equals(self, other):
+        """Test if two objects contain the same data and meta indicators
+
+        This function allows two IamdDataFrame instances to be compared against
+        each other to see if they have the same timeseries data and meta
+        indicators. NaNs in the same location of the meta table are considered
+        equal."""
```
znicholls: `other` type should be in docstring

pyam/core.py
```diff
@@ -992,23 +941,27 @@ def check_aggregate_region(self, variable, region='World', subregions=None,
...
 
             if exclude_on_fail:
-                self._exclude_on_fail(diff.index.droplevel([2, 3]))
+                self._exclude_on_fail(_meta_idx(df_region[rows].reset_index()))
 
-            return IamDataFrame(diff, region=region).timeseries()
+            _df = pd.concat(
+                [pd.concat([df_region[rows], df_subregions[rows]], axis=1,
+                           keys=(['region', 'subregions']))],
+                keys=['World'], names=['region'])
```
```suggestion
                keys=[region], names=['subregions'])
```

Or do I misunderstand what is going on here (hard-coding 'World' doesn't look right)?

pyam/testing.py
```diff
@@ -0,0 +1,11 @@
+from pyam import compare
```
znicholls:
```suggestion
from . import compare
```

pyam/aggregate.py
```diff
@@ -0,0 +1,154 @@
...
+            # rename all components to `variable` and aggregate
+            rows = region_df._apply_filters(variable=components)
+            _df = region_df.data[rows].copy()
+            _df['variable'] = variable
+            _data = _data.add(groupby(_df, 'region'), fill_value=0)
+
+    return _data
+
+
+def groupby(df, by, method=np.sum):
```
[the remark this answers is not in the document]
danielhuppmann: refactored to `group_and_agg()`

pyam/aggregate.py
```diff
@@ -0,0 +1,154 @@
...
+
+        for c in components:
+            mapping[c] = variable
+
+    # else, use all variables one level below `variable` as components
+    else:
+        for v in variable if islistable(variable) else [variable]:
+            _components = df._variable_components(v)
+            if not len(_components):
+                logger.info(msg.format(v))
```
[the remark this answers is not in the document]
danielhuppmann: the following two lines didn't do anything if there are no components, but I added `continue` for clarity

pyam/core.py
danielhuppmann: the first part of this is indeed an error (it should use the kwarg `region`), but the name of that index level should be `region`

pyam/core.py

gidden:
@danielhuppmann I share @znicholls opinion on the test refactor here (also not blocking but with a strong preference).

I may not grok the gritty details, but is it possible to write a wrapper function in the tests that would translate the newly returned `IamDataFrame` back to a `pd.DataFrame` in the test's expected format and then apply that in each test a la `pdt.assert_frame_equals(pd_to_pyam_test(obs), exp)`?

pyam/aggregate.py
```diff
@@ -0,0 +1,154 @@
...
+    isstr,
+    KNOWN_FUNCS,
+    META_IDX,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def aggregate(df, variable, components=None, method=np.sum):
+    """Internal implementation of the `aggregate` function"""
```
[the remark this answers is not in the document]
danielhuppmann: no, the public API should be accessed via `IamDataFrame.aggregate()`

pyam/aggregate.py
```diff
@@ -0,0 +1,154 @@
...
+            for c in _components:
+                mapping[c] = v
+
+    # rename all components to `variable` and aggregate
+    _df = df.data[df._apply_filters(variable=mapping.keys())].copy()
+    _df['variable'].replace(mapping, inplace=True)
+    return groupby(_df, [], method)
+
+
+def aggregate_region(self, variable, region, subregions=None,
```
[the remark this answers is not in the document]
danielhuppmann: no, the public API should be accessed via `IamDataFrame.aggregate_region()`

danielhuppmann:
re the comments by @znicholls and @gidden:

0) sorry that this is such a big refactoring - I divided it into a suite of PRs to make it manageable, but it's still huge...

1) API of `check_internal_consistency()`: I added a kwarg `components` so that the behaviour can be controlled and passed through to `check_aggregate_region()`. However, as to whether we should use `True` or `False` as default, detecting components that exist only at the region-level is one use case out of many (it's relevant mainly in the emissions domain). And it would interfere with using (weighted) average or other methods which are equally valid and relevant, so I changed that to `False` with a  purpose.

2) about the tests: I agree that just dropping tests is a really bad idea, but the suite of tests were written so efficiently (a dozen different tests with hardly any documentation about their intent, all feeding into one master testing function with a bunch of if-else-logic to do the comparison), it just wasn't possible for me to disentangle this.
@gidden, your approach sounds smart but if you look at what was actually tested (length of indices, etc.) this won't work. The new test suite with `assert_frame_equal()` is far more explicit.
@znicholls, if you can spell out for each of the removed tests what their intent was, I'll be happy to re-introduce them in a next step.

francescolovat:
Hi @danielhuppmann,

I've been through the `pyam_first_steps.ipynb` tutorial again. 

It really looks nice. You've implemented all our suggestions in a very detailed way. I was really enjoyable to follow the steps in the notebook.

I have no further comments to add to it (I'll keep in mind some formatting features you included, e.g. the blue boxes, to potentially include them also in the `MESSAGEix` tutorials).

I regards of the previous conversation about the refactoring of tests, I have little experience with them. I'm still getting familiar with `pandas.util.testing` methods. Matt's suggestions of a wrapper function looks promising, however it is also a matter of the time left before the _openmod_ meeting.

pyam/aggregate.py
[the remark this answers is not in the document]
znicholls: ah ok. In that case it's tricky because the way they're currently written (without the underscore) suggests that they are in the public API. In https://github.com/danielhuppmann/pyam/pull/14, I put everything in the private API and then use that in the core module. That goes against what @gidden suggested re not using things outside their modules but is something I've seen elsewhere (e.g. pandas plotting stuff https://github.com/pandas-dev/pandas/tree/master/pandas/plotting) so might a suitable compromise given they shouldn't be in the public API?

znicholls:
> 1. sorry that this is such a big refactoring - I divided it into a suite of PRs to make it manageable, but it's still huge...

All good there's always a nasty conflict between time pressure and moving slowly enough that everyone can keep up (you're striking a good balance)

> 2\. API of `check_internal_consistency()`

All makes sense thanks for spelling it out.

> 3\. about the tests

fair, they were definitely not as clear as they should have been (I have since learnt that whilst this 'efficiency' approach looks good, it's actually a terrible idea as it's not explicit enough about what is going on). Let's park this in <issue> for now

pyam/aggregate.py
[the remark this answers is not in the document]
znicholls: same as comment above

pyam/aggregate.py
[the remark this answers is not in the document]
danielhuppmann: thanks, merged your PR

gidden:
Ok all, will merge this now that everything is approved/reviewed/marked for future issues. Thanks @danielhuppmann for the huge effort here and all reviewers!

PR #323, opened 2020-01-20
add io-support for frictionless `datapackage`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds functions (and a unit test) to export and read (zipped) files in the frictionless datapackage format (https://frictionlessdata.io/). It's a first stab exporting/importing only the `data` and `meta` tables, not incorporating metadata (license) or other fancy features like consistency validation.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1288,7 +1288,7 @@ class IamDataFrame(object):
             excel_writer.close()
 
     def to_datapackage(self, path, tmp='tmp'):
-        """Write object to a frictionless datapackage
+        """Write object to a frictionless Data Package
 
         More information: https://frictionlessdata.io
 
@@ -1298,7 +1298,7 @@ class IamDataFrame(object):
 
         Parameters
         ----------
-        path: string
+        path: string or pathlib.Path
             file path
         tmp: string, default 'tmp'
             name for temporary folder to export csv files
@@ -1819,11 +1819,11 @@ def concat(dfs):
 
 
 def read_datapackage(path, data='data', meta='meta'):
-    """Read timeseries data (and meta-indicators) from frictionless datapackage
+    """Read timeseries data (and meta-indicators) from frictionless Data Package
 
     Parameters
     ----------
-    path: path to datapackage
+    path: path to Data Package
         passed to ``datapackage.Package()``
     data: str, default `data`
         resource containing timeseries data in IAMC-compatible format
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -10,7 +10,12 @@ import pandas as pd
 
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from datapackage import Package
+
+try:
+    from datapackage import Package
+    HAS_DATAPACKAGE = True
+except IoError:
+    HAS_DATAPACKAGE = False
 
 try:
     import ixmp
@@ -1303,6 +1308,9 @@ class IamDataFrame(object):
             file path
         """
 
+        if not HAS_DATAPACKAGE:
+            raise ImportError('required package `datapackage` not found!')
+
         with TemporaryDirectory(dir='.') as tmp:
             # save data and meta tables to a temporary folder
             self.data.to_csv(Path(tmp) / 'data.csv', index=False)
@@ -1808,7 +1816,7 @@ def concat(dfs):
 
 
 def read_datapackage(path, data='data', meta='meta'):
-    """Read timeseries data (and meta-indicators) from frictionless Data Package
+    """Read timeseries data and meta-indicators from frictionless Data Package
 
     Parameters
     ----------
@@ -1820,6 +1828,9 @@ def read_datapackage(path, data='data', meta='meta'):
         (optional) resource containing a table of categorization and
         quantitative indicators
     """
+
+    if not HAS_DATAPACKAGE:
+        raise ImportError('required package `datapackage` not found!')
 
     package = Package(path)
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from datapackage import Package
 
 try:
@@ -1287,7 +1288,7 @@ class IamDataFrame(object):
         if close:
             excel_writer.close()
 
-    def to_datapackage(self, path, tmp='tmp'):
+    def to_datapackage(self, path):
         """Write object to a frictionless Data Package
 
         More information: https://frictionlessdata.io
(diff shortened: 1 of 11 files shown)


danielhuppmann:
Following the discussions at the openmod workshop, this PR implements a first stab at frictionless-datapackage-input/output of IAMC-like-timeseries data (plus categorisations and quantitative indicators).

@Ludee, @cmutel, @rgieseke [1], any comments? is anything critical missing?

[1] I played around with [pandas-datapackage-reader](https://github.com/rgieseke/pandas-datapackage-reader) at first, but the direct path using `datapackage` seemed easier in the end...

pyam/core.py
```diff
@@ -1284,6 +1287,33 @@ def export_meta(self, excel_writer, sheet_name='meta'):
         if close:
             excel_writer.close()
 
+    def to_datapackage(self, path='datapackage.zip', tmp='tmp'):
+        """Export object to a
```
rgieseke: ... "Data Package" missing?

setup.py
```diff
@@ -29,6 +29,8 @@
     'matplotlib<=3.0.2',
     'seaborn',
     'six',
+    'pathlib',
```
rgieseke: Which Python version is Pyam supporting? `pathlib` should be part of Py3.4+ or so i think ...

danielhuppmann: indeed, thanks for pointing this out (CI is failing for other reasons)

pyam/core.py
danielhuppmann: sorry about this, the docstring got lost during committing

rgieseke:
As Data Packages have become used in the energy modelling community this could be a very helpful feature. I'm still not sure whether a format with metadata in separate files (datapackage.json) is perfect for science (were files end up being mailed around), though the ZIP format might help with that.

In general i'm not sure whether the wide format of pyam (years on "x-axis") works nice with Data Package column based validation tools and metadata format tools, as this would require updating the metadata if a new year is added to the data.

danielhuppmann:
Thanks for your review comments @rgieseke!

> In general i'm not sure whether the wide format of pyam (years on "x-axis") works nice with Data Package column based validation tools and metadata format tools, as this would require updating the metadata if a new year is added to the data.

pyam works internally with a **long format** of timeseries data, and the implementation of `to_datapackage()` writes the csv-file in that way. This is in contrast to the function [IamDataFrame.to_csv()](https://pyam-iamc.readthedocs.io/en/latest/api.html#pyam.IamDataFrame.to_csv), which writes the more human-readable (and smaller if unzipped) **wide format**.

By the way, we spent some time on casting different table structures into an IAMC-compatible format, and this should pretty much work out-of-the-box:

 - long or wide
 - variables as columns
 - identifiers spread out over multiple columns (rather than the concatenated „variable“ column that we use)
 - a weird R „feature“ that adds an X to an integer column name when writing to file

No documentation yet, but a suite of [unit tests](https://github.com/IAMconsortium/pyam/blob/master/tests/test_cast_to_iamc.py)... If there is demand for a docs page of this, please make an issue!

rgieseke:
>   long or wide

Awesome, then there shouldn't be any problem with a Data Package i think (which could also easily be *read* by pandas-datapackage-reader ...

pyam/core.py
```diff
@@ -1284,6 +1287,48 @@ def export_meta(self, excel_writer, sheet_name='meta'):
         if close:
             excel_writer.close()
 
+    def to_datapackage(self, path, tmp='tmp'):
+        """Write object to a frictionless datapackage
```
rgieseke:
```suggestion
        """Write object to a frictionless Data Package
```

That's how they spell it in their docs.

pyam/core.py
```diff
@@ -1284,6 +1287,48 @@ def export_meta(self, excel_writer, sheet_name='meta'):
...
+
+        More information: https://frictionlessdata.io
+
+        Returns the saved :class:`datapackage.Package`. When adding metadata
+        (descriptors), please follow the `template` defined by
+        https://github.com/OpenEnergyPlatform/metadata
+
+        Parameters
+        ----------
+        path: string
```
rgieseke:
```suggestion
        path: string or pathlib.Path
```

a `Path` of a `Path` remains the same `Path` so no changes required below

pyam/core.py
```diff
@@ -1771,3 +1816,38 @@ def concat(dfs):
         else:
             _df.append(df, inplace=True)
     return _df
+
+
+def read_datapackage(path, data='data', meta='meta'):
+    """Read timeseries data (and meta-indicators) from frictionless datapackage
```
rgieseke:
```suggestion
    """Read timeseries data (and meta-indicators) from frictionless Data Package
```

pyam/core.py
```diff
@@ -1771,3 +1816,38 @@ def concat(dfs):
...
+
+    Parameters
+    ----------
+    path: path to datapackage
```
rgieseke:
```suggestion
    path: path to Data Package
```

pyam/core.py
```diff
@@ -1284,6 +1287,48 @@ def export_meta(self, excel_writer, sheet_name='meta'):
...
+        https://github.com/OpenEnergyPlatform/metadata
+
+        Parameters
+        ----------
+        path: string
+            file path
+        tmp: string, default 'tmp'
+            name for temporary folder to export csv files
+            (deleted after saving datapackage)
+        """
```
cmutel: If you creating a tempdir only to delete, you can use `tempfile.TemporaryDirectory` as a context manager (example)[https://docs.python.org/3/library/tempfile.html#tempfile-examples], and let the stdlib manage deleting it.

pyam/core.py
```diff
@@ -1284,6 +1287,48 @@ def export_meta(self, excel_writer, sheet_name='meta'):
...
+
+        path, tmp = Path(path), Path(tmp)
+        if tmp in list(tmp.parent.iterdir()):
+            msg = 'folder `{}` already exists in this directory!'
```
cmutel: If you still want to use `tmp`, then here you could just do `if tmp.exists():`.

pyam/core.py
```diff
@@ -1284,6 +1287,48 @@ def export_meta(self, excel_writer, sheet_name='meta'):
...
+            raise ValueError(msg.format(tmp))
+
+        # save data and meta tables to a temporary folder
+        tmp.mkdir()
+        self.data.to_csv(tmp / 'data.csv', index=False)
+        self.meta.to_csv(tmp / 'meta.csv')
+
+        # cast tables to datapackage
+        package = Package()
+        package.infer('{}/*.csv'.format(tmp))
```
cmutel: Maybe you have it elsewhere, but the documentation for the metadata is not covered here. I guess there is some magic with the `meta.csv` file? Hard to know what to do to change e.g. the license for newcomers.

pyam/core.py
danielhuppmann: thanks for the suggestion! implemented, much nicer...

pyam/core.py
danielhuppmann: true...

pyam/core.py
danielhuppmann: The `meta` table is used for categorization and quantitative indicators of the scenarios. For the IPCC SR15, we used that for warming categories (Below 1.5°C, 1.5°C with low overshoot, etc.) or cumulative CO2 emissions over the century. [Read the docs page](https://pyam-iamc.readthedocs.io/en/stable/data.html#the-meta-table)

Currently, `pyam` doesn't support metadata in the frictionless sense (license, sources, etc.).  That's why I implemented `to_datapackage()` such that it returns the `Package` (with only data and meta table), and it's up to the user to add descriptors in his/her workflow.

If this is picked up by a broader range of users with more advanced requirements like proper metadata handling, I'd be happy to discuss implementation. But as a pragmatic first step, I'd not add it to this PR.

danielhuppmann:
thanks @rgieseke and @cmutel for the helpful review comments!

tests/test_io.py
```diff
@@ -3,7 +3,8 @@
 import numpy as np
 import pytest
 
-from pyam import IamDataFrame
+from pyam import IamDataFrame, read_datapackage
+from pyam.testing import assert_frame_equal
```
gidden: would suggest

`from pyam.testing import assert_frame_equal as pyam_assert_frame_equal`

to delineate from other asserts

setup.py
```diff
@@ -29,6 +29,7 @@
     'matplotlib<=3.0.2',
     'seaborn',
     'six',
+    'datapackage'
```
gidden: move to optional deps

pyam/core.py
```diff
@@ -8,6 +8,10 @@
 import numpy as np
 import pandas as pd
 
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from datapackage import Package
```
gidden: let's make this optional with:

```
try:
    from datapackage import Package
    HAS_DATAPACKAGE = True
except IoError:
    HAS_DATAPACKAGE = False
```

and then where datapackage is used either have a decorator (like we do in tests) or an explicit check for `HAS_DATAPACKAGE`

tests/test_io.py
danielhuppmann: renamed that function to `assert_iamframe_equal` instead

setup.py
danielhuppmann: done

pyam/core.py

danielhuppmann:
@gidden, I found the relevant command in AppVeyor to install optional dependencies, but not for Travis CI (which is now failing because of the missing `datapackage`). Can you point to the right line I need to add?

danielhuppmann:
ok, @gidden, seems the optional dependency now also works on Travis

danielhuppmann:
now doing the optional-dependency install correctly via the makefile...

Issue #317, opened 2020-01-11
Aggregation tests


znicholls:
In https://github.com/IAMconsortium/pyam/pull/315, the aggregation tests were removed. They should be added back in future in a way which makes the intent of the tests clearer. Old testing code is below

```python

```

danielhuppmann:
Thanks @znicholls - though I didn't just remove all aggregation tests, most of them were just refactored (and ordered so that it's all tests for `[check_]aggregate()`, then `[check_]aggregate_region()`, then `check_internal_consistency()`, with the same order of features tested).

A useful first step is to make a list of all tests that were removed, with a line or two indicating the feature or specific corner case that was covered by the test (which was sometimes not obvious to me). I can then go through and indicate whether this is covered in one of the new tests or if it was dropped. Then we can re-insert those tests that were indeed dropped during the refactoring.

gidden:
I agree on the first step - identifying which tests were *not* moved over during the refactor. I can try to help then put those missing back in during my now very productive train rides between Berlin and Vienna =). When we get a list here, I'll check back in.

PR #333, opened 2020-02-15
Df ops

gidden:
# Description of PR

Originally in #276, @znicholls wrote:

> This PR supercedes #229 and adds a subtraction operation for IamDataFrames. It is specifically very limited in scope as there are likely many kinks to iron out. Further PRs can add extra operations, ``other`` types etc.

This PR starts as rebased on #276 and adds additional ops.

- [ ] Tests Added
- [ ] Documentation Added

```
```


pyam/core.py
```diff
@@ -1565,6 +1537,76 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
         if not inplace:
             return ret
 
+    def subtract(self, other, axis, new_name, ignore_meta_conflict=False):
```
danielhuppmann: 1) suggest to use `axis='variable'` as default

2) suggest to not have this operation with another `IamDataFrame`as input, but specify `a` and `b` directly to be used as filters (i.e, `a - b`).

Say you have an `IamDataFrame` with total CO2 emissions and energy-specific CO2 emissions, and you want to compute the difference and append it to the original data.

With the currently implemented API, you'll have
```
co2_total = df.filter(variable='Emissions|CO2')
co2_ene = df.filter(variable='Emissions|CO2|Energy')
diff = co2_total.subtract(co2_ene, axis='variable', name='Emissions|CO2|Other')
df.append(diff, inplace=True)
```

compared to
```
df.subtract(name='Emissions|CO2|Other', a='Emissions|CO2', b='Emissions|CO2|Energy',
            axis='variable', append=True)
```

tests/conftest.py
```diff
@@ -67,6 +67,127 @@
     columns=IAMC_IDX + [2005, 2010],
 )
 
+mg_ascen = ['MSG-GLB', 'a_scen']
```
danielhuppmann: I made this comment previously, I think this dataframe is not idea for testing - too large (why multiple regions?), non-standard variables (CO2 emissions from cars and tar), why use `MSG` and `IMG` as model names?

znicholls: My two cents on this: this df is meant for integration tests, not unit tests. Integration tests are always harder, because they’re testing more things but they can catch bugs which unit tests don’t. Of course, totally up to you guys whether you want integration tests and the difficulty they add or not. 

As an integration test, it needs to cover more stuff. Hence multiple regions, weird variables (so you have to think carefully about what is going on, it’s not obvious) and model names that also require some thought.

gidden: I have no strong opinion here, but would note that there are lots of tests that already depend on these. I *do* have a strong opinion for more tests than fewer. IMO we should spend our efforts primarily on in library code and secondarily on the tests (unless the effort is in making more tests!).

pyam/core.py
```diff
@@ -1565,6 +1537,76 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+        other : pyam.IamDataFrame
+            Object containing timeseries data to subtract
+
+        axis : str
+            Column to use when subtracting the two sets of data
+            (e.g. ``variable``)
+
+        new_name : str
+            String to write in ``join_col`` in the output timeseries e.g.
+            (``variable 1 - variable 2``)
```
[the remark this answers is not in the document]
gidden: docstring has been updated

pyam/core.py
gidden: > * suggest to use `axis='variable'` as default
> *

don't disagree
> suggest to not have this operation with another IamDataFrameas input, but specify a and b directly to be used as filters (i.e, a - b).

Here, I do disagree. Our original goal with the `IamDataFrame` class was to avoid direct inheritance, but to make the interface as close to a `pd.DataFrame` as possible using composition (as suggested by [pandas devs themselves](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures).

The method signature as implemented closely mimics the [`pd.Dataframe.subtract()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.subtract.html) method signature, which i would suggest keeping.

I think a middle way solution here would be to also add methods like:

```
pyam.subtract(df, a=..., b=..., ...)
```

which behaves as you suggest

danielhuppmann: I agree with the high-level principle to follow pandas, but we should also consider where our use cases diverge. And with maths operations, I think that within-`IamDataFrame` are a more common use case than comparing two instances.

In response to your middle-way suggestion, I'd think that switching the behaviour would make a lot more sense:
 - `IamDataFrame.subtract(name, a, b, axis)` for the within-dataframe operations
 - `pyam.substract(df1, df2)` for the across-dataframe operations

tests/conftest.py
danielhuppmann: Of course I fully agree to have more tests, and also have integration tests with more elaborate datasets. But I also think we should spend a few extra minutes on devising useful datasets for the tests, to avoid situations like #315 (and pending #317) where maintenance becomes difficult (impossible).

My main concern here is not to use "real" model names with made-up data and when using "realistic" variables, stay close to used conventions. 

And @gidden, this test data is currently not used in `master` and the tests in this PR use actual data comparison sparsely, so changing it shouldn't take more than 5 minutes.

pyam/core.py
gidden: It seems we then disagree here. The position I come from is that every operation function on an object I have ever seen or written myself comes in the form of `object.op(other, ...)`. 

Given that we

a) agree that in general operations are useful
b) disagree on the exact interface for those operations

I would suggest then that we provide here a warning that this is an experimental feature which is volatile and may change soon.

tests/conftest.py
gidden: Ok, I will then ask @znicholls to address this at his earliest convenience

znicholls: Will be a while (weeks to months) before I find time for this. Given this df isn’t in master anymore, I’d suggest writing it however suits the use case you’re testing here, and I’ll deal with bringing something back in when I get to #315.

gidden:
Ok, everyone. Please see here a revised implementation. Notably, all `ops` specific code now lives in `ops.py`. There are corresponding (minimal) tests in `test_ops.py`. I tried to keep this interface as general as possible so we can separate the discussion to how `IamDataFrame.subtract()` should look vs. "how should we do subtraction". 

This is hopefully flexible enough that we can implement both ways and debate the relative benefits over beer =)... noting that  have at the moment chosen to implement *my* desired way!

tests/conftest.py
gidden: Ok @znicholls - I reimplemented tests using exsiting data in master

pyam/ops.py
```diff
@@ -0,0 +1,58 @@
...
+class BinaryOp(object):
+
+    def __init__(self, a, b, ignore_meta_conflict=False):
+        self.a_df = a
+        self.b_df = b
+        
+        # Generate meta data merge if two dataframes are not the same,
+        # and do so early to fail early
+        # TODO: merge_meta needs tests
+        self.meta = pd.DataFrame() if a is b else \
```
danielhuppmann: 1) if this is an attempt to bypass merging `meta` in the `subtract_inplace()` function, this won't work because `a` and `b` are filtered for different values, hence different.

2) on a meta-level (pun intended), say you subtract scenario a from scenario b, what should happen to the `meta` table? New blank meta table, checking if they are equivalent and keeping them? Should something be written to the log?

pyam/core.py
```diff
@@ -1565,10 +1539,79 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+            The type of ``other`` is not yet supported
+        """
+        op = BinaryOp(self, other, ignore_meta_conflict)
+        data, meta = op.calc(subtract_op, axis, new_name)
+        ret = IamDataFrame(data)
+        for col in meta:
+            ret.set_meta(meta[col])
+        return ret
+
+    def subtract_inplace(self, a, b, axis='variable', new_name=None,
```
danielhuppmann: suggest to switch order of arguments so that the function can be called like
```
subtract_inplace('Primary Energy', 'Primary Energy|Fossil', 'Primary Energy|Non-fossil')
```
like `a - b = c`

pyam/core.py
```diff
@@ -1565,10 +1539,79 @@ def map_regions(self, map_col, agg=None, copy_col=None, fname=None,
...
+        ValueError
+            ``self`` or ``other`` contains more than value for ``join_col``
+
+        NotImplementedError
+            The type of ``other`` is not yet supported
+        """
+        op = BinaryOp(self, other, ignore_meta_conflict)
+        data, meta = op.calc(subtract_op, axis, new_name)
+        ret = IamDataFrame(data)
+        for col in meta:
```
danielhuppmann: why can this be not just `self.meta = meta`? Like done here:
https://github.com/IAMconsortium/pyam/blob/7d46b2c1c3a46d799c1b09dbabc9dbe88e80ee07/pyam/core.py#L1020

tests/test_ops.py
```diff
@@ -0,0 +1,228 @@
...
+    # exp value
+    exp = df.filter(variable='Primary Energy').copy()
+    exp['value'] = [0.5, 3.0]
+    exp['variable'] = [axis_value, axis_value]
+
+    # test append=False
+    idx = LONG_IDX + ['value']
+    obs = df.subtract_inplace(a, b, axis=axis, new_name=axis_value,
+                              append=False)
+    assert_frame_equal(obs.data[idx], exp.data[idx])
```
danielhuppmann: as above, should be possible to use `pyam.testing.assert_iamframe_equal()` here

tests/test_ops.py
```diff
@@ -0,0 +1,228 @@
...
+    assert_frame_equal(obs.meta, exp.meta)
+
+    # test append=True
+    obs = df.subtract_inplace(a, b, axis=axis, new_name=axis_value,
+                              append=True)
+    assert obs == None
+    obs = df.filter(variable=axis_value)
+    assert_frame_equal(obs.data[idx], exp.data[idx])
```
danielhuppmann: as above, should be possible to use `pyam.testing.assert_iamframe_equal()` here

tests/test_ops.py
```diff
@@ -0,0 +1,228 @@
...
+    axis = 'variable'
+    axis_value = 'Non-Coal PE'
+    obs = a.subtract(b, axis=axis, new_name=axis_value)
+    
+    # test data
+    exp = a.copy()
+    exp['value'] = [0.5, 3.0]
+    exp['variable'] = [axis_value, axis_value]
+    idx = LONG_IDX + ['value']
+    assert_frame_equal(obs.data[idx], exp.data[idx])
```
danielhuppmann: should be possible to use `pyam.testing.assert_iamframe_equal()` here

tests/test_ops.py
```diff
@@ -0,0 +1,228 @@
...
+    # test meta
+    assert obs_meta.empty
+    assert_frame_equal(a.meta, exp.meta)
+    
+    
+def test_calc_subtract_axis_scenario(test_df_year):
+    a = test_df_year.filter(scenario='scen_a', variable='Primary Energy')
+    b = test_df_year.filter(scenario='scen_b', variable='Primary Energy')
+    
+    op = BinaryOp(a, b)
```
danielhuppmann: this would be `a.subtract(b, axis='scenario')`, right?

tests/test_ops.py
```diff
@@ -0,0 +1,228 @@
...
+    axis = 'scenario'
+    axis_value = 'scen_a - scen_b'
+    obs_data, obs_meta = op.calc(subtract_op, axis=axis, axis_value=axis_value)
+
+    # no meta test needed as meta is being messed with here (scenarios)
```
danielhuppmann: if you expect a new `meta` dataframe (with the correct index and column `exclude=False`), you should still test that explicitly

pyam/utils.py
```diff
@@ -268,6 +268,65 @@ def sort_data(data, cols):
...
+
+    Raises
+    ------
+    ValueError
+        Values are in conflict and ``ignore_meta_conflict`` is ``False``
+
+    Returns
+    -------
+    pd.DataFrame
+        Joined metadata tables
```
danielhuppmann:
```suggestion
        Joined ``meta`` tables
```
Trying to avoid the term "metadata" to reduce confusion with actual metadata like license, source, etc.

pyam/ops.py
gidden: This is only hit if the dataframes are the exact same object (i.e., `df.subtract(df)`) - this was mostly carried over from previous implementations. I'm also happy just remove this - all tests (except for checking that its empty) pass. Let me know what you think!

pyam/core.py
gidden: done!

pyam/core.py
gidden: done!

tests/test_ops.py
gidden: done

tests/test_ops.py
gidden: done

tests/test_ops.py
gidden: done

tests/test_ops.py
gidden: done

pyam/utils.py
gidden: done

pyam/ops.py
gidden: removed - as implemented the merge always happens. we can optimize later.

danielhuppmann: 1. why would you subtract two times the exact dataframe - that would be all zero, right?

2. if you do a within-dataframe operation, you can skip that (which you should do for speed and sanity)

3. The larger point of my comment remains - if you are doing an operation on variable, region or unit (does that even make sense?), one can easily do `merge_meta()`. But what should happen when there is a new model or scenario created?

gidden: Agree on 1 and 2.

On 3 - great question. 

I would argue that operations with `axis in ['model', 'scenario']` should have undefined (no) metadata. By definition, metadata is keyed on these attributes. Any desired metadata on this `new_name` should be added by the user, right? 

E.g., in which temperature category bin does `scen1 - scen2` reside? Doesn't make sense, shouldn't be in the `meta` table. It would be super nice to be able to do

```
a = df.filter(scenario='baseline', keep=False)
b = df.filter(scenario='baseline')
result = a.subtract(b, new_name='difference from baseline')
result.bar_plot()
```

These kinds of operations are valuable (e.g., diff from a baseline), but don't need to add metadata. I believe we are in agreement here.

gidden:
back to `_inplace` per offline discussions @danielhuppmann

znicholls:
This is how we handle ops in scmdata if helpful https://github.com/openscm/scmdata/blob/master/src/scmdata/ops.py

gidden:
Finally (and happily) closing in favor of #527

Issue #335, opened 2020-02-18
Refactoring `data` handling


znicholls:
As brought up in https://twitter.com/daniel_huppmann/status/1229689895649710080, as we move to bigger datasets it's not clear that holding the data in 'long form' will be the easiest way to do things. @danielhuppmann we've started playing with holding the data in different forms in memory in https://github.com/lewisjared/scmdata, particularly https://github.com/lewisjared/scmdata/pull/21 and https://github.com/lewisjared/scmdata/pull/26. I'm not sure what the best way to proceed is, perhaps we continue to fiddle in scmdata and once we've found a moderately stable solution we work out how to do something similar in pyam?

danielhuppmann:
Thanks @znicholls for raising this issue, fine to test it in `ScmDataFrame` first. Still, can you outline the options (ideally in the descriptions above and delete my comment) or add a reference if that list already exists - I guess the simplest option is using `pd.Series` with index instead of `pd.DataFrame` (should be smaller in memory and faster), more advanced is moving to `xarray`, other options?

khaeru:
@danielhuppmann asked me to comment here.

- Using pandas.Series with pandas.MultiIndex is usually faster than pandas.DataFrame.
  - Internally, pandas uses similar data storage for the contents of Index and subclasses, as it does for the data contents of a DataFrame. See the section “The Internal Representation of a DataFrame” in [this post](https://www.dataquest.io/blog/pandas-big-data/), which gives a good explanation (h/t @zikolach).
  - Because much of this code is carefully optimized by the pandas developers, and also in C, then it will be more complete and performant than any attempt to emulate it in (pure) Python.
  - pandas should be as performant for a 'wide' DataFrame with (e.g.) 10⁴ rows and 10 columns for a 'year' dimension, as it is for a 'long' Series with 10⁴×10=10⁵ elements and 'year' as one level in a MultiIndex. However, you should write tests (see below) and check this for your typical use cases in order to choose one. See [pytest-benchmark](https://pytest-benchmark.readthedocs.io).
  - Using base pandas.Series instead of a complex derived type also allows optimized calculation methods, that are designed to operate on Series, to work on your data.
- In `ixmp.reporting`, we _want_ to use xarray.Dataset / .DataArray for multidimensional data.
  - However, ixmp data tends to be very sparse; we have quantities with 10–15 dimensions and ~10⁷ possible entries that are only 0.1% or 0.01% full.
[...]

PR #337, opened 2020-02-19
IamDataFrame: fail when initialized with n/a entries in columns

### LINKED ISSUE #336: Initializing with `None` entries in IAMC_IDX columns
# Problem description

As (again) identified by @fonfon, initialising an `IamDataFrame` with `None`/`nan` in the `IAMC_IDX` columns (model, scenario, region, variable, unit) will silently drop these columns. This is necessary because `pandas` has issues with `None`, but it can lead to unexpected behaviour.

# Suggested solution

Separate this section of the code https://github.com/IAMconsortium/pyam/blob/7d46b2c1c3a46d799c1b09dbabc9dbe88e80ee07/pyam/utils.py#L250 
into two parts:
1) remove all `nan` in the `value` column
2) raise an error if `nan` are present in the remaining dataframe

## Alternative

Add a keyword argument to use a default value, maybe an empty string (`''`)?

## Test

The test could be based on this exmaple:
https://github.com/IAMconsortium/pyam/blob/7d46b2c1c3a46d799c1b09dbabc9dbe88e80ee07/tests/test_core.py#L43

fonfon:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

When initializing an IamDataFrame, all rows with n/a entries were dropped silently.
This is expected behavior for the `value` column, but not for other columns like `unit`, where users should become aware of a problem in their input data.

Detailed changes:
- raise error when initializing IamDataFrame with n/a columns
- developer setup documentation: added virtualenv
- updated tests
- removed some unused imports

closes #336


tests/test_core.py
```diff
@@ -138,13 +151,7 @@ def _format_datetime_col(self):
...
-    tdf = test_pd_df.copy()
-    tdf["extra_col"] = "test_val"
-    tdf["extra_col"] = np.nan
-
-    res = IamDataFrame(tdf)
-    assert "Primary Energy|Coal" not in res.variables().tolist()
-
+    IamDataFrame(data=df_empty)
     drop_message = (
         "Formatted data is empty! (perhaps there is a column full of nans?)"
```
danielhuppmann: With this change, a column full of nan's would raise an error, so can you please change this message (and the test)
https://github.com/IAMconsortium/pyam/blob/7d46b2c1c3a46d799c1b09dbabc9dbe88e80ee07/pyam/utils.py#L259

to 'New IamDataFrame is empty!'

Similar to this warning message:
https://github.com/IAMconsortium/pyam/blob/7d46b2c1c3a46d799c1b09dbabc9dbe88e80ee07/pyam/core.py#L1124

pyam/utils.py
```diff
@@ -245,9 +243,12 @@ def convert_r_columns(c):
         df = pd.melt(df, id_vars=IAMC_IDX + extra_cols, var_name=time_col,
                      value_vars=sorted(melt_cols), value_name='value')
 
-    # cast value columns to numeric, drop NaN's, sort data
+    # cast value columns to numeric, drop NaN values, sort data
```
danielhuppmann:
```suggestion
    # cast `value` to numeric and drop nan, assert no nan in other cols
```

RELEASE_NOTES.md
```diff
@@ -9,6 +9,7 @@ It also adds experimental support of the frictionless `datapackage` format.
 
 - [#324](https://github.com/IAMconsortium/pyam/pull/324) Enable compatibility with pandas v1.0
 - [#323](https://github.com/IAMconsortium/pyam/pull/323) Support import/export of frictionless `datapackage` format
+- [#337](https://github.com/IAMconsortium/pyam/pull/337) IamDataFrame to throw an error when initialized with n/a entries in columns other than `value`
```
gidden: I think this should be up in the `# Next release` section

gidden:
lgtm, thanks @fonfon!! - @danielhuppmann I'll leave you to merge after edits

danielhuppmann:
thanks @fonfon, can you please rebase, go through remaining comments and move the release-notes entry.

I also added a "closes" statement to the description for keeping the PR and issue in sync.

tests/test_core.py
fonfon: If the warning refers to IamDataFrame we'd have to make sure this function is only used when initializing IamDataFrame.
This is the case currently but it might change. If we are sure that it will not change then the `format_data` function should be converted to a method of `IamDataFrame`.

For now I just changed the warning message to `'Formatted data is empty!'` - I'm happy to further adapt the PR if you think `format_data` should be moved to `IamDataFrame`.

fonfon:
Thanks for the hints! I updated the PR according to your suggestions. Except for the warning message which I explained in the comment above;

tests/test_core.py
danielhuppmann: That warning message is also fine, thanks!

In light of discussions in #335 and moving to `xarray` asbackend, there will be much more refactoring of this part anyway, so let's keep this PR nice and tidy.

PR #341, opened 2020-03-03
Use `pint` and IIASA-ene-units repo for unit conversion

### LINKED ISSUE #340: Use `pint` for unit conversion and import IIASA-energy-units repo
# Description

The current implementation of [convert_unit()](https://pyam-iamc.readthedocs.io/en/stable/api.html#pyam.IamDataFrame.convert_unit) requires to specify both the unit mapping and the conversion factor as a `dictionary`.

`df.convert_unit({'current_unit': ('target_unit', value)})`

# Suggestion

Following the nice work by @khaeru with the new reporting module for [ixmp](https://github.com/iiasa/ixmp), I suggest to use `pint` as an alternative way to call unit conversion and also change the API, e.g.

`df.convert_unit(current_unit, target_unit, factor=None)`

where the function would use the `factor` (if given) and try to use `pint` otherwise.

## Question 0:

Any objections to the API change? 

[I'd implement it in a way that current behaviour is supported for another release cycle by checking if `current_unit` is a dictionary.]

If there are objections, an alternative API for using `pint` unit conversion could be:

`df.convert_unit({'current_unit': 'target_unit'})`

## Question 1: 

Should this be a hard dependency or an optional one (as we did for `datapackage-py`)?

## Question 2: 

@khaeru also set up an [energy-units repo](https://github.com/iiasa/energy-units) with non-SI unit conversion factors often encountered in integrated assessment, energy systems modelling, transport, etc.

We could import that mapping by default into `pyam` either 
1. directly loading via [raw.githubusercontent.com](https://raw.githubusercontent.com/iiasa/energy-units/master/definitions.txt)
2. importing the repo as a submodule and try https://dependabot.com to update the submodule whenever the `energy-units` repo is updated

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

closes #340

# Description of PR

This PR adds the [pint](https://pint.readthedocs.io/) package as a dependency for easier unit conversion.

## Changes

The PR implements the outcome of the discussion in #340:
- change API to `convert_unit(self, current, to, factor=None, inplace=False)` (current API supported for another release cycle
- add `pint` as hard dependency
- add [energy-units repo](https://github.com/iiasa/energy-units) as a submodule and load `definitions.txt` on first use of `pint` for unit conversion

## Possible issues going forward

1. The new API removes the option to do multiple conversion in one operation. This can be added in the future by allowing `current`, `to` and `factor` to be lists (I'd prefer to not make this PR much longer).
2. The future development of the [energy-units repo](https://github.com/iiasa/energy-units) is not yet clear (renaming, distribution as package). This PR implements it as a submodule for now and I'll try to set up https://dependabot.com for keeping this in sync. This can be changed as the future of a solution for shared unit-mappings becomes clearer.


danielhuppmann:
@byersiiasa @Rlamboll, you commented on the related issue - if you have a moment to look at this PR and whether it aligns with your ideas, that would be great! thanks!

danielhuppmann:
thanks @znicholls for the comments!

> I'm worried that pyam.units.registry could confuse people.

I see your point about confusion. I tried to first set this up as a property of the submodule rather than a function, but this doesn’t work (only classes can have properties, it seems). But overall, I still think that having one global default registry makes more sense than having a unique registry attribute for each IamDataFrame (which would need to be duplicated every time one applies `filter()` etc.

If we have a use case where multiple registries are required, we can add this as a kwarg to `convert_unit()` so a user can be clear which registry to use in which case.

For clearing up the confusion, I see two options:

0. rename that function to `default_registry()`

1. initialize a global registry on loading `pyam` (wanted to avoid this), then this can be obtained via `pyam.units.REGISTRY` (or similar)?

2. add a `UnitConverter` class similar to what you have done in open-scm, so that this could be implemented as a property. Seems a bit excessive maybe? And also confusing - would every `pyam` instance have one global class instance so that you can call it via `pyam.UnitConverter.registry`?

@khaeru @gidden, any thoughts?

> Is the reason that all the test data units is now EJ rather than EJ/yr because Pint doesn't have yr or EJ/yr defined?

`pint` knows `yr`, but the previous tests used `EJ/y` which pint doesn’t recognise - switching to either `EJ` or `EJ/yr` would work out of the box (you’d have to convert to `MWh` or `MWh/yr` respectively). Since energy is already power x time, I thought dropping the time part made more sense. After all, if you have a value for `year=2020`, it’s clear that it’s energy per year either way. And having `EJ/yr` in sub-annual/continuous time will lead to confusion.

khaeru:
> 2. initialize a global registry on loading `pyam` (wanted to avoid this), then this can be obtained via `pyam.units.REGISTRY` (or similar)?

Pint helpfully provides [`get_application_registry()`](https://pint.readthedocs.io/en/0.11/developers_reference.html#pint.get_application_registry), which I try to use. E.g. at one point ixmp.reporting had a UNITS global variable that got imported everywhere, but then we threw that out and just call `get_application_registry()` where needed.

> If we have a use case where multiple registries are required, we can add this as a kwarg to `convert_unit()` so a user can be clear which registry to use in which case.

IMO this is a super good idea. It requires people to be explicit when they are doing unit transformations—instead of allowing them to just throw in any old string without thinking about what it signifies.

znicholls:
+1 from me on @khaeru’s comments (it looks like pint’s approach is the same as @danielhuppmann’s but using their implementation eliminates that duplication and the use of globals is in another library so should in theory be well tested).

pyam/core.py
```diff
@@ -757,26 +758,30 @@ def rename(self, mapping=None, inplace=False, append=False,
...
-        conversion_mapping: dict
-            for each unit for which a conversion should be carried out,
-            provide current unit and target unit and conversion factor
-            {<current unit>: [<target unit>, <conversion factor>]}
+        current: str (or mapping, deprecated)
+            name of current unit (to be converted from)
+        to: str
+            name of new unit (to be converted to)
+        factor: value, optional
+            conversion factor to used, defaults to pint.UnitRegistry
```
Rlamboll: "to use" is grammatically correct, possibly 'to multiply by' is clearer. This section is a little confusing because of the use of 'to' as a key word!

pyam/core.py
```diff
@@ -757,26 +758,30 @@ def rename(self, mapping=None, inplace=False, append=False,
         if not inplace:
             return ret
 
-    def convert_unit(self, conversion_mapping, inplace=False):
-        """Converts units based on provided unit conversion factors
+    def convert_unit(self, current, to=None, factor=None, inplace=False):
+        """Converts a unit using given factor or pint.UnitRegistry
```
Rlamboll: It might be good to put a line about when we can expect pint to produce the correct answers

Rlamboll:
Generally looks good to me, can we have at least one test that does use "/yr" in it though? Since that's the form most of our data takes, it's good to check that pint won't mess things up by default.

pyam/core.py
danielhuppmann: not quite sure what you mean here - would a link to the pint docs page be sufficient?

Rlamboll: As in, some description of when you will need to hard code a conversion factor when the conversion is context-dependent (Mt CO2-equiv in various frames) and when you won't (kt -> Gt etc).

pyam/core.py
danielhuppmann: changed

pyam/core.py
danielhuppmann: added two paragraph using `EJ` & `GWh` as examples, also mentioning combined units `EJ/yr`, plus a link to the pint docs and the energy-units repo.

`Mt CO2-equiv` is a too specific use case for the general docs, in my view... maybe something for another PR with some context- and custom-UnitRegistry-handling...

danielhuppmann:
> Generally looks good to me, can we have at least one test that does use "/yr" in it though? 

I extended the test `test_convert_unit_with_pint` to cover both `EJ` and `EJ/yr`

danielhuppmann:
requested changes implemented:

 - extended the docstring of the function to show what to expect from `pint` (requested by @Rlamboll)
 - extend the convert-with-pint unit tests to cover combined units (ÈJ/yr`, requested by @Rlamboll)
 - switch to `pint.get_default_registry()` rather than global variable

doc/source/tutorials/aggregating_downscaling_consistency.ipynb
```diff
@@ -366,15 +366,15 @@
...
-    "    ['World', 'Primary Energy', 'EJ/y', 7, 15],\n",
-    "    ['World', 'Primary Energy|Coal', 'EJ/y', 4, 11],\n",
-    "    ['World', 'Primary Energy|Wind', 'EJ/y', 2, 4],\n",
-    "    ['region_a', 'Primary Energy', 'EJ/y', 4, 8],\n",
-    "    ['region_a', 'Primary Energy|Coal', 'EJ/y', 2, 6],\n",
-    "    ['region_a', 'Primary Energy|Wind', 'EJ/y', 2, 2],\n",
-    "    ['region_b', 'Primary Energy', 'EJ/y', 3, 6],\n",
-    "    ['region_b', 'Primary Energy|Coal', 'EJ/y', 2, 4],\n",
-    "    ['region_b', 'Primary Energy|Wind', 'EJ/y', 0, 2],\n",
+    "    ['World', 'Primary Energy', 'EJ', 7, 15],\n",
```
gidden: If I follow the conversation correctly, can this be changed back to `EJ/yr`?

pyam/core.py
```diff
@@ -757,26 +758,42 @@ def rename(self, mapping=None, inplace=False, append=False,
...
-            where = ret.data['unit'] == current_unit
-            ret.data.loc[where, 'value'] *= factor
-            ret.data.loc[where, 'unit'] = new_unit
-        if not inplace:
-            return ret
+        # TODO: deprecate using `dict` in next release (>=0.6.0)
+        # TODO: make `to` required
+        if isinstance(current, dict) and to is None and factor is None:
+            deprecation_warning('Use explicit keyword arguments instead!',
+                                'Using a dictionary to convert units')
```
gidden: General comment - could you explicitly use the kwarg name (helps for future and current eyes to know what is being called), so, e.g., `deprecation_warning('..', type='..')`?

pyam/units.py
```diff
@@ -0,0 +1,41 @@
...
+
+# get application pint.UnitRegistry and load energy-units
+_REGISTRY = pint.get_application_registry()
+file = Path(__file__).parents[1] / 'energy-units' / 'definitions.txt'
+_REGISTRY.load_definitions(str(file))
+
+
+def convert_unit(df, current, to, factor=None, inplace=False):
+    """Internal implementation of unit conversion with explicit kwargs"""
+    ret = copy.deepcopy(df) if not inplace else df
```
gidden: do we need `deepcopy` here, or can we use `df.copy()`?

tests/conftest.py
```diff
@@ -17,34 +17,34 @@
 
 
 TEST_DF = pd.DataFrame([
-    ['model_a', 'scen_a', 'World', 'Primary Energy', 'EJ/y', 1, 6.],
-    ['model_a', 'scen_a', 'World', 'Primary Energy|Coal', 'EJ/y', 0.5, 3],
-    ['model_a', 'scen_b', 'World', 'Primary Energy', 'EJ/y', 2, 7],
+    ['model_a', 'scen_a', 'World', 'Primary Energy', 'EJ', 1, 6.],
```
gidden: same comment on units as above (`EJ/yr`?)

gidden:
Just a few minor comments, mostly on understanding. 

On the git submodule, I think this is fine in the near term. Once the units becomes more stable, I would suggest we simply store a copy in `pyam` unless it is installable in the future, then we can depend on it explicitly.

Overall looks great, thanks a bunch @danielhuppmann !

pyam/units.py
danielhuppmann: `df.copy()` is just a shim for `copy.deepcopy(df)` but I refactored all occurrences for nicer code...

doc/source/tutorials/aggregating_downscaling_consistency.ipynb
danielhuppmann: done throughout all documents, and inverted the pint-test (to test both `EJ` and `EJ/yr`)

tests/conftest.py
danielhuppmann: changed

danielhuppmann:
implemented additional changes suggested by @gidden, merging as approved by @gidden & @znicholls

PR #343, opened 2020-03-06
add test for require_variable()

jkikstra:
This PR:
- tests the behaviour of using a list of years as an argument for `require_variable()`
- clarifies the docstring about the expected behaviour

- [x] Tests Added
- [x] Documentation Added

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -608,8 +608,8 @@ class IamDataFrame(object):
         unit: str, default None
             name of unit (optional)
         year: int or list, default None
-            years (optional)
-            check whether any of the provided years is in the scenario data 
+            years (optional), check whether any of the provided years is in the
+            scenario data 
         exclude_on_fail: bool, default False
             flag scenarios missing the required variables as `exclude: True`
         """
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -609,6 +609,7 @@ class IamDataFrame(object):
             name of unit (optional)
         year: int or list, default None
             years (optional)
+            check whether any of the provided years is in the scenario data 
         exclude_on_fail: bool, default False
             flag scenarios missing the required variables as `exclude: True`
         """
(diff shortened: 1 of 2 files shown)


pyam/core.py
```diff
@@ -608,7 +608,8 @@ def require_variable(self, variable, unit=None, year=None,
         unit: str, default None
             name of unit (optional)
         year: int or list, default None
-            years (optional)
+            years (optional), check whether any of the provided years is in the
+            scenario data
```
danielhuppmann:
```suggestion
            check whether the variable exists for ANY of the years (if a list)
```

tests/test_core.py
```diff
@@ -547,6 +547,29 @@ def test_require_variable_top_level(test_df):
     assert list(test_df['exclude']) == [False, True]
 
 
+def test_require_variable_year_list(test_df):
+    test_df.data = test_df.data[1:]
+    df = test_df
```
danielhuppmann:
```suggestion
    df = IamDataFrame(test_df.data[1:])
```

tests/test_core.py
```diff
@@ -547,6 +547,29 @@ def test_require_variable_top_level(test_df):
...
+    df.require_variable(variable='Primary Energy',
+                        year=[2005, 2010],
```
danielhuppmann: you could have a line `years = [2005, 2010]` at the start of that function and then use this here and below, to make it more obvious to the next person that it's two different ways of testing for the same years.

And maybe add a one-line docstring like "checking for variable in any of the years" before that section and "checking for variable in all of the years" before the next one

tests/test_core.py
```diff
@@ -547,6 +547,29 @@ def test_require_variable_top_level(test_df):
...
+                        exclude_on_fail=True)
+    df.filter(exclude=False, inplace=True)
+
+    assert len(df.variables()) == 2
+    assert len(df.scenarios()) == 2
+
+    test_df.data = test_df.data[1:]
+    df = test_df
```
danielhuppmann:
```suggestion
    df = IamDataFrame(test_df.data[1:])
```

Issue #284, opened 2019-11-05
add unit tests to conda-forge recipe


danielhuppmann:
The recipe for building the release on `conda-forge` should include installing and running the unit tests.

Issue where this was first raised: https://github.com/conda-forge/pyam-feedstock/pull/11

PR #350, opened 2020-03-23
add tests and tutorials to `data_package`

danielhuppmann:
# Description of PR

This PR adds the tests and tutorials folders (and subfolders) to `package_data` in `setup.py`, so that they are copied as part of the installation. This is a step towards enabling #284.


gidden:
Hi @danielhuppmann - going through PRs here. My preference for installing tests would be to follow other standard packages - best example I have is [xarray](https://github.com/pydata/xarray/blob/master/xarray/tests/__init__.py).

PR #359, opened 2020-04-06
Rename filter payload attribute to timeslices

zikolach:
- add logging to filter elements counts

- [ ] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -330,7 +330,8 @@ class Connection(object):
         # pass empty list to API if all regions selected
         if len(regions) == len(self.regions()):
             regions = []
-
+        logger.debug(f"Prepared filter for {len(regions)} region(s), "
+                     f"{len(variables)} variables and {len(runs)} runs")
         data = {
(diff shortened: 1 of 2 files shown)


pyam/iiasa.py
```diff
@@ -330,15 +330,16 @@ def _match(data, patterns):
...
+                     f"{len(variables)} variables and {len(runs)} runs")
         data = {
             "filters": {
                 "regions": list(regions),
                 "variables": list(variables),
                 "runs": list(runs),
                 "years": [],
                 "units": [],
-                "times": []
+                "timeslices": []
```
danielhuppmann, line 342: wouldn't it make more sense for this to be called "subannual"?

zikolach: I don't like `subannual` to be name of the filter which contains list of time slices. Same with DB table name. If you think it won't work for you we can change it later on (it will require redeployment of all Scenario Explorer instances).

PR #370, opened 2020-04-20
Make filter work for np 64 format years/dates

Rlamboll:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR
Allowed filter to work with 
np.int64 years and np.datetime64 dates.

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -354,9 +354,9 @@ def _escape_regexp(s):
 def years_match(data, years):
     """Return rows where data matches year"""
     years = [years] if (
-        isinstance(years, int) or isinstance(years, np.int64)
+        isinstance(years, (int, np.int64))
     ) else years
-    dt = datetime.datetime
+    dt = (datetime.datetime, np.datetime64)
     if isinstance(years, dt) or isinstance(years[0], dt):
         error_msg = "`year` can only be filtered with ints or lists of ints"
         raise TypeError(error_msg)
@@ -426,7 +426,7 @@ def datetime_match(data, dts):
     """Matching of datetimes in time columns for data filtering"""
     dts = dts if islistable(dts) else [dts]
     if any([not (
-            isinstance(i, datetime.datetime) or isinstance(i, np.datetime64)
+            isinstance(i, (datetime.datetime, np.datetime64))
     ) for i in dts]):
         error_msg = (
             "`time` can only be filtered by datetimes and datetime64s"
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -354,7 +354,7 @@ def _escape_regexp(s):
 def years_match(data, years):
     """Return rows where data matches year"""
     years = [years] if (
-            isinstance(years, int) or isinstance(years, np.int64)
+        isinstance(years, int) or isinstance(years, np.int64)
     ) else years
     dt = datetime.datetime
     if isinstance(years, dt) or isinstance(years[0], dt):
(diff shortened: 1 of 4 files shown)


pyam/utils.py
```diff
@@ -353,7 +353,9 @@ def _escape_regexp(s):
 
 def years_match(data, years):
     """Return rows where data matches year"""
-    years = [years] if isinstance(years, int) else years
+    years = [years] if (
+        isinstance(years, int) or isinstance(years, np.int64)
```
gidden, line 357: This can be `isinstance(years, [int, np.int64])`  though I think `isinstance(years, np.integer)` [should work](https://stackoverflow.com/questions/37726830/how-to-determine-if-a-number-is-any-type-of-int-core-or-numpy-signed-or-not)

pyam/utils.py
```diff
@@ -423,9 +425,11 @@ def conv_strs(strs_to_convert, conv_codes, name):
 def datetime_match(data, dts):
     """Matching of datetimes in time columns for data filtering"""
     dts = dts if islistable(dts) else [dts]
-    if any([not isinstance(i, datetime.datetime) for i in dts]):
+    if any([not (
+            isinstance(i, datetime.datetime) or isinstance(i, np.datetime64)
```
gidden, line 429: similar comment here on grouping together, but maybe [numpy has a check for both anyway](https://stackoverflow.com/questions/23063362/consistent-way-to-check-if-an-np-array-is-datetime-like/39020498)?

tests/test_core.py
```diff
@@ -317,6 +317,18 @@ def test_filter_day(test_df, test_day):
         assert unique_time[0] == expected
 
 
+def test_filter_numpy_vals(test_df):
```
gidden, line 320: Can this have a more verbose/descriptive name?

tests/test_core.py
```diff
@@ -317,6 +317,18 @@ def test_filter_day(test_df, test_day):
...
+    dates = test_df[test_df.time_col].unique()
+    if test_df.time_col == "year":
```
gidden, line 322: I think this pattern would be a little cleaner as:

```
key = 'year' if test_df.time_col == "year" else 'time'
res_0 = test_df.filter(**{key: dates[0]})
res = test_df.filter(**{key: dates})
```

pyam/utils.py
Rlamboll: First suggestion works, second doesn't seem to (may depend on numpy version)

pyam/utils.py
Rlamboll: I don't think so, datetime is not clearly related to datetime64

tests/test_core.py
Rlamboll: ok

tests/test_core.py
Rlamboll: That is neater, thanks

pyam/utils.py
```diff
@@ -353,8 +353,10 @@ def _escape_regexp(s):
 
 def years_match(data, years):
     """Return rows where data matches year"""
-    years = [years] if isinstance(years, int) else years
-    dt = datetime.datetime
+    years = [years] if (
+        isinstance(years, (int, np.int64))
+    ) else years
+    dt = (datetime.datetime, np.datetime64)
```
gidden, line 359: looks extraneous?

gidden:
One final comment, then g2g from my side!

gidden:
Thanks a bunch @Rlamboll !

PR #373, opened 2020-05-10
Show head of duplicate rows in error message

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~
# Description of PR

Initializing an IamDataFrame with duplicate rows already raised an error. This PR extends the error message to include the head of the duplicate rows.

This came up as part of @zikolach and @vruijven working on the ixmp-server-integration pipeline and will hopefully simplify identifying the source of problematic data uploads.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 # Next Release
 
 ## Individual Updates
+- [#373](https://github.com/IAMconsortium/pyam/pull/373) Extends the error message when initializing with duplicate rows.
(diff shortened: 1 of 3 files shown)


danielhuppmann:
re @zikolach 

> Do you know why Travis build failed?

I think that this is due to a timeout for the IIASA server - the pyam tests are always a bit fickle since we have tests against the rest-api. Usually restarting the build on travis is enough. But that is one more reason why I wanted to run the tests against an integration-test instance rather than the iamc15 instance, hoping this would give faster response time.

gidden:
Thanks @danielhuppmann !

PR #386, opened 2020-05-28
Removed equiv before performing unit conversion

Rlamboll:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

### DIFF
--- a/pyam/units.py
+++ b/pyam/units.py
@@ -26,21 +26,24 @@ def convert_unit(df, current, to, factor=None, registry=None, context=None,
     # Convert using a pint.UnitRegistry; default the one from iam_units
     registry = registry or iam_units.registry
 
+    # Make versions without -equiv
+    clean_current = _remove_equivs(current)
+    clean_to = _remove_equivs(to)
     # Pair of (magnitude, unit)
-    qty = [ret.data.loc[where, 'value'].values, current]
+    qty = [ret.data.loc[where, 'value'].values, clean_current]
 
     try:
         # Create a vector pint.Quantity
         qty = registry.Quantity(*qty)
     except pint.UndefinedUnitError:
         # *qty* might include a GHG species; try GWP conversion
-        result, to = convert_gwp(context, qty, to)
+        result, to = convert_gwp(context, qty, clean_to)
     except AttributeError:
         # .Quantity() did not exist
         raise TypeError(f'{registry} must be `pint.UnitRegistry`') from None
     else:
         # Ordinary conversion, using an empty Context if none was provided
-        result = qty.to(to, context or pint.Context())
+        result = qty.to(clean_to, context or pint.Context())
 
     # Copy values from the result Quantity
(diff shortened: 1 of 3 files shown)


Rlamboll:
Seems to have failed a test for time reasons, I don't think I can restart it though

danielhuppmann:
Yes, there is some version-conflict in the py3.6 dependencies - guess we'll need to remove that test at some point...

pyam/units.py
```diff
@@ -26,21 +26,24 @@ def convert_unit(df, current, to, factor=None, registry=None, context=None,
     # Convert using a pint.UnitRegistry; default the one from iam_units
     registry = registry or iam_units.registry
 
+    # Make versions without -equiv
```
danielhuppmann, line 29: I think that having an own function for this simple removal is a bit excessive, you could do that in one line with 
```
current, _to = [i.replace('-equiv', 'e') for i in [current, to]]
```
(note the `_to` as an alternative to kee p `to` for the renaming)

tests/test_units.py
```diff
@@ -46,6 +47,22 @@ def test_convert_unit_with_pint(test_df, current, to):
     assert_converted_units(df, current, to, exp)
 
 
+def test_convert_eqiv_units(test_df):
```
danielhuppmann, line 50: Should be enough to just add a line to the parametrisation of `test_convert_gwp()`?

Rlamboll: Can do, but I don't really trust this as about a quarter of these tests are broken on my machine, I don't understand why.

pyam/units.py
Rlamboll: By leaving out the 'e' this works for all units, even ones without the alias set up.

tests/test_units.py
danielhuppmann: did you install the latest version of `iam_units` from GitHub?

Rlamboll: nope, turns out it had moved on

danielhuppmann:
thanks @Rlamboll, looks great - can you still pull my suggested changes in https://github.com/Rlamboll/pyam/pull/2 (or do you disagree?), then good to go!

Rlamboll:
Honestly I find those variable names very confusing because unless you read that specific line of code, you'd have no idea what those variables were and I would likely assume current was still the input value if I wanted to use it later.

danielhuppmann:
> Honestly I find those variable names very confusing because unless you read that specific line of code, you'd have no idea what those variables were and I would likely assume current was still the input value if I wanted to use it later.

I pushed another commit to the PR-branch so that the "cleaned" (without `-equiv`) unit-args are both saved as new internal variables. But I think that cleaning `current` to `_current` is much more python than `clean_current`...

Still, if you can't be convinced, please cherry-pick  the other commits (removed unused utility function and add extra test).

Rlamboll:
Well, you're the one doing the upkeep so if you want those names you can have them.

PR #393, opened 2020-06-09
Fix importing ABC from collections in Python 3.

### LINKED ISSUE #377: Deprecation warnings for python 3.9
Just noticed these two warnings about upcoming deprecations concerning the run-control and plotting packages.

> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

tirkarthi:

- [ ] Tests Added
- [ ] Documentation Added

```
```

# Description of PR

Fixes #377 . It also fixes several more instances in addition to the one reported in #377


danielhuppmann:
thanks @tirkarthi for this fix! I actually started on something similar, see https://github.com/danielhuppmann/pyam/commit/c4c8f19d53496496c871d4576353f41ba8629ae7 - is there any specific reason for preferring
`from six.moves.collections_abc import Iterable`
over
`from collections.abc import Iterable`?

tirkarthi:
ABC can be imported from collections only in Python. In Python 3 collections.abc is recommended. Using six would ensure it works on both python 2 and 3. I assumed Python 2 is supported by the project. If not then directly using collections.abc is a simple solution. Does the project support Python 2?

danielhuppmann:
Thanks! I believe that all major packages are dropping support for python 2, and we started using `f'{}'`-strings for formatting, which are incompatible with versions before 3.6 anyway (afaik)...

tirkarthi:
Thanks, I will modify the PR

tirkarthi:
@danielhuppmann I have force pushed my changes to ensure collections.abc is used instead of six.

danielhuppmann:
Thanks @tirkarthi! Can you still add a line to the release notes? Note that this is the first PR since the last release, so you'll need to insert it between `# Next Release` and `# Release v0.6.0`.

PR #396, opened 2020-06-14
Enable casting to IamDataFrame multiple times

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

I recently noticed that casting an IamDataFrame again (i.e., `IamDataFrame(IamDataFrame(<file>))` raises a non-intuitive error message.

This PR implements a simple approach to circumvent this problem, with all attributes of the first and second casting of the IamDataFrame instance pointing to the same objects. This follows the logic for pandas.DataFrames (as far as I understand it) where performing an operation on one instance has an impact also on the second (see the new test to show how this works in pyam).

An alternative approach may be to do a full copy during the double-casting.

## Use case

I wrote a script recently that cast an object to an `IamDataFrame` ([see here](https://github.com/openENTRANCE/nomenclature/blob/69703d073a6367332aab1d9d648dc01a9ec483c6/nomenclature/__init__.py#L82)) before performing a few checks and operations - this failed when the object was already an `IamDataFrame`. That function has been extended to first check before casting, but this may be a problem that other users run into.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -90,6 +90,11 @@ class IamDataFrame(object):
     kwargs :code:`sheet_name` ('data') and :code:`meta_sheet_name` ('meta')
     Calling the class with :code:`meta_sheet_name=False` will
     skip the import of the 'meta' table.
+
+    When initializing an :class:`IamDataFrame` from an object that is already
+    an :class:`IamDataFrame` instance, the new object will be hard-linked to
+    all attributes of the original object - so any changes on one object
+    (e.g., with :code:`inplace=True`) may also modify the other object!
     """
     def __init__(self, data, **kwargs):
         """Initialize an instance of an IamDataFrame"""
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -93,6 +93,14 @@ class IamDataFrame(object):
     """
     def __init__(self, data, **kwargs):
         """Initialize an instance of an IamDataFrame"""
+        if isinstance(data, IamDataFrame) and not kwargs:
+            for attr, value in data.__dict__.items():
+                setattr(self, attr, value)
+        else:
+            self._init(data, **kwargs)
+
+    def _init(self, data, **kwargs):
+        """Process data and set attributes for new instance"""
         # import data from pd.DataFrame or read from source
         if isinstance(data, pd.DataFrame) or isinstance(data, pd.Series):
             _data = format_data(data.copy(), **kwargs)
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -100,7 +100,10 @@ class IamDataFrame(object):
     """
     def __init__(self, data, **kwargs):
         """Initialize an instance of an IamDataFrame"""
-        if isinstance(data, IamDataFrame) and not kwargs:
+        if isinstance(data, IamDataFrame):
+            if kwargs:
+                msg = 'Invalid arguments `{}` for initializing an IamDataFrame'
+                raise ValueError(msg.format(kwargs))
             for attr, value in data.__dict__.items():
                 setattr(self, attr, value)
         else:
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -46,6 +46,19 @@ def test_init_df_with_index(test_pd_df):
     pd.testing.assert_frame_equal(df.timeseries().reset_index(), test_pd_df)
 
 
+def test_init_from_iamdf(test_df_year):
```
znicholls, line 49: Do you have a test for what happens if there are kwargs?

Rlamboll:
I think this does what it initially set out to do, and am happy to approve it. However the one use-case that I would understand for this is 'refreshing' databases. I don't know if you're aware (@znicholls has described this as a bug before but I assumed it was intentional) but changing the data doesn't update the values extracted by calls like "df.scenarios()", so I often call df = DataFrame(df.data) to ensure everything is consistent after modifying data directly. To me it would be most intuitive if these had the same effect regarding consistency.

danielhuppmann:
> However the one use-case that I would understand for this is 'refreshing' databases. I don't know if you're aware (@znicholls has described this as a bug before but I assumed it was intentional) but changing the data doesn't update the values extracted by calls like "df.scenarios()", so I often call df = DataFrame(df.data) to ensure everything is consistent after modifying data directly. To me it would be most intuitive if these had the same effect regarding consistency.

I'm not quite sure what you mean with "refreshing" databases - do you mean an `IamDataFrame`? Also, "everything is consistent" - this should be the case anyway if you use the proper functions like `rename()`, `convert_unit()`, `append()`.

`df.scenarios()` returns a list of scenarios at the time when you call the function - it is not intended to be updated when `df` is changed...

Maybe bring this into a new issue if it should be discussed more thoroughly?

About the use case, I added a section to the PR description.

Rlamboll:
> - this should be the case anyway if you use the proper functions like rename(), convert_unit(), append().
I don't, I directly operate on the .data and then make it consistent afterwards. I don't think it's necessarily a problem, just something to be aware of. I will start a "bug report" for clarity anyway.

tests/test_core.py
danielhuppmann: added a test!

danielhuppmann:
@znicholls added a test and an additional paragraph in the notes on the to-be-expected behavior. Please say if that is sufficiently clear!

tests/test_core.py
```diff
@@ -46,6 +46,25 @@ def test_init_df_with_index(test_pd_df):
...
+    assert all(test_df_year.scenarios().values == ['scen_b', 'scen_foo'])
+
+    # overwrites on the new object do not have effects on the original object
+    df = df.rename(scenario={'scen_foo': 'scen_bar'})
+    assert all(df.scenarios().values == ['scen_b', 'scen_bar'])
+    assert all(test_df_year.scenarios().values == ['scen_b', 'scen_foo'])
+
+
+def test_init_from_iamdf_raises(test_df_year):
+    # casting an IamDataFrame instance again with extra args fails
```
znicholls, line 64: should there be any error message in particular here? I can imagine that getting a ValueError just cause you added kwargs would be confusing? Maybe it should say, "If you init with an IamDataFrame, you cannot use kwargs"?

pyam/core.py
```diff
@@ -90,9 +90,22 @@ class IamDataFrame(object):
     kwargs :code:`sheet_name` ('data') and :code:`meta_sheet_name` ('meta')
     Calling the class with :code:`meta_sheet_name=False` will
     skip the import of the 'meta' table.
+
+    When initializing an :class:`IamDataFrame` from an object that is already
+    an :class:`IamDataFrame` instance, the new object will be hard-linked to
+    all attributes of the original object - so any changes on one object
+    (e.g., with :code:`inplace=True`) may also modify the other object!
```
znicholls, line 97: Maybe add that 'This is intended behaviour and consistent with Pandas but may be confusing for those who are not used to the Pandas/Python universe.'

tests/test_core.py
danielhuppmann: added an explicit error message (leaving the door open for valid keyword arguments in the future)

pyam/core.py
danielhuppmann: done

danielhuppmann:
@gidden, do you mind taking a look?

PR #401, opened 2020-06-20
Read iiasa-creds by default from known location

danielhuppmann:

- ~Tests Added~ (requires more work on GitHub Actions to store secrets, will be added as part of the IIASA-API-tests refactoring
- [x] Documentation Added
# Description of PR

This PR adds a utility to save the iiasa-api credentials in a known location and provides a utility to store them (`pyam.iiasa.set_config(<user>, <password>)`). This will make it easier for users to not write their password in workflow scripts. Providing credentials as kwargs to `Connection` is marked as deprecated and will be removed in the future.

Documentation and tutorials are updated and built [here](https://pyam-iamc.readthedocs.io/en/iiasa-creds-config/) on readthedocs.

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -98,29 +98,28 @@ def _get_token(creds, base_url):
 
 
 class Connection(object):
-    """A class to facilitate querying an IIASA scenario explorer database"""
+    """A class to facilitate querying an IIASA scenario explorer database
 
+    Parameters
+    ----------
+    name : str, optional
+        A valid database name. For available options, see
+        valid_connections().
+    creds : str, :class:`pathlib.Path`, list-like, or dict, optional
+        By default, this function will (try to) read user credentials which
+        were set using :meth:`pyam.iiasa.set_config(<user>, <password>)`.
+        Alternatively, you can provide a path to a yaml file
+        with entries of  'username' and 'password'.
+    base_url : str, custom authentication server URL
+
+    Notes
+    -----
+    Providing credentials as an ordered container (tuple, list, etc.)
+    or as a dictionary with keys `user` and `password` is (still) supported
+    for backwards compatibility. However, this option is NOT RECOMMENDED
+    and will be deprecated in future releases of pyam.
+    """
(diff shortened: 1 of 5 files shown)


pyam/iiasa.py
```diff
@@ -23,74 +25,114 @@
...
-    if creds is None:  # get anonymous auth
+    """Parse credentials and get token from IIASA authentication service"""
+    plaintextcreds = True
+
+    # try reading default config or parse file
+    if creds is None:
+        creds = _get_config()
+        plaintextcreds = False
+    elif isinstance(creds, Path) or isstr(creds):
+        _creds = _get_config(creds)
```
zikolach, line 65: nit: maybe rename `creds` to `creds_or_config_path`

danielhuppmann:
> One thing I would maybe considered refactoring is update _get_token to avoid returning tuple.

I'm not sure what you mean here?

pyam/iiasa.py
danielhuppmann: I don't think that passing a custom path to a creds file will be used much (other than by us for developing), because you only have one user-password anyway... so I'd prefer to keep the API concise.

zikolach:
> > One thing I would maybe considered refactoring is update _get_token to avoid returning tuple.
> 
> I'm not sure what you mean here?

I meant that `_get_token` is doing 2 things: handling credentials and getting token, if you split it into 2 functions than maybe result of `_get_token` will be just token (as originally) instead of tuple containing token and user name.

danielhuppmann:
> I meant that _get_token is doing 2 things: handling credentials and getting token, if you split it into 2 functions than maybe result of _get_token will be just token (as originally) instead of tuple containing token and user name. But this is not critical as _get_token is just private/internal function.

I went a few steps into the rabbit hole of separating these two functions - but I didn't find a way to do it elegantly, so let's leave a refactoring until a more comprehensive update of the API...

danielhuppmann:
ignoring codecov/project, this should have been set to "informational"...

PR #403, opened 2020-06-20
test first version of plotting.boxplot

byersiiasa:

- [x] Tests Added
- [x] Documentation Added (to plotting examples)
# Description of PR

This PR adds possibility of plotting directly to a boxplot, currently implemented using seaborn. Seaborn implementation is preferred over standard matplotlib/pandas boxplot options due to the ease of adding 3rd dimension "hue" - which allows grouping of boxplots.


byersiiasa:
@danielhuppmann  can you please check the test failure?
Trying to check the `type` of df in the function - 
`    if type(df) == pyam.core.IamDataFrame:`
but the tests don't like that.

pyam/core.py
```diff
@@ -1531,7 +1531,16 @@ def bar_plot(self, *args, **kwargs):
         df = self.as_pandas(with_metadata=True)
         ax = plotting.bar_plot(df, *args, **kwargs)
         return ax
+        
+    def boxplot(self, *args, **kwargs):
```
gidden, line 1535: can this be `box_plot` to match other functions please?

pyam/plotting.py
```diff
@@ -496,6 +497,67 @@ def bar_plot(df, x='year', y='value', bars='variable',
     return ax
 
 
+def boxplot(df, y='value', x='year', hue=None, palette='Spectral_r',
```
gidden, line 500: same request here - `box_plot()`

pyam/plotting.py
```diff
@@ -496,6 +497,67 @@ def bar_plot(df, x='year', y='value', bars='variable',
...
+            ax=None, legend=False, **kwargs):
+    """ Plot boxplot of data using seaborn.boxplot
+
+    Parameters
+    ----------
+    df : pandas.DataFrame or pyam.IamDataFrame
```
gidden, line 506: to match other functions, this should only take a `pd.DataFrame` 

the thought here is that these utilities are purely for plotting, and internal manipulations to the `pyam.DataFrame` should happen in that scope (`core.py` or other) and not here

pyam/plotting.py
```diff
@@ -496,6 +497,67 @@ def bar_plot(df, x='year', y='value', bars='variable',
...
+        default: 'Spectral_r'
+    ax : matplotlib.Axes, optional
+    legend : bool, optional
+        Include a legend
+        default: False
+    kwargs : Additional arguments to pass to the pd.DataFrame.plot()
+    """
+
+    # Check data_type
+    if type(df) == pyam.core.IamDataFrame:
```
gidden, line 533: all data preparation should go in the `box_plot()` function in `core.py`. see, e.g., https://github.com/IAMconsortium/pyam/blob/90bcef21019217d7030a5d94d8d142ad3fbdc6af/pyam/core.py#L1497

gidden:
cool thanks @byersiiasa ! it looks like this only works for variables at the moment, do you want to extend to other data (e.g., metadata, model, scenario)?

pyam/core.py
danielhuppmann: I disagree, I would follow the pandas convention where it's also called `boxplot`

pyam/plotting.py
danielhuppmann: as above, I'd rather keep it as is...

byersiiasa:
> cool thanks @byersiiasa ! it looks like this *only works for variables* at the moment, do you want to extend to other data (e.g., metadata, model, scenario)?

@gidden   I think `y` probably only  `value` or  a numerical column from meta_data; but `x` and `hue` should be fine taking `model`, `scenario`, region` , `year` or anything else.

Or do you mean something else?

pyam/plotting.py
danielhuppmann: agree with @gidden - let's not overload the functionality within the plotting module, this should only receive correctly formatted data from `IamDataFrame.boxplot()` or a similarly formatted pandas.DataFrame

pyam/plotting.py
```diff
@@ -496,6 +497,62 @@ def bar_plot(df, x='year', y='value', bars='variable',
...
+        Include a legend
+        default: False
+    kwargs : Additional arguments to pass to the pd.DataFrame.plot() 
+    
+    
+    """
+    
+    # Check data_type
+    if type(df) == pyam.core.IamDataFrame:
+        if hue==None:
```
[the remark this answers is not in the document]
danielhuppmann: this could be just `if not hue`

RELEASE_NOTES.md
```diff
@@ -1,5 +1,5 @@
 # Next Release
-
+- [#395] (https://github.com/IAMconsortium/pyam/pull/403_ - Add boxplot function
```
danielhuppmann, line 2:
```suggestion

- [#403](https://github.com/IAMconsortium/pyam/pull/403) Add boxplot function.
```

danielhuppmann:
> @danielhuppmann can you please check the test failure?
> Trying to check the `type` of df in the function -
> ` if type(df) == pyam.core.IamDataFrame:`
> but the tests don't like that.

Just tried running this is a notebook, it also doesn't like it. In general, you want `isinstance(df, IamDataFrame)`, but in this case (as noted by @gidden) the data-processing should be moved out of this function anyway...

danielhuppmann:
Thanks @byersiiasa for taking a first stab at a boxplot function!

I'm a bit worried that having plots created by different libraries will make maintenance more difficult going forward... But I also see (after some playing around now) that seaborn is easier than matplotlib for this.

Still, two recommendations:
 - follow the pandas style and use `by` rather than `hue` in the public-facing API - after all, the aim is to group the data *by*. some column(s)...
 - it would be nice to being able to use the colors set in runcontrol out-of-the-box

byersiiasa:
thanks @gidden @danielhuppmann 
Ok. Implemented almost all suggested changes, e.g.
`by` instead of `hue`
made more consistent with other plotting  functions
have left as `boxplot` for now - you will have to duel it, but my thinking was same as Daniels.

Comment about `run_control` is acknowledged, may try another time. Have removed the `palette` argument for now - user can specify as they want.

Tested the `plot_boxplot` example script - but the data is not hugely suited for boxplot. ideally would have more models or scenarios. This could be added to `data.csv`, with caveat that `filter` calls in every plotting example may need to be checked / adjusted
or could add a separate dataset e.g. `data_bp.csv`

danielhuppmann:
> Tested the plot_boxplot example script - but the data is not hugely suited for boxplot. ideally would have more models or scenarios.

I would be in favour of a better dataset for the plot examples (and tests)...

byersiiasa:
ok - so hopefully this has nice enough example plot for @danielhuppmann to have a look
I added some more scenarios to the `-examples/data.csv` - hope that's ok!

danielhuppmann:
Thanks @byersiiasa - as discussed, I squash-and-rebased your branch to the latest master, and started a new PR.
Closing in favor of #440

PR #409, opened 2020-06-27
Remove travis and appveyor CI config

danielhuppmann:

# Description of PR

This PR:
 - [x] Removes the configuration files for travis and appveyor.
 - [x] Badges on the docs pages have been updated (see [latest on RTD](https://pyam-iamc.readthedocs.io/en/latest)).
 - [x] integration with travis and appveyor has been deactivated.


danielhuppmann:
@gidden, can you check whether there are any other files or settings that can be removed as part of the migration to GitHub Actions and codecov? thanks!

gidden:
@znicholls - this would remove files required for the `make` commands (namely environment files). we were using `make test` etc on our various CIs, but now this is not needed for github actions. 

I think we have three options:
1. remove also the `Makefile` on which this depends
2. keep the environment files
3. use `make` in the github actions

I would prefer 1 or 3, as otherwise things will get out of sync. Other preferences?

znicholls:
Having done this a few times now (see eg github.com/openscm/scmdata or github.com/openscm/openscm-twolayermodel), I’d go for 1. The ultimate source of truth is the CI. The Makefile is a pure convenience but I don’t think you or @danielhuppmann use it so there’s no point keeping it. If I get back into serious development I’ll add it back in but there’s no point having it unless it’s useful.

danielhuppmann:
thanks @gidden & @znicholls for the input!

danielhuppmann:
just realized that I'm actually using `make publish-on-pypi` for every release...

danielhuppmann:
re-added the environment files needed by `make`

znicholls:
https://github.com/danielhuppmann/pyam/pull/16/files fixes up the `Makefile` so you can remove the `ci` directory again

danielhuppmann:
thanks @znicholls for the assist!

PR #411, opened 2020-06-29
Add a feature to pass an explicit weight dataframe to `downscale_region()`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds an option to pass an explicit weight-dataframe with region (and possibly others) as index dimension and the time domain as columns. Implementation based on many discussions with @maartenbrinkering on his specific use case.


danielhuppmann:
@fabiosferra, maybe also of interest to you?

l-welder:
Tested the new features locally and everything seems to be working as intended. Looking forward to add to this (=

danielhuppmann:
Thanks @l-welder for the review and noticing the change in the docs. We actually started out specifying the default in the docstrings, but (if I recall correctly) @khaeru pointed out that the [NumPy-style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) convention is to just write "optional", because the default value is anyway obvious from the function signature. So I'm changing this whenever I stumble over the legacy docstrings...

PR #412, opened 2020-07-02
Add building the docs to GitHub Actions CI and update RTD config

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added
# Description of PR

This PR adds a GitHub Action recipe for testing that the docs build successfully. It also adds the doc-dependency to the pip-install extra-requirements rather than having it as a separate Conda environment, and updates the RTD config to use pip-install. The docs-pages are built [here](https://pyam-iamc.readthedocs.io/en/devops-docs-ci/).


danielhuppmann:
@gidden @znicholls I don't think that a review is strictly necessary here, but if you have a minute to take a look...

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,47 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
znicholls, line 1: Comment is wrong no?

danielhuppmann: oops. thanks for spotting!

PR #416, opened 2020-07-16
Include `meta` in new IamDataFrames returned by aggregation functions

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added
# Description of PR

This PR ensures that all aggregation functions include the `meta` table of the original IamDataFrame when returning a new instance (`append=False`). As part of the implementation, all test dataframes now have one or two meta column, and the merging of meta tables was moved from the `append()` function to an own function in utils to be reused.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -267,8 +267,9 @@ class IamDataFrame(object):
                **kwargs):
         """Append any IamDataFrame-like object to this object
 
-        Columns in `other.meta` that are not in `self.meta` are merged.
-        Conflicting region-variable-unit-year rows raise a `ValueError`.
+        Indicators in `other.meta` that are not in `self.meta` are merged.
+        Missing values are set to `NaN`.
+        Conflicting `data` rows always raise a `ValueError`.
 
         Parameters
         ----------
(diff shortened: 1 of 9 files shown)


pyam/core.py
```diff
@@ -254,8 +267,8 @@ def append(self, other, ignore_meta_conflict=False, inplace=False,
                **kwargs):
         """Append any IamDataFrame-like object to this object
 
-        Columns in `other.meta` that are not in `self.meta` are always merged,
-        duplicate region-variable-unit-year rows raise a ValueError.
+        Columns in `other.meta` that are not in `self.meta` are merged.
```
znicholls, line 270: Is this meant to be columns or rows? i.e. would new columns be added (and missing values filled with nan), or are new rows added for new scenarios etc.?

danielhuppmann: The result is a "complete" meta dataframe (one row for each model-scenario, one column per indicator, missing values set to nan). I rewrote the docstring for clarity.

danielhuppmann:
Thanks for the review @znicholls 

> The one thing I couldn't see was an explicit test for the behaviour you've changed?

The new behavior is tested by adding meta-columns in the test dataframes provided in `conftest.py`, so the tests check that the new IamDataFrame instances returned by aggregation functions also have these meta indicators (via `assert_iamframe_equal()`). You could test the test behavior by only extracting the changes to the `conftest.py` and the test files, but not `core.py` and `utils.py` - that should fail.

Anyway, I added an explicit test for `merge_meta()` to make sure that the precedence of left over right is not accidentally changed in the future.

pyam/core.py
znicholls: beauty thanks

PR #420, opened 2020-08-03
Refactor internal `data` object to pandas.Series

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR is a first step towards solving #335, refactoring the internal `data` object to a `pandas.Series` for better performance.

 - [x] add a getter `data` to return the internal pd.Series as pd.DataFrame object to maintain current behaviour
 - [x] add a setter `data` to taking a `pd.DataFrame` (for backward compatibility, see https://github.com/GranthamImperial/silicone/issues/109 - but to be deprecated in the future)
 - [x] add utility functions for performing renaming directly on the index for performance
 - [x] refactor the unit-conversion feature to the new formulation
 - [x] refactor the `append()` function
 - [x] refactor `timeseries()` and `pivot_table()` functions
 - [x] refactor `interpolate()` implementation
 - [x] refactor `categorize()` and `validate()` implementation

Because some of the tests worked directly on `IamDataFrame.data`, they needed to be refactored too.


danielhuppmann:
FYI @gidden @znicholls @Rlamboll @khaeru, this PR implements one solution discussed in #335: refactoring the `data` timeseries table from a pandas.DataFrame to a pandas.Series. Playing around with some trial workflow, it brings a substantial improvement in performance. Shout-out to @macflo8 for substantial work on refactoring individual functions...

This PR is probably too large to be reviewed in the usual sense, but any comments would nonetheless be appreciated!

khaeru:
Seems like a great advance, glad to hear it worked!

I see there was further discussion/playing around with xarray in the other issue that I missed, having muted it. Will add some comments there with info that's come up recently.

pyam/core.py
```diff
@@ -744,7 +753,9 @@ def rename(self, mapping=None, inplace=False, append=False,
         idx = ret.meta.index.isin(_make_index(ret.data[rows]))
 
         # if `check_duplicates`, do the rename on a copy until after the check
-        _data = ret.data.copy() if check_duplicates else ret.data
+        # _data = ret.data.copy() if check_duplicates else ret.data
+        # TODO reactivate this avoidance of creating a copy
```
znicholls, line 757:
```suggestion
```

Is this solved?

RELEASE_NOTES.md
```diff
@@ -1,5 +1,6 @@
 # Next release
 
+- [#420](https://github.com/IAMconsortium/pyam/pull/420) Refactor `data` object to a pandas.Series
```
znicholls, line 3: Has support for datetime times also been dropped (or were the deleted tests redundant)?

```suggestion
- [#420](https://github.com/IAMconsortium/pyam/pull/420) Add `pyam.IamDataFrame._data` to handle data internally. Make `pyam.IamDataFrame._data` a `pandas.Series` to improve performance (greatly reduces duplication of metadata compared to a long dataframe)
```

znicholls (commented):
Very nice. Some brief thoughts as requested

RELEASE_NOTES.md
danielhuppmann: datetime support has absolutely not been dropped (though having a tutorial for it would be nice to make it easier to advertise this feature...)

The tests were removed per our email conversation on July 23 where you confirmed that they are redundant

danielhuppmann: Re the suggestion, not sure if this is the best place - will add this extended description somewhere else in the release notes

pyam/core.py
danielhuppmann: This is not solved, and not all functions are (intended to be) refactored in this PR - in particular this one and the aggregation functions still use the dataframe implementation (always creating a copy of the pandas.Series rather than the more efficient "only create a copy if necessary").

There was a trade-off on how many functions to refactor in one PR - the plan is that @macflo8 and I continue to work on this throughout September and then have a new release.

Rlamboll:
Speed-up is remarkable, good job! Following the latest push, it mostly replicates previous behaviour. 
However note that this does not fix this issue https://github.com/IAMconsortium/pyam/issues/423

PR #427, opened 2020-09-23
Add an `info()` function and use in `print(IamDataFrame)`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds an `info()` function similar to the pandas equivalent ([read the docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.info.html)) and when calling `print()` on an xarray object, which is the also used in `print(df)` (via `__repr__`).

The output in the first-steps tutorial is:
```
Index dimensions:
 * model    : AIM/CGE 2.1, GENeSYS-MOD 1.0, ... WITCH-GLOBIOM 4.4 (8)
 * scenario : 1.0, CD-LINKS_INDCi, CD-LINKS_NPi, ... Faster Transition Scenario (8)
 * region   : R5ASIA, R5LAM, R5MAF, R5OECD90+EU, R5REF, R5ROWO, World (7)
 * variable : ... (6)
 * unit     : EJ/yr, Mt CO2/yr, °C (3)
 * year     : 2010, 2020, 2030, 2040, 2050, 2060, 2070, 2080, ... 2100 (10)
Meta indicators:
   exclude (bool) False (1)
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,7 @@ via getter and setter functions.
 
 ## Individual updates
 
+- [#427](https://github.com/IAMconsortium/pyam/pull/427) Add an `info()` function and use in `print(IamDataFrame)`
 - [#424](https://github.com/IAMconsortium/pyam/pull/424) Add a tutorial reading results from a GAMS model (via a gdx file).
 - [#420](https://github.com/IAMconsortium/pyam/pull/420) Add a `_data` object (implemented as a pandas.Series) to handle timeseries data internally.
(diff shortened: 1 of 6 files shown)


danielhuppmann:
Rebased and implemented changes suggested by @Rlamboll in https://github.com/danielhuppmann/pyam/pull/21:
 - adding number of unique items per index dimension and meta indicator type
 - smarter use of line width (can be set kwarg so that we can make it nicer in Jupyter notebooks later)
 - make showing memory usage optional

danielhuppmann:
merging as this has no direct implications on any workflow and formatting can be easily adapted later

PR #431, opened 2020-10-01
Bugfix when replacing index values leads to duplicates

### LINKED ISSUE #430: Error when performing unit conversion in new pyam
Some bug in the new production will not allow the conversion of units to a unit already found in the dataframe. It's fine if there isn't such a unit. 

Here's an example:

```
import pyam
import pandas as pd
# construct df
_mc = "model_c"
_sa = "scen_a"
_sb = "scen_b"
_eco2 = "Emissions|CO2"
_gtc = "Gt C/yr"
_ech4 = "Emissions|CH4"
_mtch4 = "Mt CH4/yr"
_msrvu = ["model", "scenario", "region", "variable", "unit"]
simple_df = pd.DataFrame(
    [
        [_mc, 1, "World", _eco2, _gtc, 0, 1000, 5000],
        [_mc, _sb, "World", _eco2, _gtc, 1, 1000, 5000],
        [_mc, _sa, "World", _ech4, _mtch4, 0, 300, 500],
        [_mc, _sb, "World", _ech4, _mtch4, 1, 300, 500],
    ],
    columns=_msrvu + [2010, 2030, 2050],
)
simple_df = pyam.IamDataFrame(simple_df)
# convert units
simple_df.convert_unit(to=_gtc, current=_mtch4, factor=28)
# That fails. However, the following does not since it does not have the pre-existing unit 
simple_df.filter(variable=_ech4).convert_unit(to=_gtc, current=_mtch4, factor=28)
```

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This fixes the bug reported by @Rlamboll in #430. The replacing does not do transitive renaming within one operation, similar to the behaviour of `pandas.Series.replace()`.

closes #430

### DIFF
--- a/pyam/index.py
+++ b/pyam/index.py
@@ -8,9 +8,24 @@ def get_index_levels(df, level):
 
 def replace_index_values(df, level, mapping):
     """Replace one or several category-values at a specific level"""
-    return df.index.set_levels([mapping[i] if i in mapping else i
-                                for i in get_index_levels(df, level)], level)
+    index = df.index.copy()
+    n = index._get_level_number(level)
+    unused_levels = False
 
+    for key, value in mapping.items():
+        _levels = list(index.levels[n])
+        try:  # replace in index codes if value (target) is in the index levels
+            _k = _levels.index(key)
+            _v = _levels.index(value)
+            _codes = index.codes[n]
+            index = index.set_codes([_v if c == _k else c for c in _codes], n)
+            unused_levels = True
+        except ValueError:  # else replace key for value in the levels
+            index = index.set_levels([value if l == key else l
+                                     for l in _levels], n)
+    if unused_levels:
+        index = index.remove_unused_levels()
+    return index
 
 def append_index_level(index, codes, level, name, order=False):
     """Append a level to a pd.MultiIndex"""
--- a/pyam/index.py
+++ b/pyam/index.py
@@ -10,8 +10,8 @@ def replace_index_values(df, level, mapping):
     """Replace one or several category-values at a specific level"""
     index = df.index.copy()
     n = index._get_level_number(level)
-    unused_levels = False
 
+    # replace the levels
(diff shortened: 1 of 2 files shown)


Rlamboll (approved):
I guess this is fine, but I wonder how concerned we are with speed here. If we are trying to speed things up, I'd imagine not assigning values to _k and _v for a single use would be more efficient. I also wonder if performing the switch via an if statement rather than try-except would be faster and safer (particularly if no one has ever had success the first way before).  Also, is df["scenario"] = [list] future-proof? I thought we were discouraging that notation.

danielhuppmann:
> ... not assigning values to _k and _v for a single use would be more efficient. 

Just tried the following comparison.

```
x = list(range(10**5))

[True if i == x.index(9999) else False for i in range(10**5)]

_i = x.index(9999)
[True if i == _i else False for i in range(10**5)]
```
The second one is orders of magnitude faster, because the expression in the list comprehension isn't executed 10**5 times (but only once).

> I also wonder if performing the switch via an if statement rather than try-except would be faster and safer 

Try-except is faster and I don't see why it wouldn't be unsafe.

> is df["scenario"] = [list] future-proof? I thought we were discouraging that notation.

This is operating on the `DataFrame`, not an `IamDataFrame`.

Rlamboll:
OK, I tried it out with implicit _k and _v and it is indeed the same speed or slower, so go ahead and merge.

danielhuppmann:
@Rlamboll, I have been thinking about the transitive renaming that would have been allowed by my first implementation. So I refactored it and added a specific test for it (third parametrization of `test_replace_index_level`).

Can you test again in a real-world example if this now works as expected?

Rlamboll:
Yes, that behaves as I would expect on my test case. It also looks a lot better now and is easier to follow the flow of logic, rather than needing to work out how valueerrors could occur in the code.

danielhuppmann:
thank you for the review!

PR #450, opened 2020-10-29
Set-up deferred logging with basicConfig

### LINKED ISSUE #449: Logging etiquette
The current default logging behaviour by pyam is rude, if you actually want to configure logging.

Usually, one can expect a zen-like experience by just invoking
```python
import logging
logging.basicConfig(level="ERROR")
```

But pyam's initialization in 
https://github.com/IAMconsortium/pyam/blob/f071e790de7fdc3ab6a1135e4b31be7f395cff15/pyam/__init__.py#L14
and 
https://github.com/IAMconsortium/pyam/blob/f071e790de7fdc3ab6a1135e4b31be7f395cff15/pyam/__init__.py#L21-L26
does not respect these settings.

With those, one has to additionally issue
```python
pyam.logger.setLevel(level="ERROR")
```

Annoyingly,
```python
logging.getLogger().setLevel("ERROR")
```
will still show `INFO` messages coming from pyam, since they are already written out into stderr by the `pyam` logger.

That's why the [Configuring logging for a library section](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) in the logging tutorial has this note:

> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers.

I see two alternatives for improvement:

# Best practice
Move the initialization into a separate routine `setup_logger` and tell everyone to use
```python
import pyam
pyam.setup_logging()
```
(`setup_logging` should take keyword arguments like level)

If they forget about it, it's not that problematic either as long as we get in tandem rid of the `NullHandler`, because since python 3.2 the default behaviour of the logging library for a `LogRecord` which bubbles through the logging tree without encountering a Handler is to send the message to stderr, if it is at least a warning (without timestamp and loglevel), as described in [What happens if no configuration is provided](https://docs.python.org/3/howto/logging.html#what-happens-if-no-configuration-is-provided).

Looks like so:
image

# Least change (but still slightly rude)
Get the root logger in `__init__.py` and check whether the user has set up any handlers before, as like:
```python
rootLogger = logging.getLogger()
if not rootLogger.hasHandlers():
    # add handlers to rootLogger
```

That way, the user experience for the average pyam user is unchanged, but one can configure logging, by using
```python
logging.basicConfig(level="ERROR")
```
as long as one does it *before* `import pyam`.

And it is possible to make it shut up with
```python
logging.getLogger().setLevel("ERROR")
```
even after the import is through.

Thoughts? @danielhuppmann @gidden @znicholls .
I do strongly favor the *Best practice* option but would prepare a PR for either!

coroa:
Fixes #449.

See also discussion in the issue.

- [x] Tests Added
- [x] Documentation Added

# Description of PR

As outlined as option 2a in https://github.com/IAMconsortium/pyam/issues/449#issuecomment-718725665 , this PR creates a `ConfigPseudoHandler` with the sole job of setting up logging after receiving the first log message.

To set up logging it calls `logging.basicConfig` and optionally emits a logging message about having it set up.

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -21,7 +21,7 @@ and methods.
 
 **Notebook logging behaviour**
 
-Pyam wants to provide sensitive defaults for users unfamiliar with
+|pyam| wants to provide sensible defaults for users unfamiliar with
 `setting up python's logging library <https://realpython.com/python-logging/#basic-configurations>`_,
 and therefore will provide a basic configuration by invoking
 
@@ -33,7 +33,7 @@ and therefore will provide a basic configuration by invoking
 if (and only if):
 
 1. it determines to be running within a notebook, and
-2. logging is still *unconfigured by the time the first logging message is to be emitted*.
+2. logging is still *unconfigured by the time the first logging message by |pyam| is to be emitted*.
 
 **Intersphinx mapping**
 
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -19,6 +19,22 @@ and methods.
    api/variables
 
 
+**Notebook logging behaviour**
+
+Pyam wants to provide sensitive defaults for users unfamiliar with
+`setting up python's logging library <https://realpython.com/python-logging/#basic-configurations>`_,
+and therefore will provide a basic configuration by invoking
+
+.. code-block:: python
+
+   import logging
+   logging.basicConfig(level="INFO", format="%(name)s - %(levelname)s: %(message)s")
+
+if (and only if):
+
+1. it determines to be running within a notebook, and
+2. logging is still *unconfigured by the time the first logging message is to be emitted*.
+
 **Intersphinx mapping**
 
 To use sphinx.ext.intersphinx_ for generating automatic links from your project
--- a/pyam/__init__.py
+++ b/pyam/__init__.py
@@ -26,8 +26,10 @@ try:
             u'IPython.OutputArea.prototype._should_scroll = '
             u'function(lines) { return false; }'
         )
-
-        log_msg = "Running in a notebook, setting root logging level to INFO"
+        log_msg = (
+            "Running in a notebook, "
+            "setting up a basic logging config at level INFO"
+        )
         defer_logging_config(
             logger, log_msg, level="INFO",
             format="%(name)s - %(levelname)s: %(message)s",
--- a/pyam/__init__.py
+++ b/pyam/__init__.py
@@ -10,26 +10,28 @@ from pyam.run_control import *
 from pyam.iiasa import read_iiasa  # noqa: F401
 from pyam.datareader import read_worldbank  # noqa: F401
 
+from pyam.logging import defer_logging_config
+
 logger = logging.getLogger(__name__)
-logger.addHandler(logging.NullHandler())
 
-# in Jupyter notebooks: disable autoscroll and set logger to info
+# in Jupyter notebooks: disable autoscroll and set-up logging
 try:
-    get_ipython().run_cell_magic(u'javascript', u'',
-                                 u'IPython.OutputArea.prototype._should_scroll = function(lines) { return false; }')
+    from ipykernel.zmqshell import ZMQInteractiveShell
+    from IPython import get_ipython
 
-    logger.setLevel(logging.INFO)
+    shell = get_ipython()
+    if isinstance(shell, ZMQInteractiveShell):
+        shell.run_cell_magic(
+            u'javascript', u'',
+            u'IPython.OutputArea.prototype._should_scroll = '
+            u'function(lines) { return false; }'
+        )
 
-    stderr_info_handler = logging.StreamHandler()
-    formatter = logging.Formatter('%(name)s - %(levelname)s: %(message)s')
-    stderr_info_handler.setFormatter(formatter)
-    logger.addHandler(stderr_info_handler)
-
-    log_msg = (
-        "Running in a notebook, setting `{}` logging level to `logging.INFO` "
-        "and adding stderr handler".format(__name__)
-    )
-    logger.info(log_msg)
+        log_msg = "Running in a notebook, setting root logging level to INFO"
+        defer_logging_config(
+            logger, log_msg, level="INFO",
+            format="%(name)s - %(levelname)s: %(message)s",
+        )
 
 except Exception:
     pass
--- a/tests/test_tutorials.py
+++ b/tests/test_tutorials.py
@@ -56,6 +56,16 @@ def test_pyam_first_steps(capsys):
     assert errors == []
     assert os.path.exists(os.path.join(tut_path, 'tutorial_export.xlsx'))
 
+    def has_log_output(cell):
+        return (
+            cell['cell_type'] == 'code' and
+            any(
+                'Running in a notebook' in output.get('text', '')
+                for output in cell['outputs']
+            )
+        )
+    assert any(has_log_output(cell) for cell in nb['cells'])
+
 
 def test_data_table_formats():
     fname = os.path.join(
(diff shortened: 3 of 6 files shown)


pyam/__init__.py
```diff
@@ -11,25 +11,26 @@
...
-    logger.info(log_msg)
+    from ipykernel.zmqshell import ZMQInteractiveShell
+    from IPython import get_ipython
+
+    shell = get_ipython()
+    if isinstance(shell, ZMQInteractiveShell):
+        shell.run_cell_magic(
+            u'javascript', u'',
+            u'IPython.OutputArea.prototype._should_scroll = function(lines) { return false; }'
+        )
```
coroa, line 25: The cell magic does not seem to have any effect for me! What is it's purpose? Should i remove it?

pyam/__init__.py
```diff
@@ -11,25 +11,26 @@
...
-    log_msg = (
-        "Running in a notebook, setting `{}` logging level to `logging.INFO` "
-        "and adding stderr handler".format(__name__)
-    )
-    logger.info(log_msg)
+    from ipykernel.zmqshell import ZMQInteractiveShell
+    from IPython import get_ipython
+
+    shell = get_ipython()
+    if isinstance(shell, ZMQInteractiveShell):
```
coroa, line 21: Testing against ZMQInteractiveShell differentiates, ipython in the terminal from the notebook.

pyam/__init__.py
danielhuppmann: By default, Jupyter notebook switches to scrolling within a cell output if it's too long (more than 100 lines).

Try running `[i for i in range(0, 110)]` in a completely fresh Jupyter notebook instance (not just restarted kernel of one notebook) and compare the output before and after running that cell magic.

@gidden had some very strong feelings about auto-scrolling in notebooks in the past (he strongly preferred switching it off), so I would not remove that section.

coroa: Ok, not a problem in jupyterlab, but I see the issue in the notebook. And I definitely wouldn't want to incur @gidden's wrath.

pyam/__init__.py
```diff
@@ -10,26 +10,30 @@
 from pyam.iiasa import read_iiasa  # noqa: F401
 from pyam.datareader import read_worldbank  # noqa: F401
 
+# appease linter
+from pyam.utils import ConfigPseudoHandler
```
danielhuppmann, line 14: don't think that this comment is relevant for future devs
```suggestion
from pyam.utils import ConfigPseudoHandler
```

danielhuppmann (commented):
I don't fully understand what kind of magic is supposed to be happening here, but I'm pretty sure it's not happening. Importing pyam in a Jupyter notebook and doing an operation that should write INFO-level output does not set up logging with INFO level and does not show output (noth.

(I tested as far as I could and the issue must be somewhere in the ConfigPseudoHandler).

A snippet where I would expect two INFO-log-statements ("running in a notebook" and "has required variable")
```
import pandas as pd
import pyam

TEST_DF = pd.DataFrame([
    ['model_a', 'scen_a', 'World', 'Primary Energy', 'EJ/yr', 1, 6.],
],
)

df = pyam.IamDataFrame(TEST_DF)
df.require_variable('Primary Energy')
```

coroa:
Ah, damn, it seems the logger was filtering out log messages below WARNING level before handing them to the handler, so it did set up correctly if you produced a warning message (which i tested), but not if you produced an info message.

The work-around I found is to temporarily set the loglevel of the `pyam.logger` to 19 (ie `logging.INFO - 1`) and then switch it back to NOTSET, once we are setting up the logging config. Not pretty, but effective.

image

`Javascript Error` is due to me running in jupyterlab rather than the notebook. It works now.

The failing pytest on the windows machines is a tooling problem (pip not finding the wheel package when trying to upgrade pip).

danielhuppmann:
Thanks @coroa - I can confirm that this behaves as I expected it from the point of view of an "analysis user" (per @gidden's comment in https://github.com/IAMconsortium/pyam/issues/449#issuecomment-718456694).

About the docs - maybe adding just a short paragraph at the [API docs overview page](https://github.com/IAMconsortium/pyam/blob/master/doc/source/api.rst), to have some reference how it is behaving in a notebook environment?

For testing, is it possible (without too much hassle) to check that the first cell of the first-steps tutorial writes an INFO statement as part of the notebook-testing?https://github.com/IAMconsortium/pyam/blob/2a569ff23811e9f779aa40e635872f4f6372f496/tests/test_tutorials.py#L53

tests/test_tutorials.py
```diff
@@ -56,6 +56,16 @@ def test_pyam_first_steps(capsys):
     assert errors == []
     assert os.path.exists(os.path.join(tut_path, 'tutorial_export.xlsx'))
 
+    def has_log_output(cell):
+        return (
+            cell['cell_type'] == 'code'
+            and any(
```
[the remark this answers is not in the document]
coroa: Charming, the "Best practice" descriptions are cyclic: [W504](https://www.flake8rules.com/rules/W503.html) suggests to use what triggers [W503](https://www.flake8rules.com/rules/W503.html) suggests to use what triggers W504 :).

Since [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) suggests to use the latter, i'll disable the W503.

doc/source/api.rst
```diff
@@ -19,6 +19,22 @@ and methods.
...
+
+.. code-block:: python
+
+   import logging
+   logging.basicConfig(level="INFO", format="%(name)s - %(levelname)s: %(message)s")
+
+if (and only if):
+
+1. it determines to be running within a notebook, and
+2. logging is still *unconfigured by the time the first logging message is to be emitted*.
```
danielhuppmann, line 36: this is "the first logging message to be emitted **by pyam**", right? Or really any logging message?

danielhuppmann (dismissed):
two minor suggestions for the docs page

doc/source/api.rst
```diff
@@ -19,6 +19,22 @@ and methods.
    api/variables
 
 
+**Notebook logging behaviour**
+
+Pyam wants to provide sensitive defaults for users unfamiliar with
```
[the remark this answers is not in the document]
coroa: Sure to both :)

doc/source/api.rst
coroa: Indeed

danielhuppmann (commented):
Looks good to me, thanks! I also marked docs and tests as "done" in the description.

coroa:
Great, good to go from my side as well

danielhuppmann:
@znicholls, you chimed in a lot in the discussion of the related issue, but I know you are super-busy, so I'm not formally asking for a review from you. Still, if you have a few minutes, it would be very much appreciated... (if I don't hear from you, I'll merge in ~2 days).

znicholls:
Looks good to me too

pyam/__init__.py
```diff
@@ -10,26 +10,28 @@
...
+
+    shell = get_ipython()
+    if isinstance(shell, ZMQInteractiveShell):
+        shell.run_cell_magic(
+            u'javascript', u'',
+            u'IPython.OutputArea.prototype._should_scroll = '
+            u'function(lines) { return false; }'
+        )
+
+        log_msg = "Running in a notebook, setting root logging level to INFO"
```
znicholls, line 30: I would add here that we are also running basic config for the user

coroa: I am trying to keep it short! How is

Running in a notebook, setting up a basic logging config at level INFO

danielhuppmann (approved):
Thanks @coroa for the improvement, thanks @znicholls for taking the time to review!

PR #451, opened 2020-10-30
Use ordinary pint conversion only if both units are parsable by pint

### LINKED ISSUE #442: Failing unit conversion from C to CO2
Unit conversion from CO2 to C (carbon) works as expected, but converting from C to CO2 fails.

```
import pandas as pd
import pyam

TEST_DF = pd.DataFrame([
    ['model_a', 'scen_a', 'World', 'Emissions', 't C/yr', 1, 6.],
],
)

pyam.IamDataFrame(TEST_DF).convert_unit('t C/yr', 't CO2/yr', context='gwp_AR5GWP100')
```

My hunch is that this is because `C` is defined as 'coulomb' in the standard pint registry.

Also, as reported in the related issue https://github.com/IAMconsortium/units/issues/23, I would expect the conversion to work without needing to specify a GWP-context.

coroa:

- [x] Tests Added
- [ ] ~Documentation Added~

# Description of PR

Fix unit conversions from `C` to `CO2eq`

When determining whether the ordinary pint conversion suffices, the previous implementation only considers whether the original unit can or cannot be parsed by pint. In the case of `C` pint misinterprets as Coulomb and then only fails when asked to convert to CO2eq, which it is unable to interpret.

Therefore, the main part of this change moves the `.to(newUnit)` conversion into the try-except clause, so that the late fail still triggers the GWP conversion.

The PR also deprecates the usage of prefixing GWP contexts with `gwp_` in favour of using the contexts of iam-units directly, ie. in place of `gwp_AR4GWP100`, it recommends `AR4GWP100`.

And a third additional change has been sneaked in:
A fast-path if the `current` unit has not been used in the dataframe

Fixes #442 .

### DIFF
--- a/pyam/units.py
+++ b/pyam/units.py
@@ -18,7 +18,7 @@ def convert_unit(df, current, to, factor=None, registry=None, context=None,
     try:
         where = ret._data.index.get_loc_level(current, 'unit')[0]
     except KeyError:
-        return ret
+        return None if inplace else ret
 
     index_args = [ret._data, 'unit', {current: to}]
 
--- a/pyam/units.py
+++ b/pyam/units.py
@@ -4,6 +4,7 @@ import re
 import iam_units
 import pint
 
+from pyam.logging import deprecation_warning
 from pyam.index import replace_index_values
 
 logger = logging.getLogger(__name__)
@@ -79,7 +80,7 @@ class UndefinedUnitError(pint.UndefinedUnitError):
     def __str__(self):
         return super().__str__() + (
             "\nGWP conversion with IamDataFrame.convert_unit() requires a "
-            "'gwp_...' *context* and mass-based *to* units.")
+            "*context* and mass-based *to* units.")
 
 
 def extract_species(expr):
@@ -104,11 +105,14 @@ def extract_species(expr):
 def convert_gwp(context, qty, to):
     """Helper for :meth:`convert_unit` to perform GWP conversions."""
     # Remove a leading 'gwp_' to produce the metric name
-    metric = (
-        context[len('gwp_'):]
-        if context is not None and context.startswith('gwp_')
-        else context
-    )
+    if context is not None and context.startswith('gwp_'):
+        context = context[len("gwp_"):]
+        deprecation_warning(
+            f"Use context='{context}' instead",
+            type='Prefixing a context with "gwp_"',
+            stacklevel=5
+        )
+    metric = context
 
     # Extract the species from *qty* and *to*, allowing supported aliases
     species_from, units_from = extract_species(qty[1])
--- a/pyam/units.py
+++ b/pyam/units.py
@@ -18,7 +18,7 @@ def convert_unit(df, current, to, factor=None, registry=None, context=None,
     try:
         where = ret._data.index.get_loc_level(current, 'unit')[0]
     except KeyError:
-        where = [False] * len(ret)
+        return ret
 
     index_args = [ret._data, 'unit', {current: to}]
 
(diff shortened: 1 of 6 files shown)


danielhuppmann:
thanks @coroa for tackling this issue! @khaeru, given that you have added this feature, it would be great if you could review...

khaeru (changes requested):
The PR template asks for an addition to the release notes. It would also be best to change in the documentation to ensure the new/changed behaviour is fully & precisely described.

These, plus perhaps some text in the description at the top of this PR, are necessary to in order review the code/test changes.

coroa:
> It would also be best to change in the documentation to ensure the new/changed behaviour is fully & precisely described.

The behaviour is not supposed to change.

I can add small note to https://pyam-iamc.readthedocs.io/en/stable/tutorials/unit_conversion.html#4.-Use-contexts-to-specify-conversion-metrics , that `AR4GWP` (w/o the leading `gwp_` is also a permissible argument), but i would rather like to have your ok that this change stays before that.

Did the rest.

khaeru:
Currently the docs for `IamDataFrame.convert_unit()` say:
https://github.com/IAMconsortium/pyam/blob/c1ec71fdb43ac425c30c9ee37c8128fd514b1d9b/pyam/core.py#L937-L942

When integrating the iam_units package with pyam, I preserved the "gwp_" prefix because that's what was there previously, and I was trying to minimize changes. Personally, I very much agree it is simpler if this is not required: one less thing to explain and for the user to remember. So I welcome the change. @danielhuppmann as maintainer would have the final veto.

If it stays, then this docstring should be changed to something like:
```
…
In this case, *context* must be the name of a specific global warming potential
(GWP) metric supported by :mod:`iam_units`, e.g. 'AR5GWP100' (optionally
prefixed by 'gwp\_', e.g. 'gwp\_AR5GWP100').
```

This is what I mean by "precisely described".

pyam/units.py
```diff
@@ -18,7 +18,7 @@ def convert_unit(df, current, to, factor=None, registry=None, context=None,
     try:
         where = ret._data.index.get_loc_level(current, 'unit')[0]
     except KeyError:
-        where = [False] * len(ret)
+        return ret
```
coroa, line 21:
```suggestion
    return None if inplace else ret
```

danielhuppmann:
Thanks @coroa and @khaeru! I agree that using the context name (`AR5-GWP100`) without leading `gwp_` as standard/recommended approach is preferred. I think the `gwp_*` approach is a legacy from an earlier implementation of the iam-units package. Fine for me to keep or mark as deprecated, to be removed in release >=10.0

coroa:
Great, then I'll update the PR to reflect this, ie include doc amendments and emit a deprecation warning for the longer context name.

coroa:
I think, I fixed all documentation correctly. The test failures look like an IIASA server issue.

tests/test_units.py
```diff
@@ -92,7 +92,10 @@ def test_convert_unit_with_custom_registry(test_df):
 
     # Using "-equiv" after a unit should make no difference
     ('AR4GWP100', 'co2-equiv', 1.),
-    ('AR5GWP100', 'ch4-equiv', 28)
+    ('AR5GWP100', 'ch4-equiv', 28),
+
+    # Converting C -> CO2e should work
+    ('AR4GWP100', 'C', 11 / 3)
```
danielhuppmann, line 98: conversion from C to CO2 or CO2-eq shouldn't require a context at all, see https://github.com/IAMconsortium/units/issues/23 - adding a todo as a reminder to update the test

```suggestion
    # TODO remove context once IAMconsortium/units#23 is resolved
    ('AR4GWP100', 'C', 11 / 3)
```

danielhuppmann (approved):
Thank you @coroa!

Issue #351, opened 2020-03-24
Nans in additional columns


Rlamboll:
When a dataframe has additional data created, e.g. by interpolate, additional columns are often given np.nan as the value, rather than, say, an empty string or a 0.0.. This is presumably a choice rather than a bug. However when there are nans in these columns, they will not appear in timeseries, and will also disappear if the dataframe is recalculated by `pyam.IamDataFrame(database_with_nans.data)`. This seems like a bug.

danielhuppmann:
Thanks for raising this issue @Rlamboll but I'm not sure which part here you see as a bug and which is bad (or unexpected) design choices - or if this is related to some not-yet-supported aspect of continuous-time data.

My understanding:
 - [interpolate()](https://pyam-iamc.readthedocs.io/en/latest/api/iamdataframe.html#pyam.IamDataFrame.interpolate) (and the underlying [fill_series()](https://pyam-iamc.readthedocs.io/en/latest/api/timeseries.html)) return `nan` if the requested time is outside of the range of a specific timeseries (doing something else would be not inter- but extrapolation)
 - `nan` are dropped when initializing a new `IamDataFrame` because `pandas` behaves strangely with `nan` in columns (examples provided @znicholls, see the [docs](https://pyam-iamc.readthedocs.io/en/latest/data.html#custom-columns-of-the-data-table)). But calling [timeseries(https://pyam-iamc.readthedocs.io/en/latest/api/iamdataframe.html#pyam.IamDataFrame.timeseries)] fills `nan` in any time-column where row data is missing (ie. if at least one data point exists for that time index).

Can you specify or provide a short test case that illustrates what you would expect to see?

Rlamboll:
Perhaps I wasn't clear, all these nans occur in additional columns. Interpolate returns a value correctly in the `value` column, but not in other columns (not a bug). However I'd expect that initialising a dataframe only drops cases where important columns are nan, not where any column is nan. If it is going to do that, can we infill those columns with something other than nan. Here is an example (appologies for lack of indentation in this format):


import pandas as pd
import pyam


tdb = pd.DataFrame(
        [["_ma", "_sa", "World", "CO2", "Gt CO2", 0, 2, 3], 
            ["_ma", "_sa", "World", "CH4", "Gt CH4", 0, 200, 300],
        ],
        columns=["model", "scenario", "region", "variable", "unit", "sub_annual"] + [2050, 2070],
    )
df = pyam.IamDataFrame(tdb)
[...]

PR #456, opened 2020-11-10
Interp update

### LINKED ISSUE #240: `interpolate` does not follow other return conventions
We recently had a private communication regarding confusion with the `interpolate` function. The user was trying to do something like:

```
df_interp = df.filter().interpolate()
```

The returned object was *not* an interpolated dataframe. We advised the user as follows:

```
alldata = pyam.IamDataFrame(data='iamc15_scenario_data_world_r1.1.xlsx')
df = alldata.filter(variable='Emissions|CO2',scenario='SSP1-19') # save the filtered dataframe explicitly
df.interpolate(2063)  # interpolate the filtered dataframe
```

I would argue this breaks our implicit datamodel of dataframe methods returning a copy unless `inplace=True` is provided. 

@danielhuppmann can you provide your thoughts?

While making such a change would break existing API, I think there is a strong argument for all functions to conform to our (from `pandas`) data model.

### LINKED ISSUE #371: `interpolate` fails if 'accidently applied twice'
If the exact same interpolate call is used twice, the underlying data layout is appended in both `wide` and `long`.

Example code:

```
df.interpolate(2016) # works
df.interpolate(2016) # bad data now
```

Image attached
interp

gidden:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR fixes a bug when there are no values to interpolate (all scenarios already have this value) and allows users to past a list-like input for interpolation of multiple time points.

closes #240
closes #371

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -507,14 +507,14 @@ class IamDataFrame(object):
         """
         ##
         # TODO remove, and add kwarg inplace=False in release >= 0.10
-        if 'inplace' is None:  # noqa: F632
+        if inplace is None:
             deprecation_warning(
                 'Behavior of `interpolate` will change to `inplace=False` '
                 'as default in a future release. Set the kwarg explicitly '
                 'to avoid this warning. Use `inplace=True` to keep current '
                 'behavior.'
             )
-            inplace = False
+            inplace = True
         ##
 
         # setup
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -545,7 +545,7 @@ class IamDataFrame(object):
 
         # replace underlying data object
         # TODO naming time_col could be done in timeseries()
-        df.columns.name = ret.time_col 
+        df.columns.name = ret.time_col
         df = df.stack()  # long-data to pd.Series
         df.name = 'value'
         ret._data = df
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -507,7 +507,7 @@ class IamDataFrame(object):
         """
         ##
         # TODO remove, and add kwarg inplace=False in release >= 0.10
-        if 'inplace' == None:
+        if 'inplace' is None:
             deprecation_warning(
                 'Behavior of `interpolate` will change to `inplace=False` '
                 'as default in a future release. Set the kwarg explicitly '
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -506,7 +506,7 @@ class IamDataFrame(object):
             passed to :meth:`pandas.DataFrame.interpolate`
         """
         ##
-        # TODO remove, and add kwarg inplace=False in next release
+        # TODO remove, and add kwarg inplace=False in release >= 0.10
         if 'inplace' not in kwargs:
             deprecation_warning(
                 'Behavior of `interpolate` will change default to `False` in '
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -507,7 +507,7 @@ class IamDataFrame(object):
         """
         ##
         # TODO remove, and add kwarg inplace=False in release >= 0.10
-        if 'inplace' is None:  # noqa: E632
+        if 'inplace' is None:  # noqa: F632
             deprecation_warning(
                 'Behavior of `interpolate` will change to `inplace=False` '
                 'as default in a future release. Set the kwarg explicitly '
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -489,7 +489,7 @@ class IamDataFrame(object):
         df = df.unstack(level=columns, fill_value=fill_value)
         return df
 
-    def interpolate(self, time, **kwargs):
+    def interpolate(self, time, inplace=None, **kwargs):
         """Interpolate missing values in the timeseries data
         
         This method uses :meth:`pandas.DataFrame.interpolate`,
@@ -507,7 +507,7 @@ class IamDataFrame(object):
         """
         ##
         # TODO remove, and add kwarg inplace=False in release >= 0.10
-        if 'inplace' not in kwargs:
+        if 'inplace' is None:
             deprecation_warning(
                 'Behavior of `interpolate` will change to `inplace=False` '
                 'as default in a future release. Set the kwarg explicitly '
@@ -516,7 +516,7 @@ class IamDataFrame(object):
             )
             inplace = True
         else:
-            inplace = kwargs.pop('inplace')
+            inplace = False
         ##
 
         # setup
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -507,8 +507,8 @@ class IamDataFrame(object):
         # TODO remove, and add kwarg inplace=False in next release
         if 'inplace' not in kwargs:
             deprecation_warning(
-                'Behavior of `interpolate` will change default to `False` in a '
-                'future release. Explicitly set it to avoid this warning.'
+                'Behavior of `interpolate` will change default to `False` in '
+                'a future release. Explicitly set it to avoid this warning.'
             )
             inplace = True
         else:
@@ -520,11 +520,11 @@ class IamDataFrame(object):
         interp_kwargs = dict(method='slinear', axis=1)
         interp_kwargs.update(kwargs)
         time = list(time) if islistable(time) else [time]
-        ## TODO - have to explicitly cast to numpy datetime to sort later,
+        # TODO - have to explicitly cast to numpy datetime to sort later,
         # could enforce as we do for year below
         if self.time_col == 'time':
             time = list(map(np.datetime64, time))
-        if self.time_col == 'year' and not all(isinstance(x, int) for x in time):
+        elif not all(isinstance(x, int) for x in time):
             raise ValueError(
                 'The `time` argument `{}` contains non-integers'.format(time)
             )
@@ -544,14 +544,15 @@ class IamDataFrame(object):
             df[col] = newdf[col]
 
         # replace underlying data object
-        df.columns.name = ret.time_col  # TODO this could be done in timeseries()
+        # TODO naming time_col could be done in timeseries()
+        df.columns.name = ret.time_col  
         df = df.stack()  # long-data to pd.Series
         df.name = 'value'
         ret._data = df
 
         if not inplace:
             return ret
-        
+
     def swap_time_for_year(self, inplace=False):
         """Convert the `time` column to `year`.
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -545,7 +545,7 @@ class IamDataFrame(object):
 
         # replace underlying data object
         # TODO naming time_col could be done in timeseries()
-        df.columns.name = ret.time_col  
+        df.columns.name = ret.time_col 
         df = df.stack()  # long-data to pd.Series
         df.name = 'value'
         ret._data = df
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -490,8 +490,10 @@ class IamDataFrame(object):
         return df
 
     def interpolate(self, time, **kwargs):
-        """Interpolate missing values in timeseries (linear interpolation by
-        default). Internally, :meth:`pandas.DataFrame.interpolate` is used.
+        """Interpolate missing values in the timeseries data
+        
+        This method uses :meth:`pandas.DataFrame.interpolate`,
+        which applies linear interpolation by default
 
         Parameters
         ----------
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -512,7 +512,7 @@ class IamDataFrame(object):
         df = self.timeseries()
         if time in df.columns:
             df = df[np.isnan(df[time])]
-        if df.empty: # no values with time missing, empty op
+        if df.empty:  # no values with time missing, empty op
             return
 
         # apply fill_series, re-add time dimension to index, set series name
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -491,7 +491,7 @@ class IamDataFrame(object):
 
     def interpolate(self, time, inplace=None, **kwargs):
         """Interpolate missing values in the timeseries data
-        
+
         This method uses :meth:`pandas.DataFrame.interpolate`,
         which applies linear interpolation by default
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -514,8 +514,6 @@ class IamDataFrame(object):
                 'to avoid this warning. Use `inplace=True` to keep current '
                 'behavior.'
             )
-            inplace = True
-        else:
             inplace = False
         ##
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -552,7 +552,7 @@ class IamDataFrame(object):
         df.columns.name = ret.time_col
         df = df.stack()  # long-data to pd.Series
         df.name = 'value'
-        ret._data = df
+        ret._data = df.sort_index()
 
         if not inplace:
             return ret
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -507,7 +507,7 @@ class IamDataFrame(object):
         """
         ##
         # TODO remove, and add kwarg inplace=False in release >= 0.10
-        if 'inplace' is None:
+        if 'inplace' is None:  # noqa: E632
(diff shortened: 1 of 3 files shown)


gidden:
seems like test failures are related to iiasa queries!

danielhuppmann:
> seems like test failures are related to iiasa queries!

already pinged @peterkolp and @fonfon about this...

danielhuppmann:
@gidden, is this related to #371 - not quite sure what the bug to be resolved is?

danielhuppmann (approved):
Lgtm, thanks @gidden!

Minor quibble: I guess a substantial performance increase could be had by avoiding doing `df.timeseries()` for every iteration, instead trying something like:

```
        df = self.timeseries()
        time = set(time) if islistable(time) else set([time]). # cast to set to drop duplicates
        lst = [None] * len(time)

        # apply fill_series, re-add time dimension to index, set series name
        for i, t in enumerate(time):
            lst[i] = df[np.isnan(df[t])].apply(fill_series, raw=False, axis=1, time=t).dropna()
            lst[i].index = append_index_level(
                index=_values.index, codes=[0] * len(_values),
                level=[t], name=self.time_col, order=self._data.index.names)
            lst[i].name = 'value'

        # append interpolated values to `_data` and sort index
        self._data = self._data.append(pd.concat(lst)).sort_index()
```

znicholls:
Looks like @danielhuppmann has the review under control. My only thought: if you cast to wide format first, could you get an extra performance boost by using panda’s or numpy’s inbuilt Interpolation rather than looping over all the times?

coroa:
> My only thought: if you cast to wide format first, could you get an extra performance boost by using panda’s or numpy’s inbuilt Interpolation rather than looping over all the times?

Agreed.

```python
if not islistable(time):
    time = [time]
pyam.IamDataFrame(
    idf.timeseries().reindex(columns=np.sort(idf.year + time)).interpolate(method='slinear', axis=1),
    meta=idf.meta
)
```
has proven to beat pyam's implementation by heaps and bounds.

danielhuppmann:
Indeed, it would be great to switch to a more performant implementation of the interpolation!

One issue to keep in mind, though, is that [pandas.DataFrame.interpolate](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.interpolate.html) fills in all missing values in the dataframe, not just the values for one new column. That's why I didn't use it at the time. And [numpy.interp](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) assumes that the values are constant outside the timeseries domain, which is also not a valid general assumption.

znicholls:
> is that [pandas.DataFrame.interpolate](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.interpolate.html) fills in all missing values in the dataframe, not just the values for one new column

Yes you'd have to choose the subset you want to interpolate first if you didn't just want to interpolate everything. 

Another thought: you could just also change the API so the user just specifies the new time points they desire (including any which are already present) rather than just adding **extra** time points alone.

So, say you had data for 2010, 2020 and 2030 and you wanted to interpolate to have 2010, 2015, 2020, 2025 and 2030. 

In the current implementation you'd do

```python
df.interpolate([2015, 2025])
```

But we could just make it so it's

```python
df.interpolate([2010, 2015, 2020, 2025, 2030])
```

And just leave all the calculations to pandas. That interface would mean that interpolating would always leave you with a nan-free dataset.

> And [numpy.interp](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) assumes that the values are constant outside the timeseries domain

[scipy interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) will extrapolate if you want with the `fill_value` argument.

p.s. @coroa 

> heaps and bounds

was that a typo?

coroa:
> p.s. @coroa
> 
> > heaps and bounds
> 
> was that a typo?



> # Please confirm that this PR has done the following:
> 
>     * [x]  Tests Added
> 
>     * [x]  Documentation Added
> 
>     * [ ]  Description in RELEASE_NOTES.md Added
> 
> 
> ## Adding to RELEASE_NOTES.md (remove section after adding to RELEASE_NOTES.md)
> 
> Please add a single line in the release notes similar to the following:
> 
> ```
> - (#XX)[http://link-to-pr.com] Added feature which does something
> ```
> 
> # Description of PR
> 
> This PR fixes a bug when there are no values to interpolate (all scenarios already have this value) and allows users to past a list-like input for interpolation of multiple time points.



> > is that [pandas.DataFrame.interpolate](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.interpolate.html) fills in all missing values in the dataframe, not just the values for one new column
> 
> Yes you'd have to choose the subset you want to interpolate first if you didn't just want to interpolate everything.
> 
> Another thought: you could just also change the API so the user just specifies the new time points they desire (including any which are already present) rather than just adding **extra** time points alone.
> 
> So, say you had data for 2010, 2020 and 2030 and you wanted to interpolate to have 2010, 2015, 2020, 2025 and 2030.
> 
> In the current implementation you'd do
> 
> ```python
> df.interpolate([2015, 2025])
> ```
> 
> But we could just make it so it's
> 
> ```python
> df.interpolate([2010, 2015, 2020, 2025, 2030])
> ```
> 
> And just leave all the calculations to pandas. That interface would mean that interpolating would always leave you with a nan-free dataset.
> 
> > And [numpy.interp](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) assumes that the values are constant outside the timeseries domain
> 
> [scipy interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) will extrapolate if you want with the `fill_value` argument.
>

`method='slinear'` of pandas is indeed passed to scipy's `interp1d` (together with all extra `**kwargs`) and its default is to raise a `ValueError`, unless `fill_value` is set to f.ex. `'extrapolate'.

> p.s. @coroa
> 
> > heaps and bounds
> 
> was that a typo?

something like that :)

coroa:
Otherwise, something along the lines of (untested)
```python
df = self.timeseries()
new_values = df.apply(lambda s: pd.Series(scipy.interpolate.interp1d(s.index, s.values, method='slinear', **kwargs)(time), time), axis=1)
for year in time:
    if year in df:
        df[year].fillna(new_values[year], inplace=True)
    else:
        i = bisect(df.columns, year)
        df.insert(i, year, new_values[year])
return pyam.IamDataFrame(df, meta=self.meta)
```
should work fine.

gidden:
Thanks for the discussion all! I'll take a look at the suggestions, update, and flag for final review.

gidden:
Ok folks, I have updated implementation as attached in this notebook. Let me know what you think - it should maintain prior behavior (only updating columns asked for). 
[interpolate_implementation (1).pdf](https://github.com/IAMconsortium/pyam/files/5540789/interpolate_implementation.1.pdf)

gidden:
Ok... this took me a bit further down a rabbit hole than I expected. The basic approach follows what @coroa suggested earlier, and includes:

1. pivot to wide-format once
2. interpolate using `pandas.DataFrame.interpolate`, passing along kwargs
3. include back only time values that were asked for (keeping old behavior)
4. return a copy by default (in the future)

There were a few issues I ran into that I will raise to others. Namely an issue for @danielhuppmann / @Rlamboll regarding interpolation with `extra_col`s and an issue with `datetime.datetime` for @znicholls. I will ping on the specific lines.

@coroa happy if you take a look at the final version here too.

pyam/core.py
```diff
@@ -489,34 +489,69 @@ def pivot_table(self, index, columns, values='value',
...
+
+        # setup
+        ret = self.copy() if not inplace else self
+        interp_kwargs = dict(method='slinear', axis=1)
+        interp_kwargs.update(kwargs)
+        time = list(time) if islistable(time) else [time]
+        # TODO - have to explicitly cast to numpy datetime to sort later,
+        # could enforce as we do for year below
+        if self.time_col == 'time':
+            time = list(map(np.datetime64, time))
```
gidden, line 526: @znicholls I had to convert to `np.datetime` here in order to make calls to `np.sort` and `np.unique` later on - numpy and datetime objects didn't play nice together. If you have a better suggestion let me know, but I observed that the datetime pyam object internally stores as `np.datetime`, so had to convert input.

tests/test_core.py
```diff
@@ -627,10 +627,54 @@ def test_interpolate(test_pd_df):
...
+    ))
+    obs = df.interpolate([2005, 2012], inplace=False)
+    assert_iamframe_equal(obs, exp)
+
+
 def test_interpolate_extra_cols():
     # check hat interpolation with non-matching extra_cols has no effect (#351)
     EXTRA_COL_DF = pd.DataFrame([
         ['foo', 2005, 1],
+        ['foo', 2010, 2],
```
gidden, line 676: @Rlamboll and @danielhuppmann I had to update this test to get correct behavior. The current implementation uses the wide-form `timeseries` object, which treats each index  value (row) as an indepedent observation. Therefore, the `foo` and `bar` entries here are each interpolated. If I left it as is, I would get an error, because there are too few datapoints to interpolate (i.e., 1 each)

Rlamboll: OK, I believe this is consistent with the current state as of the discussion in #351. It's unclear whether this is the desired behaviour or not, but it's what we're used to.

Rlamboll (approved):

pyam/core.py
```diff
@@ -489,34 +489,69 @@ def pivot_table(self, index, columns, values='value',
...
+        """Interpolate missing values in timeseries (linear interpolation by
+        default). Internally, :meth:`pandas.DataFrame.interpolate` is used.
 
         Parameters
         ----------
-        time : int or datetime
+        time : int or datetime, or list-like thereof
              Year or :class:`datetime.datetime` to be interpolated.
              This must match the datetime/year format of `self`.
+        inplace : bool, default False
```
danielhuppmann, line 501: Per numpy guidelines see section 4 [here](https://numpydoc.readthedocs.io/en/latest/format.html)
```suggestion
        inplace : bool, optional
```

pyam/core.py
```diff
@@ -489,34 +489,69 @@ def pivot_table(self, index, columns, values='value',
...
-        _values.index = append_index_level(
-            index=_values.index, codes=[0] * len(_values),
-            level=[time], name=self.time_col, order=self._data.index.names)
-        _values.name = 'value'
+        # replace underlying data object
+        # TODO naming time_col could be done in timeseries()
+        df.columns.name = ret.time_col
+        df = df.stack()  # long-data to pd.Series
+        df.name = 'value'
+        ret._data = df
```
danielhuppmann, line 551: To get `timeseries()` in the right order after the operation
```suggestion
        ret._data = df.sort_index()
```

tests/test_core.py
danielhuppmann: Slightly concerned that this behavior might throw users that have an IamDataFrame where just one timeseries-row has only one datapoint...

tests/test_core.py
```diff
@@ -642,8 +686,11 @@ def test_interpolate_extra_cols():
     df2 = df.copy()
     df2.interpolate(2007)
 
-    # assert that interpolation didn't change any data
-    assert_iamframe_equal(df, df2)
+    # interpolate should work as if this is a new index
```
danielhuppmann, line 689: Not clear to me what this comment means...

tests/test_core.py
```diff
@@ -627,10 +627,54 @@ def test_interpolate(test_pd_df):
     assert all([True if isstr(i) else ~np.isnan(i) for i in df.data.foo])
 
 
+def test_interpolate_time_exists(test_pd_df):
+    _df = test_pd_df.copy()
+    df = IamDataFrame(_df)
```
danielhuppmann, line 632: You can use `test_df_year` to avoid having to cast to an IamDataFrame in the test... (Also, `test_df_pdf` already serves a copy).

```suggestion
def test_interpolate_time_exists(test_df_year):
```

tests/test_core.py
```diff
@@ -627,10 +627,54 @@ def test_interpolate(test_pd_df):
...
+    df.interpolate(2005)
+    obs = df.filter(year=2005).data['value'].reset_index(drop=True)
+    exp = pd.Series([1.0, 0.5, 2.0], name='value')
+    pd.testing.assert_series_equal(obs, exp)
+
+
+def test_interpolate_multiple_times(test_pd_df):
+    _df = test_pd_df.copy()
+    df = IamDataFrame(_df)
```
danielhuppmann, line 641: As above...

Also suggest to rename the test for clarity
```suggestion
def test_interpolate_with_list(test_df_year):
```

pyam/core.py
```diff
@@ -489,34 +489,69 @@ def pivot_table(self, index, columns, values='value',
         df = df.unstack(level=columns, fill_value=fill_value)
         return df
 
-    def interpolate(self, time):
-        """Interpolate missing values in timeseries (linear interpolation)
+    def interpolate(self, time, **kwargs):
```
danielhuppmann, line 492: Wondering if it makes sense to already include the kwarg `inplace=None` here to enable auto-complete...

pyam/core.py
```diff
@@ -489,34 +489,69 @@ def pivot_table(self, index, columns, values='value',
...
              This must match the datetime/year format of `self`.
+        inplace : bool, default False
+            if True, do operation inplace and return None
+        kwargs
+            passed to :meth:`pandas.DataFrame.interpolate`
         """
-        if self.time_col == 'year' and not isinstance(time, int):
+        ##
+        # TODO remove, and add kwarg inplace=False in next release
+        if 'inplace' not in kwargs:
```
danielhuppmann, line 508: if suggestion above to add kwarg to signature is implemented
```suggestion
        if 'inplace' is None:
```

danielhuppmann (changes requested):
Looks really great, thanks @gidden! Just a bunch of minor comments...

I ran a test on my standard large-data ensemble (4 scenarios, 2m data points) and can report:
 - old implementation (using two separate calls for interpolation): 64.4 sec
 - new implementation (using two separate calls for interpolation): 50.8 sec
 - new implementation (calling it via a list): 29.3 sec

pyam/core.py
znicholls: Yes the datetime vs. np.datetime difference is a constant source of annoyance so this casting is unavoidable I think. Could potentially be removed from pyam if the backend were xarray (it has better support for all sorts of time axis handling than pandas), but that's a job for another day.

gidden: 👍

tests/test_core.py
gidden: For the moment, an error will be raised if trying to interpolate with a single data point. I suggest we allow this, as I am not sure we 'want' to support a use case of pyam data with a single data point (this is what metadata is for, right)? @danielhuppmann I will let you mark if resolved.

tests/test_core.py
gidden: have cleared it up

gidden:
Ok @danielhuppmann - should be good for final review now. I must admit, I am a bit surprised at the rather minor speedups (@coroa can you confirm these numbers match your observations?). But in general, this PR was meant to support additional input, not optimization, so no need to block it there.

tests/test_core.py
danielhuppmann: I'm worried if a user has a table like the one below and wants to interpolate for 2015.

| variable       | 2010 | 2020 | 2030 |
|----------------|------|------|------|
| Primary Energy | 1    | 2    | 3    |
| Population     | 5    | nan  | nan  |

I would expect a return-object like

| variable       | 2010 | 2015 | 2020 | 2030 |
|----------------|------|------|------|------|
| Primary Energy | 1    | 1.5 | 2    | 3    |
| Population     | 5    | nan | nan  | nan  |

danielhuppmann (approved):
Thanks @gidden, really nice!

danielhuppmann:
resolved merge conflicts

PR #458, opened 2020-11-17
Enable `Path` as arg 

danielhuppmann:

- ~Tests Added~
- [x] Documentation Added

# Description of PR

This PR was initiated by a comment by @znicholls (https://github.com/IAMconsortium/pyam/pull/454#discussion_r524185343) about using `tmpdir` facility in the tests.

Changes:
 - enable initializing an `IamDataFrame` with a `Path` or `LocalPath` object (in addition to passing a file path as string)
 - raise a dedicated error message when initializing without meaningful args 'IamDataFrame constructor not properly called!' rather than showing the message `Initializing from list is not supported, ...` (which was previously shown in that case)
 - rework the initialization workflow and add more comments
 - use `tmpdir` facility in the io-tests

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 import itertools
 import logging
 import string
@@ -108,7 +109,9 @@ def write_sheet(writer, name, df, index=False):
 
 def read_pandas(path, default_sheet='data', *args, **kwargs):
     """Read a file and return a pandas.DataFrame"""
-    if path.endswith('csv'):
+    if isinstance(path, Path) and path.suffix == '.csv':
+        df = pd.read_csv(path, *args, **kwargs)
+    elif isstr(path) and path.endswith('csv'):
         df = pd.read_csv(path, *args, **kwargs)
     else:
         xl = pd.ExcelFile(path)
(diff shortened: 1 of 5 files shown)


pyam/utils.py
```diff
@@ -108,7 +109,9 @@ def write_sheet(writer, name, df, index=False):
 
 def read_pandas(path, default_sheet='data', *args, **kwargs):
     """Read a file and return a pandas.DataFrame"""
-    if path.endswith('csv'):
+    if isinstance(path, Path) and path.suffix == '.csv':
+        df = pd.read_csv(path, *args, **kwargs)
+    elif isstr(path) and path.endswith('csv'):
```
znicholls, line 114: given the casting above, is it even possible to get here?

znicholls (approved):
lgtm

pyam/utils.py
danielhuppmann: That's what I thought, then this one bit me... https://github.com/IAMconsortium/pyam/blob/33156c62a95465785d7fc08b80fa3408ebe3eb36/pyam/core.py#L1640

And it got me thinking that some users may be calling that function directly for convenience in other situations, so rather stay backward-compatible...

znicholls: also good

danielhuppmann: On second thought, I think you're right... Avoid duplication within the function and just do the casting in the other location where the function is called...

PR #459, opened 2020-11-17
Add get_variable_components() function incl doc and tests

mabudz:

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR

This pull request adds a new function 'get_variable_components()' to utils.py. It returns (or merges and returns) the variable component(s) in regard to the hierarchy (indicated by ``|``) of x to the specified level(s).

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -585,3 +585,10 @@ def reduce_hierarchy(x, depth):
     _x = x.split('|')
     depth = len(_x) + depth - 1 if depth < 0 else depth
     return '|'.join(_x[0:(depth + 1)])
+
+def get_variable_components(x, depths):
+    """Return (or merge and return) the variable component(s) in regard to
+    the hierarchy (indicated by ``|``) of x to the specified depth(s)"""
+    depths = [depths] if type(depths) == int else depths
+    _x = x.split('|')
+    return '|'.join([_x[d] for d in depths])
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -587,9 +587,22 @@ def reduce_hierarchy(x, depth):
     return '|'.join(_x[0:(depth + 1)])
 
 
-def get_variable_components(x, depths):
-    """Return (or merge and return) the variable component(s) in regard to
-    the hierarchy (indicated by ``|``) of x to the specified depth(s)"""
-    depths = [depths] if type(depths) == int else depths
+def get_variable_components(x, level, join=False):
+    """Return components for requested level in a list or join these in a str.
+
+    Parameters
+    ----------
+    x : str
+        Uses ``|`` to separate the components of the variable.
+    level : int or list of int
+        Position of the component.
+    join : bool or str, optional
+        If True, IAMC-style (``|``) is used as separator for joined components.
+    """
+    level = [level] if type(level) == int else level
     _x = x.split('|')
-    return '|'.join([_x[d] for d in depths])
+    if join is False:
+        return [_x[i] for i in level] if islistable(level) else _x[level]
+    else:
+        join = '|' if join is True else join
+        return join.join([_x[i] for i in level])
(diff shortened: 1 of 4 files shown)


pyam/utils.py
```diff
@@ -585,3 +585,11 @@ def reduce_hierarchy(x, depth):
     _x = x.split('|')
     depth = len(_x) + depth - 1 if depth < 0 else depth
     return '|'.join(_x[0:(depth + 1)])
+
+
+def get_variable_components(x, depths):
```
danielhuppmann, line 590: Suggest to rename the arg to `level` similar to [pandas.Index.get_level_values(level)](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.get_level_values.html)

Also, I think that the most common use case will to just separate the components and return as a list, so I would change the behavior to the following options:

```python
get_variable_components('foo|bar|baz, [0, 1])  # default returns list
> ['foo', 'bar']
get_variable_components('foo|bar|baz, [0, 1], join=True)  # True returns IAMC-style
> 'foo|bar'
get_variable_components('foo|bar|baz, [0, 1], join='-')  # any other symbol can be used to concatenate
> 'foo-bar'
```

```suggestion
def get_variable_components(x, level, join=False):
```

pyam/utils.py
```diff
@@ -585,3 +585,11 @@ def reduce_hierarchy(x, depth):
...
+    """Return (or merge and return) the variable component(s) in regard to
+    the hierarchy (indicated by ``|``) of x to the specified depth(s)"""
+    depths = [depths] if type(depths) == int else depths
+    _x = x.split('|')
+    return '|'.join([_x[d] for d in depths])
```
danielhuppmann, line 595: To implement the change suggested above, maybe try something like
```suggestion
    if join is False:
        return [x[i] for i in level] if islistable(level) else x[level]
    join = '|' if join is True else join
    return '|'.join([_x[d] for d in depths])
```

pyam/utils.py
```diff
@@ -585,3 +585,11 @@ def reduce_hierarchy(x, depth):
     _x = x.split('|')
     depth = len(_x) + depth - 1 if depth < 0 else depth
     return '|'.join(_x[0:(depth + 1)])
+
+
+def get_variable_components(x, depths):
+    """Return (or merge and return) the variable component(s) in regard to
```
danielhuppmann, line 591: Please use a one-line summary, then an empty line, then the rest of the description, see the [numpydoc docs](https://numpydoc.readthedocs.io/en/latest/format.html)

pyam/utils.py
mabudz: Great suggestions! I will implement all ;-)

tests/test_utils.py
```diff
@@ -213,3 +213,27 @@ def test_merge_meta():
 
     obs = utils.merge_meta(left, right, ignore_meta_conflict=True)
     pdt.assert_frame_equal(exp, obs)
+
+
+def test_get_variable_components_int():
+    assert utils.get_variable_components('foo|bar|baz', 1) == ['bar']
```
danielhuppmann, line 219: I was a bit surprised why calling the function with an integer would return a list, not the string. Then I realized that this line 
```python
        level = [level] if type(level) == int else level
```
in the function turn the `level` arg into a list and makes the second part of 
```python
        return [_x[i] for i in level] if islistable(level) else _x[level]
```
obsolete (because `level` will always be a list at this point).

If you agree that the returning the string makes more sense, moving the cast-to-list line into the `else` section would do the trick...

mabudz: I agree. And by the way a very nice way to internalize the git workflow :)

danielhuppmann (approved):
Thanks @mabudz for this nice improvement! 5 new tests for 7 new lines of actual code is a really nice ratio...

Please rebase and add a line to the release notes, then good to go!

mabudz:
Since there seem to be no merge conflicts between branches, do you mean to merge the commits into one?

danielhuppmann:
My guess is that there will be a merge conflict after you add the line to the release notes, so you can either:
1. rebase (or squash & rebase, ie what you suggest), then add a new commit with the release note
2. add a new commit, then resolve the merge conflict manually

PR #462, opened 2020-11-28
release candidate 0.9.0

danielhuppmann:
This PR

 - updates the release notes with a highlights section
 - kicks the can of removing deprecated functions down the road because it not even two months since the last release


danielhuppmann:
merging after publishing on test.pypi.org

PR #481, opened 2021-01-11
Enable custom index columns

danielhuppmann:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- ~Name of contributors Added to AUTHORS.rst~
- [x] Description in RELEASE_NOTES.md Added

# Description of PR

Per discussions with related energy modelling projects (in particular OpenEnergyPlatform, SENTINEL), it became obvious that the restriction of the IAMC format to use `['model', 'scenario']` as the index (ie the organizing structure of timeseries data) is a limitation of the current package implementation.

Two specific examples:
 - The OpenEnergyPlatform is discussing a data model indexed on `['source', 'scenario']` to highlight that not all data are derived from models but can also come from *reference sources* <sup>[[citation needed]]()</sup>
 - When reading multiple versions of a model/scenario from an IIASA database, the API returns the index `['model', 'scenario', 'version]`, which can currently not be processed by pyam.

This PR extends the IamDataFrame class to accept an `index` argument such that timeseries data following the two examples above (or other index columns) can be handled by pyam. This PR is intended for discussion and as a first step to full-fledged support - only the initialization and `IamDataFrame.info()` are known to work at this point.

See this example below for an illustration

```python
import pandas as pd
import pyam

TEST_DF = pd.DataFrame([
    ['model_a', 'scen_a', 1, 'World', 'Primary Energy', 'EJ/yr', 1, 6.],
    ['model_a', 'scen_a', 2, 'World', 'Primary Energy|Coal', 'EJ/yr', 0.5, 3],
    ['model_a', 'scen_b', 3, 'World', 'Primary Energy', 'EJ/yr', 2, 7],
],
    columns=['source', 'scenario', 'version', 'region', 'variable', 'unit'] + [2005, 2010],
)

df = pyam.IamDataFrame(TEST_DF, index=['source', 'scenario', 'version'])
df.info()
```

This PR also solves an problem with the dynamically created attributes like `d.year` or `df.time`, closes #485

PR #486, opened 2021-01-22
Enable reading non-default scenario versions from IIASA API

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR uses the "custom-index" feature (currently in beta support, see #481) to read data of non-default scenario versions from the IIASA RestAPI and return an IamDataFrame with a non-standard index.


danielhuppmann:
@benke-pik, you may like this new feature... Try `pyam.read_iiasa(<instance>, default=False)` to get not just the latest version of a model-scenario but all of versions uploaded to the IIASA database...

danielhuppmann:
Merging as there were no critical comments or objections...

PR #488, opened 2021-01-27
Read all sheets starting with `data` when reading from xlsx

### LINKED ISSUE #479: Proposal to read in multiple sheets (from Excel)
Planning to extend `pyam.IamDataFrame()` such that if the excel file has multiple sheets, they are each read in and concatenated.

Proposal would be to either:

- accept sheets if starting with "data", e.g. data1, data2, etc.. (could be automatically?)
- accept a`str` or `list` of sheetnames, e.g. `sheetname=...`

Any preferences or other suggestions?

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR allows reading multiple sheets when initializing an IamDataFrame from an xlsx file. It also changes the default such that all sheets starting with 'data' (including a sheet called 'data' ) will be read.

This PR pulls a feature from the ixmp-server processing workflow to the (public) pyam feature list.

closes #479

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -111,7 +111,7 @@ def write_sheet(writer, name, df, index=False):
             pass
 
 
-def read_pandas(path, default_sheet_name='data*', *args, **kwargs):
+def read_pandas(path, sheet_name='data*', *args, **kwargs):
     """Read a file and return a pandas.DataFrame"""
     if isinstance(path, Path) and path.suffix == '.csv':
         df = pd.read_csv(path, *args, **kwargs)
@@ -119,7 +119,7 @@ def read_pandas(path, default_sheet_name='data*', *args, **kwargs):
         xl = pd.ExcelFile(path)
         sheet_names = pd.Series(xl.sheet_names)
         if len(sheet_names) > 1:
-            sheets = kwargs.pop('sheet_name', default_sheet_name)
+            sheets = kwargs.pop('sheet_name', sheet_name)
             # apply pattern-matching for sheet names (use * as wildcard)
             sheets = sheet_names[pattern_match(sheet_names, values=sheets)]
             df = pd.concat([xl.parse(s, *args, **kwargs) for s in sheets])
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -111,15 +111,20 @@ def write_sheet(writer, name, df, index=False):
             pass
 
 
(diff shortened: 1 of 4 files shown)


pyam/utils.py
```diff
@@ -111,15 +111,20 @@ def write_sheet(writer, name, df, index=False):
...
         xl = pd.ExcelFile(path)
-        if len(xl.sheet_names) > 1 and 'sheet_name' not in kwargs:
-            kwargs['sheet_name'] = default_sheet
-        df = pd.read_excel(path, *args, **kwargs)
+        sheet_names = pd.Series(xl.sheet_names)
+        if len(sheet_names) > 1:
+            sheets = kwargs.pop('sheet_name', sheet_name)
+            # apply pattern-matching for sheet names (use * as wildcard)
+            sheets = sheet_names[pattern_match(sheet_names, values=sheets)]
+            df = pd.concat([xl.parse(s, *args, **kwargs) for s in sheets])
```
byersiiasa, line 125: This is pythonic....
in practice, particularly because dealing with many sheets generally means large data and longer load in time, I have done this with a for loop (actually looping pyam.IamDataFrame + pyam.append):
1. print each sheet name to give user feedback on the progress
2.  because if there is an error reading in a sheet, then this helps debug in which sheet a problem is located.
something like:
```
for s in sheets:
    print(f'loading {s}')
    df.append(pyam.IamDataFrame(xl, sheet_name=s),
                                                  inplace=True)
```
or as you have it, something like?
`df = df.append(xl.parse(s, *args, **kwargs))`

danielhuppmann: One reason that I implemented it in this way is that `IamDataFrame.append()` is not very efficient - for every append, it checks whether there are any duplicate entries in the timeseries data. So first doing a concatenation in pandas and only then casting it to an IamDataFrame (and doing the validation for duplicates) should be much faster.

If you can run a test and feel that the additional logging is still helpful, maybe
```
lst = []
for s in sheets:
    logger.info(f'Reading sheet {s}')
    lst.append(xl.parse(s, *args, **kwargs))
df = pd.concat(lst)
```

byersiiasa: I suspected it was speed - and it is actually in the `pyam` checks where the good checks are done.
Anyways - great, I'll add this suggestion above.

byersiiasa:
Did some tests with multiple sheets - works great 
many thanks

Issue #490, opened 2021-02-03
Allow loading 'meta' without `exclude` column


danielhuppmann:
Currently, loading a 'meta' table from file (e.g., as part of initializing an IamDataFrame from xlsx) requires that the column `exclude` exists in the table. Per this comment (https://github.com/iiasa/ariadne-intern-workflow/pull/2#pullrequestreview-581508321) by @fbenke-pik, this is cumbersome as part of the IIASA-Scenario-Explorer submission procedure. Also, there seem to be some issues when the file is empty or doesn't have any entries beyond the headers.

danielhuppmann:
closed via #491

PR #491, opened 2021-02-04
Allow loading meta from file without exclude columns or with empty rows

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR makes the `load_meta()` function (which is also part of initializing an IamDataFrame from an xlsx file) less strict, such that it does not require an `exclude` column and passes silently if only headers are given in the `meta` sheet.

solves #490


danielhuppmann:
@fbenke-pik, this should solve your request for more convenient upload to the IIASA Scenario Explorer infrastructure - please let me know if there are other use cases that are still causing problems

danielhuppmann:
closing and reopening to re-execute the tests and see if Codecov now is included in the tests, see #493

fbenke-pik (approved):

PR #493, opened 2021-02-04
Fix stale Codecov integration

danielhuppmann:
# Description of PR

For some reason, reports were not uploaded to Codecov for a few months. This PR resets the codecov-GitHub-Actions module version to 1, which seems to fix that issue.


danielhuppmann:
Seems that Codecov and Github are confused whether to show Codecov checks... Will merge for now and have an eye on the behavior going forward...

PR #494, opened 2021-02-07
Small performance improvements

coroa:

- ~Tests Added~
- ~Documentation Added~

# Description of PR

Small performance improvements:

1. Unit conversion: by using `loc` we are actually manipulating the copy of the data inplace instead of creating a new full copy of the unit column.
2. `__len__`: We use the length of the `_data`-Series instead of creating the temporary long dataframe.
3. Use numpy indexing for re-coding in `replace_index_values`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,7 @@ starting with `data` will be parsed for timeseries data.
 
 ## Individual updates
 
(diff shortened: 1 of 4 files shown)


danielhuppmann:
This looks great, thanks @coroa! Do you want to add this to the release notes, or do you think it's not necessary?

danielhuppmann (approved):
Thanks @coroa

PR #497, opened 2021-02-08
Read data from the UNFCCC Data Inventory

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

## Overview

While playing around with the really nice **unfccc-di-api** package ([read the docs](https://unfccc-di-api.readthedocs.io/en/latest/) by @mikapfl, I ran into a number of issues that I found unintuitive (why CO₂) and difficult to wrap my head around. So I ended up writing up quite some code to get the UNFCCC-data into a format that I feel comfortable with... and knowing only a hammer, I turned this problem into a nail. Sorry.

## Rendered docs

The new function is rendered [here](https://pyam-iamc.readthedocs.io/en/reader-unfccc/api/database.html#reading-unfccc-inventory-data) and the tutorial is rendered [here](https://pyam-iamc.readthedocs.io/en/reader-unfccc/tutorials/unfccc.html).

## Main features

 - get rid of all the subscripts both in the API (kwarg `gases`) and the returned data (so annoying)
 - add the species to the units for compatibility with the [iam-units](https://github.com/IAMconsortium/units) package
 - implement a dictionary to translate the four identifier-columns `['category', 'classification', 'measure', 'gas']` into an IAMC-style variable
 - add a "tier" argument to make it easy to extract the level of detail that is necessary (implementing a "tier 1" as a first step)

## Caveats

The implementation follows the lesson I learned from @gidden: do something that works first, and then worry about how to make it fast/efficient/elegant. Apologies for any very bad code. Additional tiers should be added as follow-up PRs, the current implementation just shows how I think the API could make it easy for energy-systems modellers and return something that is close to scenario ensembles used by the IPCC and IAM model comparison exercises.

Any comments are welcome!


danielhuppmann:
@mikapfl, if you think some of the harmonization features would also be relevant/helpful for the unfccc-di-api package, I'd be happy to migrate the suggested changes to the other repo!

mikapfl:
I definitely see the general value of the normalization of gas names without super- and subscripts, I think *nobody* wants non-ascii gas names. Maybe add a (by default True) argument to `query`, something like `normalize_gas_names`? I'd be happy to accept patches for that. (-:

danielhuppmann:
Moved most of the formatting-wrangling to the **unfccc-di-api** package per invitation by @mikapfl, added an explicit dependency version requirement.

danielhuppmann:
As there were no comments or objections, merging to use some of the new utils-features in another branch.

PR #503, opened 2021-03-02
Fix timeseries regressions

coroa:

- [x] Tests Added
- ~Documentation Added~ -> no changes
- ~Description in RELEASE_NOTES.md Added~ -> I'd say not noteworthy enough! objections?

# Description of PR

1. Re-implement `iamc_index` argument for dropping `extra_cols`

2. Guarantee sort order in time axis
    
   During the `_data` refactor, the `timeseries` method was refactored from `pivot_table` to a simple `unstack`. when earlier (as in sortorder) models, scenario, variables have their first year after later ones, the time axis will be out-of-order.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -707,7 +707,10 @@ class IamDataFrame(object):
         if self.empty:
             raise ValueError("This IamDataFrame is empty!")
 
-        df = self._data.unstack(level=self.time_col).rename_axis(None, axis=1)
+        s = self._data
+        if iamc_index:
+            s = s.droplevel(self.extra_cols)
+        df = s.unstack(level=self.time_col).rename_axis(None, axis=1).sort_index(axis=1)
 
         if df.index.has_duplicates:
             raise ValueError(
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -693,7 +693,7 @@ class IamDataFrame(object):
 
         Parameters
         ----------
-        iamc_index : bool, default False
+        iamc_index : bool, optional
             if True, use `['model', 'scenario', 'region', 'variable', 'unit']`;
             else, use all 'data' columns
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -714,12 +714,9 @@ class IamDataFrame(object):
                     "Cannot use IAMC-index with continuous-time data format!"
                 )
             s = s.droplevel(self.extra_cols)
+
         df = s.unstack(level=self.time_col).rename_axis(None, axis=1).sort_index(axis=1)
 
-        if df.index.has_duplicates:
-            raise ValueError(
-                "Data with IAMC-index has duplicated index, use `iamc_index=False`"
-            )
         return df
 
     def reset_exclude(self):
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -709,6 +709,10 @@ class IamDataFrame(object):
 
         s = self._data
         if iamc_index:
+            if self.time_col == "time":
+                raise ValueError(
+                    "Cannot use IAMC-index with continuous-time data format!"
+                )
             s = s.droplevel(self.extra_cols)
         df = s.unstack(level=self.time_col).rename_axis(None, axis=1).sort_index(axis=1)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -145,6 +145,16 @@ def test_df_year():
     yield df
 
 
+# minimal IamDataFrame for specifically testing 'time'-column features
+@pytest.fixture(scope="function")
+def test_df_time():
+    tdf = TEST_DF.rename({2005: TEST_DTS[0], 2010: TEST_DTS[1]}, axis="columns")
+    df = IamDataFrame(data=tdf)
+    for i in META_COLS:
+        df.set_meta(META_DF[i])
+    yield df
+
+
 # minimal test data as pandas.DataFrame (only 'year' time format)
 @pytest.fixture(scope="function")
 def test_pd_df():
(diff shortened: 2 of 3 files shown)


coroa:
Black will be fine with that line! I'd suggest waiting for the black changes to land. I'll rebase afterwards, same for other branches.

pyam/core.py
```diff
@@ -686,7 +686,10 @@ def timeseries(self, iamc_index=False):
         if self.empty:
             raise ValueError('this `IamDataFrame` is empty')
 
-        df = self._data.unstack(level=self.time_col).rename_axis(None, axis=1)
+        s = self._data
+        if iamc_index:
```
danielhuppmann, line 690: Note that an IamDataFrame could also use `time` instead of `year` or a custom index like `['model', 'scenario', 'version']`- in this case, just dropping the extra-columns will not yield the expected IAMC-index.

danielhuppmann (changes requested):
Thanks for the catching that I dropped this feature during the refactoring...

Food for thought inline...

pyam/core.py
coroa: I have not used sub-yearly dataframes yet. Could you point me at an example? And how is the iamc_index defined in that case?

danielhuppmann: Unfortunately, we don't have any decent documentation yet for these formats...

1. continuous-time format: use a column 'time' (instead of 'year') with format `datetime`, e.g.

    | **model** | **scenario** | **region** | **variable**   | **unit** | **time** | **value** |
    |-----------|--------------|------------|----------------|----------|----------|----------|
    | MESSAGE   | CD-LINKS 400 | World      | Primary Energy | EJ/y     |    2020-03-09 13:10 |    500.7 |
    | ...       | ...          | ...        | ...            | ...      |      ... |      ... |      ... |

    Many features are tested for this format, see https://github.com/IAMconsortium/pyam/blob/7bcedfaf5cec135ab0f9cb58c310cb694ddf6c7f/tests/conftest.py#L129

    This should raise a `ValueError("Cannot use iamc-index with continuous-time data format!")`.

2. subannual-column: for a time representation that can't be handled as continuous-time (e.g., representative time-steps like `summer-day`), we use a 'subannual' column.

    | **model** | **scenario** | **region** | **variable**   | **unit** | **subannual** | **2005** | **2010** | **2015** |
    |-----------|--------------|------------|----------------|----------|---------------|----------|----------|----------|
    | MESSAGE   | CD-LINKS 400 | World      | Primary Energy | EJ/y     |  winter-day |  462.5 |    500.7 |      ... |
    | ...       | ...          | ...        | ...            | ...      |      ... |      ... |      ... |

   In this case (ie., df.time_col == "year"), `subannual` will just be dropped as part of `drop level(self.extra_cols
)`

pyam/core.py
```diff
@@ -707,7 +707,10 @@ def timeseries(self, iamc_index=False):
         if self.empty:
             raise ValueError("This IamDataFrame is empty!")
 
-        df = self._data.unstack(level=self.time_col).rename_axis(None, axis=1)
+        s = self._data
+        if iamc_index:
+            s = s.droplevel(self.extra_cols)
```
danielhuppmann, line 712:
```suggestion
            if self.time_col != "year":
                 raise ValueError("Cannot use `iamc_index=True` with continuous-time data format!")
            s = s.droplevel(self.extra_cols)
```

pyam/core.py
coroa: Thanks!

Regarding 2:
> In this case (ie., df.time_col == "year"), subannual will just be dropped as part of drop level(self.extra_cols )

I guess that is also the desired behaviour, is it? It's quite likely to run into the "Data with IAMC-index has duplicates"  `ValueError`.

pyam/core.py
danielhuppmann: No strong preference...

pyam/core.py
danielhuppmann: Yes, this was/is intentional... The rationale at the time: we may introduce fancy methods that add extra-columns, like open-scm adding a column "climate_model" to track which model was used for the temperature assessment. Then, it would be helpful for users to have an easy method to get rid of the extra-columns, for example to pass the timeseries data another script or tool that doesn't have the flexibility of pyam and can only handle a "strict" IAMC format.

If that works by simple dropping the extra-col (using `iamc_index=True`, fine - otherwise, the user has to come up with something smarter...

I'd keep that option for now (even if only mildly useful), and maybe we add a `drop()´ at a later point to replace it...

pyam/core.py
```diff
@@ -707,7 +707,14 @@ def timeseries(self, iamc_index=False):
...
+        s = self._data
+        if iamc_index:
+            if self.time_col == "time":
+                raise ValueError(
+                    "Cannot use IAMC-index with continuous-time data format!"
+                )
+            s = s.droplevel(self.extra_cols)
+        df = s.unstack(level=self.time_col).rename_axis(None, axis=1).sort_index(axis=1)
 
         if df.index.has_duplicates:
```
danielhuppmann, line 719: Based on your earlier comment that checking for duplicates is expensive, and if we assume that the `self_data` has a unique index to begin with, this check only needs to happen if `iamc_index=True`...

coroa: It is expensive the first time you do it, but it is also cached! So you should mainly avoid doing it in a loop! Since we typically verified it after construction, it's basically free here:

image

coroa: Also, easy to test: Result -> Nothing to worry about.

image

danielhuppmann: got it, thanks!

pyam/core.py
coroa: Fine!

danielhuppmann (approved):
Thanks @coroa for adding the sort and reimplementing the iamc_index-feature!

PR #507, opened 2021-03-15
Refactor aggregation-functions to avoid `data`

danielhuppmann:

- ~Tests Added~ -> no functional change
- ~Documentation Added~ -> no functional change

# Description of PR

This PR removes two cases where the full `data` dataframe was created unnecessarily as part of the region-aggregation, refactoring to work directly on the `pd.Series.index` and related functions. Should speed up some computations a little...

It also harmonizes the log and error messages to be more like pandas - start with a capital letter and use f-strings.


pyam/core.py
```diff
@@ -1503,15 +1502,15 @@ def downscale_region(
     def _all_other_regions(self, region, variable=None):
         """Return list of regions other than `region` containing `variable`"""
         rows = self._apply_filters(variable=variable)
-        return set(self.data[rows].region) - set([region])
+        return set(self._data[rows].index.get_level_values("region")) - set([region])
```
coroa, line 1505:
```suggestion
        return self._data[rows].index.get_level_values("region").difference([region])
```

No need for `set`

coroa (commented):
`recursive_aggregate` as well as `aggregate_region` still use the `data` attribute. Is that intentional?

danielhuppmann:
> `recursive_aggregate` as well as `aggregate_region` still use the `data` attribute. Is that intentional?

Now that you mention it... Replaced it is in a number of places. But `recursive_aggregate()` is a bit tricky, not sure how good the unit tests for that function are (found something that should have broken it but didn't). So I'm a bit hesitant to touch that as part of this PR.

danielhuppmann:
The PR was extended to also refactor the recursive-aggregation feature.

Per a bilateral comment by @tburandt, I also harmonized the return type of the `aggregate()` and `aggregate_region()` methods when performing "empty" aggregation, i.e., no components exist to perform the requested aggregation.

In the new implementation, an empty IamDataFrame is returned if `append=False` (instead of None).

pyam/core.py
```diff
@@ -1303,16 +1313,11 @@ def aggregate_region(
             weight=weight,
         )
 
-        # return None if there is nothing to aggregate
-        if _df is None:
-            return None
-
         # else, append to `self` or return as `IamDataFrame`
         if append is True:
```
coroa, line 1317: Why are you explicitly interpreting append=1 as append=False?

```suggestion
        if append:
```

pyam/utils.py
```diff
@@ -441,7 +444,7 @@ def _count_pipes(val):
     else:
         raise ValueError("Unknown level type: `{}`".format(level))
 
-    return list(map(test, n_pipes))
+    return list(map(test, n_pipes))[0] if unique else list(map(test, n_pipes))
```
coroa, line 447: I find 

```python
def _find_depth(data, s, level):
    # [...] previous implementation for lists only

def find_depth(data, s="", level=None):
    if isinstance(data, str):
        return _find_depth([data], s, level)[0]
    
     return _find_depth(data, s, level)
```

easier to understand

pyam/core.py
```diff
@@ -1158,46 +1164,50 @@ def normalize(self, inplace=False, **kwargs):
...
+            _df = IamDataFrame(_aggregate_recursive(self, variable), meta=self.meta)
         else:
             _df = _aggregate(self, variable, components=components, method=method)
 
-        # return None if there is nothing to aggregate
-        if _df is None:
-            return None
-
         # else, append to `self` or return as `IamDataFrame`
         if append is True:
```
coroa, line 1206:
```suggestion
        if append:
```

pyam/_aggregate.py
```diff
@@ -42,76 +52,62 @@ def _aggregate(df, variable, components=None, method=np.sum):
...
-            _df_aggregated = _df_temp.copy()
-        else:
-            _df_aggregated.append(_df_temp, inplace=True)
+        # a temporary dataframe allows to distinguish between full data and new data
+        temp_df = _df.aggregate(variable=var_list)
+        _df.append(temp_df, inplace=True)
+        data_list.append(temp_df._data)
 
-    return _df_aggregated.data
+    return pd.concat(data_list)
```
coroa, line 77: I'm pretty sure that re-implementation fixes also bugs that I had with recursive aggregation previously where it would sometimes only perform the lowest level aggregation.

pyam/_aggregate.py
```diff
@@ -42,76 +52,62 @@ def _aggregate(df, variable, components=None, method=np.sum):
...
-        _data = _group_and_agg(subregion_df.data[rows], col, method=method)
+        _data = _group_and_agg(subregion_df._data[rows], "region", method=method)
     else:
         weight_rows = subregion_df._apply_filters(variable=weight)
         _data = _agg_weight(
-            subregion_df.data[rows], subregion_df.data[weight_rows], method
+            subregion_df._data[rows], subregion_df._data[weight_rows], method
         )
 
     # if not `components=False`, add components at the `region` level
```
coroa, line 113: I'm unsure what the correct behaviour for `components=[]` would be, here it is treated the same as `components=False`. If that is correct i'd suggest using `if components:` instead of the `if components is not False`.

coroa (approved):
Looks like quite an improvement, great! Found only a few minor stylistic things, feel free to judge yourself!

pyam/_aggregate.py
coroa: In the untouched lines below this snippet :/

https://github.com/IAMconsortium/pyam/blob/8ca67945d55928cbc615b624a51dfec20079acd4/pyam/_aggregate.py#L114-L132

danielhuppmann: Agree, if you explicitly pass an empty list of components, that should be equivalent to setting components to False...

danielhuppmann: Changed as suggested

pyam/utils.py
danielhuppmann: Great suggestion, implemented!

danielhuppmann:
Thanks @coroa for the helpful review!

PR #510, opened 2021-03-21
Improve performance of `pyam.concat()`

### LINKED ISSUE #500: concat uses pandas append instead of concat
The implementation of pyams concat method loops over pandas append instead of using pandas concat method. Especially for large dataframes with subannual data concat is much faster than append.

I already implemented pandas concat in pyams concat and will push it in a minute. I hope the new feature will find its way in the next release.

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR refactors the implementation of `pyam.concat()` to use `pd.concat()` on the timeseries data rather than iteratively calling `IamDataFrame.append()`.

closes #500

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,3 +17,4 @@ The following persons contributed to the development of the |pyam| framework:
 - Maik Budzinski `@mabudz <https://github.com/mabudz>`_
 - Jarmo Kikstra `@jkikstra <https://github.com/jkikstra>`_
 - Michael Pimmer `@fonfon <https://github.com/fonfon>`_
+- Patrick Jürgens `@pjuergens <https://github.com/pjuergens>`_
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,7 @@ starting with `data` will be parsed for timeseries data.
 
 ## Individual updates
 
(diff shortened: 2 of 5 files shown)


danielhuppmann:
@pjuergens, it would be great if you could try out this implementation for your use cases and report any unexpected issues. Also, it would be interesting to hear about the performance improvements in a real-world application.

pjuergens:
@danielhuppmann thanks for working on this issue. I'll try to test the implementation today.

Concerning performance-improvement in real-world-application: We have a code in which we aggregate a lot of hourly-resolved timeseries. The bottleneck there was the pyam.append-function. Switching from append to concat on major parts of the code with my original implementation I could reduce the calculation time from 24 to 15 minutes. I'll have a deeper look at it again today when I test the new implementation.

pjuergens:
my script ran without any issues, thanks again!

coroa (approved):
LGTM. I can't find any faults in that implementation. Thanks for the nice speed-up!

The next low-hanging fruit on the speed-up quest is to add a (non-copying!?) fast-path to `__init__` for a `pd.Series` in *perfect* condition, ie an index with the right names that is unique and maybe other conditions that one would have to extract from `format_data` to avoid having to reindex.

danielhuppmann:
At some point, it will be so fast that you don't even have time for coffee while your scripts are running 😜...

PR #512, opened 2021-03-29
Fix iterable regression in concat introduced by #510

coroa:
Before #510, `concat` transparently accepted an iterable of `IamDataFrame`s, but the introduction of the subscripts `dfs[0]` and `dfs[1:]` broke this behaviour.

Examples of such iterables would be anything produced by `map(_do_something, dfs)` or `filter`.

~Here we explicitly cast to iterable using `iter` to work with iterables.~

Here we explicitly cast to list for readibility.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2314,7 +2314,7 @@ def concat(dfs, ignore_meta_conflict=False, **kwargs):
 
     Parameters
     ----------
-    dfs : list of IamDataFrames
+    dfs : iterable of IamDataFrames
         A list of :class:`IamDataFrame` instances
     ignore_meta_conflict : bool, default False
         If False, raise an error if any meta columns present in `dfs` are not identical.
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Thanks @coroa - can you give an example of where the current implementation fails? Which iterable that's not a list or a series are you using?

danielhuppmann:
When passing an empty list to the new implementation, this yields

```python
StopIteration: 
```

whereas the previous implementation returns

```python
IndexError: list index out of range
```

In contrast, pandas returns a more intelligible

```python
ValueError: No objects to concatenate
```

But it seems that the object returned by `iter()` does not have a `len()` attribute, so adding a straightforward test for length seems not possible with this approach. Maybe casting to `list` is smarter? [It's also what pandas does...]

coroa:
Sorry for the delays, I was on holidays and avoided touching my computer as much as possible.

> Thanks @coroa - can you give an example of where the current implementation fails? Which iterable that's not a list or a series are you using?

Well, in my case it was just some itertools foo, ie does not matter too much.

pyam/core.py
```diff
@@ -2340,15 +2340,20 @@ def concat(dfs, ignore_meta_conflict=False, **kwargs):
...
 
-    # cast first element in list to IamDataFrame (if necessary)
-    df = dfs[0] if isinstance(dfs[0], IamDataFrame) else IamDataFrame(dfs[0], **kwargs)
+    dfs_iter = iter(dfs)
+
+    # cast to IamDataFrame if necessary
+    def as_iamdataframe(df):
+        return df if isinstance(df, IamDataFrame) else IamDataFrame(df, **kwargs)
+
+    df = as_iamdataframe(next(dfs_iter))
```
coroa, line 2349: If one wants to catch the StopIteration, the easiest is:

```suggestion
    df = next(dfs_iter, None)
    if df is None:
        raise ValueError("No objects to concatenate")
    df = as_iamdataframe(df)
```

but I guess that becomes more and more difficult to understand without any clear gain on the other side, let me rewrite it back to list.

pyam/core.py
```diff
@@ -2314,7 +2314,7 @@ def concat(dfs, ignore_meta_conflict=False, **kwargs):
 
     Parameters
     ----------
-    dfs : list of IamDataFrames
+    dfs : iterable of IamDataFrames
         A list of :class:`IamDataFrame` instances
```
danielhuppmann, line 2318: Any object castable as IamDataFrame to be appended
```suggestion
        A list of objects castable to :class:`IamDataFrame`
```

danielhuppmann (approved):
One minor suggestion to extend the docstring in an untouched (but related) line, otherwise looks great!

pyam/core.py
danielhuppmann: thanks!

PR #520, opened 2021-04-25
Allow numpy int in interpolate

coroa:

- [x] Tests Added
- [ ] ~Documentation Added~
- [ ] ~Description in RELEASE_NOTES.md Added~

# Description of PR

Numpy has its own integer types. Here we added `np.integer` as allowed type to several integerness tests especially for years.

Enables the use of helpful constructs like:
```
df.interpolate(np.r_[2020:2030, 2030:2100:5])
```

Also removed list generation of an `any` test, since it can short-circuit that way, ie consider:
```
def is_four(i):
    print(i, end=" ")
    return i == 4

print(any(is_four(i) for i in range(10)))

print(any([is_four(i) for i in range(10)]))
```
which outputs
```
0 1 2 3 4 True
0 1 2 3 4 5 6 7 8 9 True
```

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -744,6 +744,13 @@ def test_interpolate_with_list(test_df_year):
     npt.assert_allclose(obs, [3, 4, 1.5, 2, 4, 5])
 
 
+def test_interpolate_with_numpy_list(test_df_year):
+    test_df_year.interpolate(np.r_[2007 : 2008 + 1], inplace=True)
+    obs = test_df_year.filter(year=[2007, 2008])._data.reset_index(drop=True)
+    exp = pd.Series([3, 4, 1.5, 2, 4, 5], name="value")
+    pd.testing.assert_series_equal(obs, exp)
+
+
 def test_interpolate_full_example():
     cols = ["model_a", "scen_a", "World"]
     df = IamDataFrame(
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -746,9 +746,8 @@ def test_interpolate_with_list(test_df_year):
 
 def test_interpolate_with_numpy_list(test_df_year):
     test_df_year.interpolate(np.r_[2007 : 2008 + 1], inplace=True)
-    obs = test_df_year.filter(year=[2007, 2008])._data.reset_index(drop=True)
-    exp = pd.Series([3, 4, 1.5, 2, 4, 5], name="value")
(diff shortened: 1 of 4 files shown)


coroa (commented):
Stickler E203 seems to be in conflict with black!

danielhuppmann:
Thanks for the improvement, @coroa - but how about a utility function `is_int()` similar `isstr()` and `is scalar()`?

(I'd add the underscore for readability and I guess we should refactor the other functions, too)...

znicholls:
https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_integer_dtype.html might be helpful as a solution which doesn't require your maintenance

coroa:
> https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_integer_dtype.html might be helpful as a solution which doesn't require your maintenance

sounds like a wonderful suggestion.

coroa:
@danielhuppmann Where i would have used `is_int` previously, is now more sanely handled by `to_list`, and the only time i need it in interpolate, we can just as well defer to `pd.api.types.is_integer`.

Ready to merge from my side.

tests/test_core.py
```diff
@@ -744,6 +744,13 @@ def test_interpolate_with_list(test_df_year):
     npt.assert_allclose(obs, [3, 4, 1.5, 2, 4, 5])
 
 
+def test_interpolate_with_numpy_list(test_df_year):
+    test_df_year.interpolate(np.r_[2007 : 2008 + 1], inplace=True)
+    obs = test_df_year.filter(year=[2007, 2008])._data.reset_index(drop=True)
+    exp = pd.Series([3, 4, 1.5, 2, 4, 5], name="value")
+    pd.testing.assert_series_equal(obs, exp)
```
danielhuppmann, line 751: The other tests in that section use the following, any particular reason not to follow that convention?
```suggestion
    obs = test_df_year.filter(year=[2007, 2008])._data.values
    npt.assert_allclose(obs, [3, 4, 1.5, 2, 4, 5])
```

coroa:
Sorry, it seems i did not see your review! Applied your suggestion. If tests still complete successfully, it is good to go.

danielhuppmann:
Thanks @coroa - this PR will generate a number of merge-conflict with a meta-PR #598, so I cherry-picked your changes and created a new PR pointing to the branch of PR #598.

Closing in favor of #622

PR #523, opened 2021-05-04
Add feature to compute weights for de-biasing using count

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Based on a rekindling of the discussion on de-biasing of scenario ensembles by @Rlamboll, this is an initial implementation using the most basic approach to compute de-biasing weights: counting the number of model or scenario names. My hope is that having this initial PR can help to structure the discussion on provide a template to think about more advanced methods for debiasing (e.g., based on generic meta-indicators like projects, differences between certain variables, etc.)

Intended results
----------------

In the test, there is a simple IamDataFrame with three scenarios (i.e., scenario runs) from two models, where `model_a` implements scenarios (in the sense of a scenario protocol) `scen_a` and `scen_b` and `model_b` implements only `scen_a`. The method counts the number of occurrences by name on the `axis` (in the test case: "scenario"). The expected output is

|  model    | scenario | bias |
|----------|--------|-------|
|  model_a | scen_a |  0.5   |
|  model_b | scen_a |  0.5  |
|  model_a | scen_b |  1     |

Note: setting `axis=["model", "scenario"]` returns a bias-vector of 1 in a standard IAMC-template IamDataFrame.

Open questions
----------------

- ~Should the computed bias weights be inverted? i.e., `0.5, 0.5, 1`~
  -> Per discussion below: yes, changed from initial implementation
- Does the function signature make sense?
- The function directly adds the bias-weights to `df.meta` so that they can be retrieved with `df[<name>]`.
  An alternative implementation could be:
  The function `compute_bias()` returns the weights as pd.Series, so that a used would have to add it to `meta` explicitly
  e.g., `df.set_meta(meta=df.compute_bias(...), name="bias")`

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -977,7 +977,31 @@ class IamDataFrame(object):
 
         The following methods are implemented:
 
-        - "count": use the inverse of the number of scenarios grouped by `axis` names
+        - "count": use the inverse of the number of scenarios grouped by `axis` names.
+
+          Using the following method on an IamDataFrame with three scenarios
+
+          .. code-block:: python
+
+              df.compute_bias(name="bias-weight", method="count", axis="scenario")
+
+          results in the following column to be added to *df.meta*:
+
+          .. list-table::
+             :header-rows: 1
+
+             * - model
+               - scenario
+               - bias-weight
+             * - model_a
+               - scen_a
+               - 0.5
+             * - model_a
+               - scen_b
+               - 1
+             * - model_b
+               - scen_a
+               - 0.5
 
         """
         _compute_bias(self, name, method, axis)
@@ -1088,7 +1112,7 @@ class IamDataFrame(object):
     def convert_unit(
         self, current, to, factor=None, registry=None, context=None, inplace=False
     ):
-        r"""Convert all data having *current* units to new units.
+        """Convert all timeseries data having *current* units to new units.
 
         If *factor* is given, existing values are multiplied by it, and the
         *to* units are assigned to the 'unit' column.
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -11,6 +11,8 @@ import pandas as pd
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
+from pyam._debiasing import _compute_bias
+
 try:
     from datapackage import Package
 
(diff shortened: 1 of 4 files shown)


danielhuppmann:
cc @brutschki, @gidden

Rlamboll:
In conversation with @gidden we have constructed something that would take a similar input here: [https://github.com/GranthamImperial/silicone/pull/127](https://github.com/GranthamImperial/silicone/pull/127) which used a dictionary and has the weighting factor rather than the number of scenarios - I feel the latter point is always better because it will be less confusing with non-integer biases, should one have reasons for weighting scenarios based on, say, historic performance accuracy or a continuous degree-of-overlap metric. 

I vaguely plan to allow DataFrame-style inputs at some point but the dictionary {(model, scenario): weight} makes more sense when only a few model/scenarios have weights other than 1 and people are hard-coding them, if there's a pyam tool to generate this table then maybe the increased readability wins out.

danielhuppmann:
Thanks for pointing to the PR, @Rlamboll - if I understand your statement and the code in silicone correctly, you think this function should return the inverse of the count? (i.e. the weighting factor).

Rlamboll:
Yes, in this case you'd return weight=1/count.

pyam/core.py
```diff
@@ -957,6 +959,28 @@ def validate(self, criteria={}, exclude_on_fail=False):
...
+        """Compute the bias weights and add to 'meta'
+
+        Parameters
+        ----------
+        name : str
+           Column name in the 'meta' dataframe
+        method : str
+            Method to compute the bias weights, see the notes
+        axis : str
+            Index dimensions on which to apply the `method`
```
Rlamboll, line 972: I think this documentation is not the easiest to follow - as it's currently defined (without wildcard matching) only "model" or "scenario" are sensible choices of axis, so maybe giving an example would be useful.

Rlamboll (approved):

pyam/core.py
danielhuppmann: Added the example from the description of this PR to the notes section of this method

Rlamboll (approved):

Rlamboll:
Looks good to me!

Issue #355, opened 2020-03-29
Bulk aggregate


znicholls:
In https://github.com/znicholls/silicone/pull/72, @Rlamboll has added a feature to do bulk aggregation within an `IamDataFrame`. This is a convenience function, but means you could quickly specify to calculate a bunch of aggregate variables (some of which are more complex than just a pure sum) without writing a custom wrapper every time (current implementation [here](https://github.com/znicholls/silicone/pull/72/files#diff-aff65db8e0cdd377901a82592698e061L7), work in progress though). 

@danielhuppmann Does this feature already exist? If no, is it something you'd be interested in bringing across?

danielhuppmann:
Thanks @znicholls for the cross-reference. This is similar to the recent improvements of [aggregate()](https://pyam-iamc.readthedocs.io/en/latest/api/iamdataframe.html#pyam.IamDataFrame.aggregate) (#305 & #312). This now supports a `method` arg (min, max, weighted sum) and doing "bulk" aggregation by passing a list of variables to be aggregated. Similar for [aggregate_region()](https://pyam-iamc.readthedocs.io/en/latest/api/iamdataframe.html#pyam.IamDataFrame.aggregate_region), the related `check_*()` functions and the [downscale_region()](https://pyam-iamc.readthedocs.io/en/latest/api/iamdataframe.html#pyam.IamDataFrame.downscale_region) function.

The current implementation in pyam, though, only works with a list of variables when you do the "obvious" aggregation, i.e., all subcategories of each variable - whereas the implementation in silicone takes a `dict(variable=components)`. This would be a useful addition to pyam, I think.

As for the second feature, I have some concerns that this would be overloading the function. If the multiplication relates to conversion to CO2-equivalents, I'd rather use separate function for conversion and aggregation, for example:
```
_df = df.convert_unit('Mt CH4', 'Mt CO2e', context='gap_AR5GWP100')
_df.aggregate('Kyoto GHG', components=['Emissions|CO2', 'Emissions|CH4'])
```
Or the first step could use the dataframe-operations feature (work in progress by @gidden, see #333)...

Rlamboll:
The main use of the factors is to do subtraction with -1, e.g. 
`aggregate = "Emissions|CO2|Other"`
`other_CO2 = mi.infill_composite_values(
    sr15_data,
    {
        aggregate: {
            "Emissions|CO2": 1,
            "Emissions|CO2|Energy and Industrial Processes": -1,
[...]

PR #527, opened 2021-05-07
Add an in-dataframe subtraction operation

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR implements a first version of an in-IamDataFrame subtraction feature, e.g., to compute the difference between total primary energy and the components (and optionally add to the IamDataFrame).

```python
df.subtract("Primary Energy", "Primary Energy|Coal", "Primary Energy|Other", append=True)
```

This is just the most basic approach - it assumes that `a - b` is not computed for a specific row if either `a` or `b` is not defined.

# Future improvements and extensions

 - Allow setting a default value if `a` or `b` is not defined
 - Enable setting `b` as a numerical value (e.g., `Primary Energy * -1`)
 - Add support for across-IamDataFrame operations, e.g.
    ```python
    pyam.subtract(a, b, ...)
    ```
    where `a` and `b` are IamDataFrame-instances.
 - Add similar operations for addition, multiplication, shares...

### DIFF
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -5,7 +5,24 @@ def subtract(a, b):
     return a - b
 
 
-KNOWN_OPS = {"subtract": subtract}
+def add(a, b):
+    return a + b
+
+
+def divide(a, b):
+    return a / b
+
+
+def multiply(a, b):
+    return a * b
+
+
+KNOWN_OPS = {
+    "subtract": subtract,
+    "add": add,
+    "divide": divide,
+    "multiply": multiply,
+}
 
 
 def _op_data(df, a, b, name, method, axis):
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -5,9 +5,7 @@ def subtract(a, b):
     return a - b
 
 
-KNOWN_OPS = {
-    "subtract": subtract
-}
+KNOWN_OPS = {"subtract": subtract}
 
 
 def _op_data(df, a, b, name, method, axis):
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -1,27 +1,12 @@
+import operator
 from pyam.index import append_index_level
 
 
-def add(a, b):
-    return a + b
-
-
-def subtract(a, b):
-    return a - b
-
-
-def multiply(a, b):
-    return a * b
-
-
-def divide(a, b):
-    return a / b
-
-
 KNOWN_OPS = {
-    "subtract": subtract,
-    "add": add,
-    "divide": divide,
-    "multiply": multiply,
+    "add": operator.add,
+    "subtract": operator.sub,
+    "multiply": operator.mul,
+    "divide": operator.truediv,
 }
 
 
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -13,14 +13,17 @@ KNOWN_OPS = {
 def _op_data(df, a, b, name, method, axis):
     """Internal implementation of numerical operations on timeseries"""
 
-    cols = df._data.index.names.difference([axis])
-    _a = df.filter(**{axis: a})._data.groupby(cols).sum()
-    _b = df.filter(**{axis: b})._data.groupby(cols).sum()
+    if axis not in df._data.index.names:
+        raise ValueError(f"Unknown axis: {axis}")
 
     if method in KNOWN_OPS:
         method = KNOWN_OPS[method]
     else:
         raise ValueError(f"Unknown method: {method}")
+
+    cols = df._data.index.names.difference([axis])
+    _a = df.filter(**{axis: a})._data.groupby(cols).sum()
+    _b = df.filter(**{axis: b})._data.groupby(cols).sum()
 
     _value = method(_a, _b)
     _value.index = append_index_level(_value.index, 0, name, axis)
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -10,7 +10,7 @@ KNOWN_OPS = {
 }
 
 
-def _op_data(df, a, b, name, method, axis):
+def _op_data(df, method_args, name, method, axis):
     """Internal implementation of numerical operations on timeseries"""
 
     if axis not in df._data.index.names:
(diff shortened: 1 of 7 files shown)


danielhuppmann:
@pjuergens @gaurav-ganti, any comments?

Rlamboll:
In this thread #355 we discuss importing a feature from Silicone that allows construction of any aggregate value using positive or negative multipliers, which would immediately allow this and also other operations when constructing more complex values (e.g. X-Y-Z). This looks like what you're proposing in your extension work.

pjuergens:
Thanks for this approach!

I have some proposals to make the _op_data-function more flexible:

- for the keyword method allow to pass a self-defined function as well, i.e.
```python
if method in KNOWN_OPS:
    method = KNOWN_OPS[method]
elif callable(method):
    pass
else:
    raise ValueError(f"Unknown method: {method}")
```
- instead of passing the variables a and b as single parameters, use a list [a, b, ...] to be able to pass more than 2 variables. _a, _b would then need to be put in a list as well and passed to method as *args
- the _op_data-function could then be used in a calculate(self, variables, name, method, axis="variable", append=False)-function, where I can define my own method

pyam/_ops.py
```diff
@@ -0,0 +1,29 @@
...
+        method = KNOWN_OPS[method]
+    else:
+        raise ValueError(f"Unknown method: {method}")
+
+    cols = df._data.index.names.difference([axis])
+    _a = df.filter(**{axis: a})._data.groupby(cols).sum()
+    _b = df.filter(**{axis: b})._data.groupby(cols).sum()
+
+    _value = method(_a, _b)
+    _value.index = append_index_level(_value.index, 0, name, axis)
```
gidden, line 27: as an 'intermediate' developer, I would find explicitly using argument names useful to understand how this works, like

```
_value.index = append_index_level(_value.index, codes=0, level=name, name=axis)
```

danielhuppmann: Good suggestion, thanks!

byersiiasa:
lots of complicated things going on  - this iwll be very useful. I guess you are already thinking other opeartors too.

a while a go I had been converting to `.as_pandas()` and then using the operator package. Works well

```    import operator
    ops = { "+": operator.add, 
           "-": operator.sub,
           "/": operator.truediv,
           "*": operator.mul}```

**Comment by danielhuppmann** [comment]
Thanks for the fast feedback!

- @gidden, suggestion implemented, thanks!
 - @Rlamboll, thanks for bringing that discussion back to the radar, I added operations with numbers to the list of future ideas (maybe I'll move that section to an own issue to list all future developments)
 - @pjuergens, I was hoping you'd have these kind of ideas! I suggest that I'll let this PR sit over the weekend (in case of more comments or for a more thorough review), then merge, and then you can implement these improvements.

One suggestion: instead of `IamDataFrame.calculate()`, why not mimic pandas and do an `IamDataFrame.apply()` function? (see [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html))

**Comment by gidden** [comment]
@danielhuppmann and @pjuergens would you welcome additional ops PR'd into this branch if people have time to include tests for them?

**Comment by danielhuppmann** [comment]
> @danielhuppmann and @pjuergens would you welcome additional ops PR'd into this branch if people have time to include tests for them?

I'd prefer to have a large number of small PRs rather than one ops-branch that diverges from the main branch over weeks. So if there are small additions or change that fit in the scope of a first "proof-of-concept" PR, great!

But let's merge this PR before tackling big chunks of additional features.

**Comment by gidden** [comment]
added other minimal ops if they seem fit for inclusion in https://github.com/danielhuppmann/pyam/pull/29

and in any case, don't we want to add major new features in a single squashed commit anyway (i.e., developed in a feature branch before being merged into `main`?)

**Comment by danielhuppmann** [comment]
Thanks @gidden for adding the other basic operations as a PR! I added a second test for each, refactored to use `operator` as suggested by @byersiiasa) and let `_op_data()` take any number of args (as suggested by @pjuergens) to prepare for custom methods.

I think the PR is good to go - this was fun! Have a nice weekend!

File: pyam/core.py [hunk 2]
```diff
@@ -1729,6 +1730,126 @@ def col_apply(self, col, func, *args, **kwargs):
...
+        Returns
+        -------
+        :class:`IamDataFrame` or **None**
+            Computed timeseries data or None if `append=True`.
+        """
+        _value = _op_data(self, (a, b), name, "add", axis=axis)
+        if append:
+            self.append(_value, inplace=True)
+        else:
+            return IamDataFrame(_value, meta=self.meta)
```
gidden, line 1761: I guess it is possible some models/scenarios could be dropped in some of the operations if data is missing. should this be 

```
return IamDataFrame(_value, meta=self.meta[_value.index])
```

?

gidden: and if so - then this change should propagate down

Note, btw, this is why I made originally a generic `binary_op()` function, which did all the set up/tear down work - but can also be updated later!

pyam/index.py
```diff
@@ -63,8 +63,8 @@ def replace_index_values(df, level, mapping):
 def append_index_level(index, codes, level, name, order=False):
     """Append a level to a pd.MultiIndex"""
     new_index = pd.MultiIndex(
-        codes=index.codes + [codes],
-        levels=index.levels + [level],
+        codes=index.codes + [[codes] * len(index.codes[0])],
```
gidden, line 66: If there exists tests for this function, would it be possible to add one that would catch this case moving forward (i.e., which would have failed without this new update)? Or can you confirm that `test_append_index()` does so? Thanks!

pyam/utils.py
```diff
@@ -398,7 +398,8 @@ def merge_meta(left, right, ignore_conflict=False):
     if not diff.empty:
         left = left.append(right.loc[diff, :], sort=False)
 
-    return left
+    # remove any columns that are all-nan
+    return left.dropna(axis=1, how="all")
```
gidden, line 402: If there exists tests for this function, would it be possible to add one that would catch this case moving forward (i.e., which would have failed without this new update)? I couldn't find a specific test that would do so that has been added, perhaps missed. Thanks!

pyam/core.py
danielhuppmann: This is already happening automatically during initialization... https://github.com/IAMconsortium/pyam/blob/d4eafc319480b8a96bba3f6cc500c89f59646cf6/pyam/core.py#L189

pyam/index.py
danielhuppmann: The function wasn't actually operational and there were no tests before this PR. So now the function works and the test `test_append_index()` ensures that it works as expected going forward.

pyam/utils.py
danielhuppmann: All of the new ops-tests on the scenario-axis fail if this line is removed, should be enough to ensure that this works going forward...

gidden:
Looks great! Will let the CI run and then merge. Thanks @danielhuppmann !

PR #533, opened 2021-05-12
Implement an `apply()` function

### LINKED ISSUE #528: Implement an `apply()` function
Based on a suggestion by @pjuergens at https://github.com/IAMconsortium/pyam/issues/524#issuecomment-833272975, this issue takes one of the further improvements discussed in that issue:

Implement an `apply()` function mimicking pandas (see [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html)) to easily apply any function on the timeseries data (grouped along an axis and filtered by kwargs.

```python
# define any function you want to calculate
def custom_calculation(a, b, c):
    return a + b / c

# calculate a new variable with this function
df.apply(method=custom_calculation, axis="variable", append=True,
         a="First Variable", b="Second Variable", c="Last Variable")
```

pjuergens:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Implemented an `apply()` function, closes #528

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1850,7 +1850,9 @@ class IamDataFrame(object):
         else:
             return IamDataFrame(_value, meta=self.meta)
 
-    def apply(self, func, name="new variable", axis="variable", append=False, args=(), **kwds)
+    def apply(
+        self, func, name="new variable", axis="variable", append=False, args=(), **kwds
+    ):
         """Apply a function to components of timeseries data along an `axis`
 
         This function computes a function func using timeseries data selected along an `axis`
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1850,7 +1850,7 @@ class IamDataFrame(object):
         else:
             return IamDataFrame(_value, meta=self.meta)
 
-    def apply(self, name, components, func, axis="variable", append=False):
+    def apply(self, func, name="new variable", axis="variable", append=False, args=(), **kwds)
         """Apply a (self defined) function to components of timeseries data along an `axis`
 
         This function computes a (self defined) function func with arguments `components`.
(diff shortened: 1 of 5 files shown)


pyam/core.py
```diff
@@ -1850,6 +1850,37 @@ def divide(self, a, b, name, axis="variable", append=False):
         else:
             return IamDataFrame(_value, meta=self.meta)
 
+    def apply(self, name, components, func, axis="variable", append=False):
```
danielhuppmann, line 1853: This use of `components` could become confusing with more elaborate custom functions - I suggest to use `kwds` instead of `components` to follow the pandas apply function (see [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html)), so that the function could be used in this manner:

```python
def custom_calculation(a, b, c):
    return a + b / c

df.apply(custom_calculation, name="new variable", a="Primary Energy", b="Secondary Energy", c="Price")
```

pyam/core.py
```diff
@@ -1850,6 +1850,37 @@ def divide(self, a, b, name, axis="variable", append=False):
...
+        """Apply a (self defined) function to components of timeseries data along an `axis`
```
danielhuppmann, line 1854:
```suggestion
        """Apply a function to components of timeseries data along an `axis`
```

pyam/core.py
```diff
@@ -1850,6 +1850,37 @@ def divide(self, a, b, name, axis="variable", append=False):
...
+
+        This function computes a (self defined) function func with arguments `components`.
```
danielhuppmann, line 1856:
```suggestion
        This function computes a function func using timeseries data selected along an `axis`
        downselected by keyword arguments.
```

danielhuppmann (changes requested):
Thanks for this first draft, @pjuergens - this will be super-useful! A suggestion to make this more similar to pandas inline.

pjuergens:
It should be similar now to pandas apply with the possibility of both `args` (in my implementation before `components`) and `**kwargs` (in pandas-style however `**kwds`)

danielhuppmann:
FYI, I added the description of this PR to directly close the related issue when this PR gets merged

pjuergens:
Should be ready to go now :) I had to put the `_get_values` function inside `_ops.py`, because I could load `get_index_levels` from withing `utils.py`.

danielhuppmann:
> Should be ready to go now :) I had to put the `_get_values` function inside `_ops.py`, because I could load `get_index_levels` from withing `utils.py`.

Yeah, one has to be careful about circular dependencies... But it's fine to have that close to the actual use case.

danielhuppmann (approved):
Very sweet, thanks @pjuergens!

PR #534, opened 2021-05-17
added feature drop negative weights

### LINKED ISSUE #446: Average with negative weights
## Problem description

In several contexts, we discussed how to compute weighted averages when the weights are negative. The prime use case for this is an average carbon price over several regions, weighted by the (possibly negative) emissions in each region.

The current implementation of the [aggregate_region](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.aggregate_region) implements a direct weighted sum:

> value * weight / sum(weight)

This can lead to very counter-intuitive results - see more below.

## Illustration

Use the following snippet to explore the current behaviour:
```
def test_aggregate(value, weight):

    TEST_DF = pd.DataFrame([
        ['reg_a', 'Price|Carbon', 'USD/t CO2', value[0]],
        ['reg_b', 'Price|Carbon', 'USD/t CO2', value[1]],
        ['reg_a', 'Emissions|CO2', 'Mt CO2', weight[0]],
        ['reg_b', 'Emissions|CO2', 'Mt CO2', weight[1]],
    ],
        columns=['region', 'variable', 'unit', 2010],
    )

    return pyam.IamDataFrame(TEST_DF, model='model_a', scenario='scen_a')\
        .aggregate_region('Price|Carbon', weight='Emissions|CO2')._data.iloc[0]
```

1. If the sum of the weights is zero, the resulting average price is infinite
    ```
    test_aggregate(value=[1, 2], weight=[-1, 1])
    >>> inf
    ```
2. Depending on the order of the weight vector, it can also be negative infinity.
    ```
    test_aggregate(value=[1, 2], weight=[-1, 1])
    >>> -inf
    ```
3. If `sum(weight) == 0` and `value * weight == 0`, the returned value is `nan`, which is dropped when initiatilizing the IamDataFrame.
    ```
    test_aggregate(value=[1, 1], weight=[-1, 1])
    >>> Error
    ```
4. Other values can lead to situations where the "average" price is outside the range `min(value), max(value)`, e.g.
    ```
    test_aggregate(value=[1, 2], weight=[-1, 2])
    >>> 3
    ```

## Possible solutions

1. Leave as is, because it's up to the users to know this (including modeling teams uploading to the IIASA Scenario Explorer).

2. Leave as is, but write a warning message when weights are negative (question is whether users just ignore this).

3. Raise an error when any weight is negative, forcing users to explicitly choose an alternative approach (average, min, max).

4. Use the absolute of the weight vector. If there are three regions with emissions -1, 0 and 1, this would mean that the prices of region 1 and 3 are given equal weight and region 2 is ignored. This seems wrong...

5. Redefine the mathematical expression to 

    > value * weight / sum(abs(weight))

    However, this can also lead to strange outcomes, e.g., if all regional values are identical, and the weighted average is different.

6. Apply normalization (recalculate the weight vector such that `min(weight) == 0` and `max(weight) == 1`, leaving the relative distribution as is. This means that the price of the region with the lowest emissions is ignored (because its weight is 0).

7. Apply some shift/offset to make all weights positive. Suggestion by @byersiiasa: `weight = weight + 2 * min(weight)`.
    The vector `weight=[-1, 2]` would become `[1, 4]`.
    ```
    test_aggregate(value=[1, 2], weight=[1, 4])
    >>> 1.8
    ```
   This has the drawback that the offset multiplier is arbitrary - using a multiplier of 3 instead of 2 gives different results...
    ```
    test_aggregate(value=[1, 2], weight=[2, 5])
    >>> 1.7142
    ```

8. **Added** Write a warning when weights are negative and add a keyword-argument whether to drop (default) or not any values aggregated with negative weights.

9. **Added** Same as 8, but default is to include aggregation with negative weights.

10. Any other ideas...?

## Note

The function `numpy.avg()` faces the same problem. The devs decided to leave the unintuitive behaviour (solution 1 above), see https://github.com/numpy/numpy/issues/9825.

LauWien:

- [x] Tests Added
- [x] Documentation Added

# Description of PR
This PR implements the proposed 8th solution of the issue #446  

> Write a warning when weights are negative and add a keyword-argument whether to drop (default) or not any values aggregated with negative weights.

By default any values with negative weights are getting now dropped. Additionally a warning's is beeing printed: 

Warning

With the implemented boolean keyword argument `drop_negative` (default is True) this dropping can be overwritten by setting this keyword to False.

closes #446

### DIFF
--- a/pyam/_aggregate.py
+++ b/pyam/_aggregate.py
@@ -78,7 +78,7 @@ def _aggregate_recursive(df, variable):
 
 
 def _aggregate_region(
-    df, variable, region, subregions=None, components=False, method="sum", weight=None
+    df, variable, region, subregions=None, components=False, method="sum", weight=None, drop_negative=True
 ):
     """Internal implementation for aggregating data over subregions"""
     if not isstr(variable) and components is not False:
@@ -107,7 +107,7 @@ def _aggregate_region(
     else:
         weight_rows = subregion_df._apply_filters(variable=weight)
         _data = _agg_weight(
-            subregion_df._data[rows], subregion_df._data[weight_rows], method
+            subregion_df._data[rows], subregion_df._data[weight_rows], method, drop_negative
         )
 
     # if not `components=False`, add components at the `region` level
@@ -169,7 +169,7 @@ def _group_and_agg(df, by, method=np.sum):
     return df.groupby(cols).agg(_get_method_func(method))
 
 
-def _agg_weight(data, weight, method):
+def _agg_weight(data, weight, method, drop_negative):
     """Aggregate `data` by regions with weights, return indexed `pd.Series`"""
 
     # only summation allowed with weights
@@ -180,6 +180,15 @@ def _agg_weight(data, weight, method):
 
     if not data.droplevel(["variable", "unit"]).index.equals(weight.index):
         raise ValueError("Inconsistent index between variable and weight!")
+
+    if drop_negative is True:
+        if any(i < 0 for i in weight):
+            logger.warning("Some of the weights are negative. All data with negative values will be dropped. "
+                          "If you don't want the values to be dropped please use the keyword drop_negative=False")
+        # Drop negative weights
+            for index, value in weight.items():
+                if value < 0:
+                    weight = weight.drop(labels=[index])
 
     col1 = data.index.names.difference(["region"])
     col2 = data.index.names.difference(["region", "variable", "unit"])
--- a/pyam/_aggregate.py
+++ b/pyam/_aggregate.py
@@ -78,7 +78,14 @@ def _aggregate_recursive(df, variable):
 
 
 def _aggregate_region(
-    df, variable, region, subregions=None, components=False, method="sum", weight=None
+    df,
+    variable,
+    region,
+    subregions=None,
+    components=False,
+    method="sum",
+    weight=None,
+    drop_negative_weights=True,
 ):
     """Internal implementation for aggregating data over subregions"""
     if not isstr(variable) and components is not False:
@@ -103,11 +110,20 @@ def _aggregate_region(
     subregion_df = df.filter(region=subregions)
     rows = subregion_df._apply_filters(variable=variable)
     if weight is None:
+
+        if drop_negative_weights is False:
+            raise ValueError(
+                "Dropping negative weights can only be used with `weights`!"
+            )
+
         _data = _group_and_agg(subregion_df._data[rows], "region", method=method)
     else:
         weight_rows = subregion_df._apply_filters(variable=weight)
         _data = _agg_weight(
-            subregion_df._data[rows], subregion_df._data[weight_rows], method
+            subregion_df._data[rows],
+            subregion_df._data[weight_rows],
+            method,
+            drop_negative_weights,
         )
 
     # if not `components=False`, add components at the `region` level
@@ -169,7 +185,7 @@ def _group_and_agg(df, by, method=np.sum):
     return df.groupby(cols).agg(_get_method_func(method))
 
 
-def _agg_weight(data, weight, method):
+def _agg_weight(data, weight, method, drop_negative_weights):
     """Aggregate `data` by regions with weights, return indexed `pd.Series`"""
 
     # only summation allowed with weights
@@ -181,9 +197,22 @@ def _agg_weight(data, weight, method):
     if not data.droplevel(["variable", "unit"]).index.equals(weight.index):
         raise ValueError("Inconsistent index between variable and weight!")
 
+    if drop_negative_weights is True:
+        if any(weight < 0):
+            logger.warning(
+                "Some of the weights are negative. "
+                "All data weighted by negative values will be dropped. "
+                "To apply both positive and negative weights to the data, "
+                "please use the keyword argument `drop_negative_weights=False`."
+            )
+            # Drop negative weights
+            weight[weight < 0] = None
+
     col1 = data.index.names.difference(["region"])
     col2 = data.index.names.difference(["region", "variable", "unit"])
-    return (data * weight).groupby(col1).sum() / weight.groupby(col2).sum()
+    return (data * weight).groupby(col1).apply(
+        pd.Series.sum, skipna=False
+    ) / weight.groupby(col2).sum()
 
 
 def _get_method_func(method):
--- a/pyam/_aggregate.py
+++ b/pyam/_aggregate.py
@@ -78,7 +78,14 @@ def _aggregate_recursive(df, variable):
 
 
 def _aggregate_region(
-    df, variable, region, subregions=None, components=False, method="sum", weight=None, drop_negative=True
+        df,
+        variable,
+        region,
+        subregions=None,
+        components=False,
+        method="sum",
+        weight=None,
+        drop_negative=True
 ):
     """Internal implementation for aggregating data over subregions"""
     if not isstr(variable) and components is not False:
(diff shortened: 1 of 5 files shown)


tests/test_feature_aggregate.py
```diff
@@ -53,6 +52,20 @@
     columns=["variable", "unit"] + TEST_YEARS,
 )
 
+NEG_WEIGHTS_DF = pd.DataFrame(
```
LauWien, line 55: I created an additional `NEG_WEIGHTS_DF` as input data, because the initial usage of the simple_df and appending it with a data set with a negative value led to a more complicated testing function.

pyam/_aggregate.py
```diff
@@ -181,6 +191,19 @@ def _agg_weight(data, weight, method):
     if not data.droplevel(["variable", "unit"]).index.equals(weight.index):
         raise ValueError("Inconsistent index between variable and weight!")
 
+    if drop_negative is True:
+        if any(i < 0 for i in weight):
+            logger.warning(
+                "Some of the weights are negative. "
+                "All data with negative values will be dropped. "
+                "If you don't want the values to be dropped "
```
byersiiasa, line 199: like it!

pyam/_aggregate.py
```diff
@@ -169,7 +179,7 @@ def _group_and_agg(df, by, method=np.sum):
     return df.groupby(cols).agg(_get_method_func(method))
 
 
-def _agg_weight(data, weight, method):
+def _agg_weight(data, weight, method, drop_negative):
```
byersiiasa, line 182: I used this recently and found the docstring a little confusing... it wasn't really intuitive that to perform a weighted average calculation, one should actually choose `method='sum'` - even if that makes sense from the formula...
I would consider adding a comment on this in the docstring

byersiiasa (approved):

pyam/_aggregate.py
```diff
@@ -181,6 +191,19 @@ def _agg_weight(data, weight, method):
...
+            logger.warning(
+                "Some of the weights are negative. "
+                "All data with negative values will be dropped. "
+                "If you don't want the values to be dropped "
+                "please use the keyword drop_negative=False"
+            )
+            # Drop negative weights
+            for index, value in weight.items():
+                if value < 0:
+                    weight = weight.drop(labels=[index])
```
danielhuppmann, line 205: it should be possible to simply do the following
```suggestion
            weight = weight[weight>0]
```

pyam/core.py
```diff
@@ -1404,12 +1409,21 @@ def check_aggregate_region(
             (currently only supported with `method='sum'`)
         exclude_on_fail : boolean, optional
             flag scenarios failing validation as `exclude: True`
+        drop_negative : bool, default True
```
danielhuppmann, line 1412: Explicitly writing the default in the docstring is not in line with numpy-doc convention, better to simply write optional (see [here](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).

The pyam code doesn't consistently follow this convention, but we should try to apply the proper format in any new code (and clean up older functions as we go along).
```suggestion
        drop_negative : bool, optional
```

tests/test_feature_aggregate.py
danielhuppmann: Nice!

tests/test_feature_aggregate.py
```diff
@@ -299,6 +312,26 @@ def test_aggregate_region_with_components(simple_df):
...
+        .aggregate_region(variable, weight=weight, drop_negative=False)
+        ._data
+    )
+    test_2 = (
+        IamDataFrame(data)
+        .aggregate_region(variable, weight=weight, drop_negative=True)
+        ._data
+    )
+
+    assert not test_1.equals(test_2)
```
danielhuppmann, line 332: Here, it would be better to explicitly test some values, not just that they are not equal - they might be not equal for a lot of wrong reasons...

pyam/_aggregate.py
```diff
@@ -181,6 +191,19 @@ def _agg_weight(data, weight, method):
     if not data.droplevel(["variable", "unit"]).index.equals(weight.index):
         raise ValueError("Inconsistent index between variable and weight!")
 
+    if drop_negative is True:
+        if any(i < 0 for i in weight):
```
danielhuppmann, line 195: Should be more efficient to do the following:
```suggestion
        if any(weight<0):
```

pyam/_aggregate.py
```diff
@@ -107,7 +114,10 @@ def _aggregate_region(
     else:
```
danielhuppmann, line 114: Please add the following statement to the `if` just above (before computing `_data`):
```
if drop_negative is False:
    raise ValueError("Dropping negative weights can only be used with `weights`!)
```

pyam/_aggregate.py
danielhuppmann: Good idea!

tests/test_feature_aggregate.py
```diff
@@ -299,6 +312,26 @@ def test_aggregate_region_with_components(simple_df):
     assert _df.check_aggregate_region(v, components=["foo"]) is None
 
 
+@pytest.mark.parametrize(
+    "data, variable, weight",
+    ((NEG_WEIGHTS_DF, "Price|Carbon", "Emissions|CO2"),),
+)
+def test_agg_weight(data, variable, weight):
+
```
danielhuppmann, line 320: Using pytest-parametrization here is a bit overkill (unless you plan to add more versions of this test?)
```suggestion
def test_agg_weight():
    data = NEG_WEIGHTS_DF
    variable = "Price|Carbon"
    weight = "Emissions|CO2"
```

danielhuppmann:
FYI, I also added the statement "closes #446" to the PR description - so when this PR is merged, the issue will be closed automatically

tests/test_feature_aggregate.py
```diff
@@ -299,6 +312,37 @@ def test_aggregate_region_with_components(simple_df):
...
+        variable, weight=weight)
+    exp_1 = IamDataFrame(
+        pd.DataFrame([[np.nan, 5.25]], columns=[2005, 2010]),
+        model="model_a",
+        scenario="scen_a",
+        variable=variable,
+        unit="USD/tCO2",
+        region="World",
+    )
+    assert_iamframe_equal(test_1, exp_1)
```
LauWien, line 330: I know this is redundant but I get "ValueError: duplicate rows in 'data'" otherwise..

danielhuppmann (changes requested):
Thank you @LauWien!

I think that there is a flaw in your implementation, see the inline comment.

Also, the docstrings of pyam are not consistent, sometimes it's
```
arg : type, default <some value>
    bla ba
```
sometimes the (correct numpy-docs) style
```
arg : type, optional
    Bla ba
```
The idea is to pragmatically harmonize whenever a function is changed or updated - could you please do this for the docstring of the `aggregate_region()` function?

pyam/_aggregate.py
```diff
@@ -181,6 +197,21 @@ def _agg_weight(data, weight, method):
...
+                "All data weighted by negative values will be dropped. "
+                "To apply both positive and negative weights to the data, "
+                "please use the keyword argument `drop_negative_weights=False`"
+            )
+            # Drop negative weights
+            weight = weight.unstack()
+            for year in weight:
+                if any(weight[year] < 0):
+                    weight[year] = [region / float("NaN") for region in weight[year]]
+            weight = weight.stack()
```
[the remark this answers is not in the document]
LauWien: True - I didn't think about multiple scenarios! 
Yes, panda beeing "too smart" was my initial problem which led to the unstacking and stacking. Thanks for your input to replace the default `GroupBy.sum` :)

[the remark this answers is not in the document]
danielhuppmann: Writing software that is the "right level of smart" is quite difficult... 🙃

danielhuppmann:
Thank you Laura - can you please rebase and implement the other to-dos from the description (and then mark them as closed): adding yourself to the list of authors of the package and adding one line to the release notes?

PR #535, opened 2021-05-19
Units in operations

znicholls:

- [x] Tests Added
- [ ] Documentation Added
- [ ] Name of contributors Added to AUTHORS.rst

```
```

# Description of PR

Fixes (eventually) handling of units in operations.

### DIFF
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -110,7 +110,7 @@ def test_multiply_variable(test_df_year, append):
         scenario="scen_a",
         region="World",
         variable=v[2],
-        unit="EJ/yr",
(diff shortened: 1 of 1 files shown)


znicholls:
@gidden and @danielhuppmann I think you were the key inputs into #527. I had a quick look and I don't think the units are working (hopefully the changes make sense). Unfortunately I don't think there's a super obvious/easy solution. Something using pint is needed (and how we did it with scmdata) but there are speed tradeoffs which I've never seen a consensus decision on.

danielhuppmann:
Thanks @znicholls for raising this issue, good point!

PR #537, opened 2021-05-23
Add unit tests for using binary ops with numerical args

### LINKED ISSUE #530: Support binary operations with numerical values
Per a suggestion by @Rlamboll at https://github.com/IAMconsortium/pyam/pull/527#issuecomment-834279598 (and a bilateral comment by @OFR-IIASA), it would be a nice feature to extend the data-ops functions to take a numerical value, e.g.,

```python
df.multiply("Primary Energy", 3, "Primary Energy (end-use accounting)")
```

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

PR #533 implicitly generalized the binary-operation methods to enable using numbers as arguments. This PR adds unit tests to make sure this continues to work as expected, updates the docstrings, and adds a sanity check if a user would do something like `df.add(2, 3, "foo")`, where the return-object cannot be cast to an IamDataFrame.

closes #530

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 # Next Release
 
+- [#537](https://github.com/IAMconsortium/pyam/pull/537) Enhance binary ops to support numerical arguments
 - [#533](https://github.com/IAMconsortium/pyam/pull/533) Add an `apply()` function for custom mathematical operations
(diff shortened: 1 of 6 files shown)


danielhuppmann:
pinging @pjuergens for a review

danielhuppmann:
Merging without review (because this PR is mainly adding tests)

PR #531, opened 2021-05-11
Set explicit test for latest supported pandas version

danielhuppmann:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- ~Name of contributors Added to AUTHORS.rst~
- ~Description in RELEASE_NOTES.md Added~

# Description of PR

Per an issue reported by @GamzeUnlu95 that the current pyam version was incompatible with pandas 1.0.x, I added an explicit test for the "oldest" supported version of pandas and updated the requirements accordingly.

PR #540, opened 2021-05-31
adds iam-units to dependency check

gidden:
Small hotfix building on #531 which adds a bit of documentation to `setup.py` and includes `iam-units` in dependencies which are kept up to date in the new workflow.

### DIFF
--- a/.github/workflows/pytest-dependency.yml
+++ b/.github/workflows/pytest-dependency.yml
@@ -22,12 +22,13 @@ jobs:
       with:
         python-version: '3.7'
 
+    - name: Install specific out-dated version of dependencies
+      # Update the corresponding package requirements when changing
+      # minimum dependency versions
+      run: pip install pandas==1.1.1 matplotlib==3.2.0 iam-units==2020.4.12
+
     - name: Install other dependencies and package
       run: pip install -e .[tests,deploy,optional-io-formats,tutorials]
-
-    - name: Install specific out-dated version of dependencies
-      run: pip install pandas==1.1.1 matplotlib==3.2.0 iam-units==2020.4.12
-      # Update the corresponding package requirements when changing dependency versions
 
     - name: Test with pytest
       run: pytest tests --mpl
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Thanks @gidden, lgtm!

danielhuppmann:
Oh, seems that the explicit iam-units version actually doesn't work...?

.github/workflows/pytest-dependency.yml
```diff
@@ -23,8 +23,9 @@ jobs:
         python-version: '3.7'
 
     - name: Install specific out-dated version of dependencies
-      run: pip install pandas==1.1.1 matplotlib==3.2.0
-      # Update the corresponding package requirements when changing dependency versions
+      # Update the corresponding package requirements when changing
+      # minimum dependency versions
+      run: pip install pandas==1.1.1 matplotlib==3.2.0 iam-units==2020.4.12
```
danielhuppmann, line 28:
```suggestion
      run: pip install pandas==1.1.1 matplotlib==3.2.0 iam-units==2020.4.21
```

setup.py
```diff
@@ -11,6 +11,10 @@
   \/_/     \/_____/   \/_/\/_/   \/_/  \/_/
 """
 
+# NOTE TO DEVS
+# If you change a minimum version below, please explicitly set that
+# in our minimum-reqs test in the file
+# ./.github/workflows/pytest-depedency.yml
 REQUIREMENTS = [
     "argparse",
     "iam-units>=2020.4.12",
```
danielhuppmann, line 20:
```suggestion
    "iam-units>=2020.4.21",
```

danielhuppmann:
Seems that support for CO2e was added with https://github.com/IAMconsortium/units/pull/21, with the corresponding release on 2020.4.21

gidden:
Added the commits - let's see how the pipelines work out. Appears the test is already working, though =)

danielhuppmann:
Thanks @gidden

Issue #536, opened 2021-05-20
Automatted unit handling in operations


pjuergens:
This issue shall describe the problem already noted in the PR #535 

The `_op_data`-function doesn't handle units automatically. I guess, the corresponding functions `add`, `subtract`, `multiply`, `divide` and `apply` only work if the timeseries have the same units, but I didn't test yet what it does if timeseries have different units.

When doing `df.multiply('Emission Factor|Gases', 'Final Energy|Gases', 'Emissions|Gases')` I'd expect to have `Emissions|Gases` the units `Mt CO2`.

To make `_op_data` handle units automatically `pint` might be useful. Pint can do calculations with units, even with numpy-arrays. My first thoughts for it:

```python
import iam_units
# load iam-Units
registry = iam_units.registry
# calculate something with pint - this should be adapted to pyam timeseries and the units stored there
a = registry.Quantity(5, 'TW')
b = registry.Quantity(3, 'Mt')
c = a / b
# store the unit as string 'TW / Mt' - this needs to be passed to the pyam timeseries
unit = '{:~}'.format(c.units)
```

I think it's important to handle units when calculating on the variables-axis, e.g. emissions = emission factor * energy. However on other axis like regions or scenarios I'm not sure how much it makes sense and how difficult the implementation would be to handle different units at once.

danielhuppmann:
I started playing around with this feature and found a reasonably smart way to handle this using the iam-units package (which is based on pint), see https://github.com/danielhuppmann/pyam/commit/6509394e2b2ee87d8897a642e64f7ef28fdeb670

Follow-up question: how should the binary-ops function work if automated unit handling doesn't work? This happens if the units are not defined in the iam-units registry.

- Option 1: Add unknown units to the registry on the fly (which also doesn't always work, for example using `-`).
- Option 2: Add a new keyword argument `ignore_units` (default False) to all binary-ops functions.
    - If True, return an empty string as unit
    - If a string, use that as unit in the returned timeseries data.

@znicholls @khaeru @gidden @Rlamboll, any thoughts?

znicholls:
> how should the binary-ops function work if automated unit handling doesn't work?

My default here would be to raise an error about the unknown unit or let pint simply raise a dimensionality error or similar... After that, option 2 sounds good with one tweak. If `ignore_units` is `True`, I would put 'unknown' as the unit (empty string in pyam is treated as nan often so might be dangerous...).

khaeru:
[...]

PR #541, opened 2021-05-31
Support units in binary operations

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR changes the implementation of the binary ops to support units (via iam-units and pint).

Default behaviour:
 - Cast values and units to a `pint.Quantity` using the **iam-units** registry and perform operations with pint unit handling (`ignore_units=False`)
    **Warning**: By casting units to pints and then performing operations, the return-format may look different.
   In some of the tests, `EJ/yr` is transformed to `exajoule / year` (in full format) or `EJ / a` (in compact format, implemented here). This is over-ridden for some methods (addition, multiplication, division) if all units are identical.
 - Passing a pint quantity as argument will work, e.g, `df.add("Primary Energy", Quantity(2, "EJ/yr"), "new")`.
 - **pint** will raise appropriate errors if it doesn't know how to handle the units.
 - In line with common practice in the IAM community, a dimensionless quantity is marked as unit `""` (empty string) rather than the word `dimensionless` (this works in a round-trip to-from-to pint).

Alternative over-ride:
 - A user can set `ignore_units=True`, in which case the operations will work just on the values and set `unit` to "unknown" (per suggestion by @znicholls in https://github.com/IAMconsortium/pyam/issues/536#issuecomment-847569955) or to the value of `ignore_units` (if it's a string).

Known issues:
- pandas fillna does not work with non-numeric values (in particular not with pint.Quantity).

Side note:
 - I simplified the test for the `apply()` function.

To-dos in later PRs (it's already quite a beast):
 - Make the fillna feature work with pint.
 - Add `ignore_units` to the `apply()` function.
 - Support custom pint registry instances.
 - Pass unit formatting options to pint as kwargs.

closes #537 
closes #535

### DIFF
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -2,6 +2,7 @@ import operator
 import pandas as pd
 from pyam.index import append_index_level, get_index_levels
 from pyam.utils import to_list
+from iam_units import registry
 
 
 # these functions have to be defined explicitly to allow calling them with keyword args
@@ -70,14 +71,19 @@ def _op_data(df, name, method, axis, fillna=None, args=(), **kwds):
                 kwds[key] = _data[key]
 
     # apply method and check that returned object is valid
-    _value = method(*_args, **kwds)
-    if not isinstance(_value, pd.Series):
+    result = method(*_args, **kwds)
+    if not isinstance(result, pd.Series):
         msg = f"Value returned by `{method.__name__}` cannot be cast to an IamDataFrame"
-        raise ValueError(f"{msg}: {_value}")
+        raise ValueError(f"{msg}: {result}")
 
-    # insert the index level and reset the series name
-    _value.index = append_index_level(_value.index, codes=0, level=name, name=axis)
-    return _value.rename(index=None)
+    rename_args = ("dimensionless", "")
+    _value = pd.DataFrame(
+        [[i.magnitude, str(i.units).replace(*rename_args)] for i in result.values],
+        columns=["value", "unit"],
+        # append the `name` to the index on the `axis`
+        index=append_index_level(result.index, codes=0, level=name, name=axis),
+    )
+    return _value.set_index("unit", append=True)
 
 
 def _get_values(df, axis, value, cols, name):
@@ -102,7 +108,14 @@ def _get_values(df, axis, value, cols, name):
 
     """
     if any(v in get_index_levels(df._data, axis) for v in to_list(value)):
-        _data = df.filter(**{axis: value})._data.groupby(cols).sum().rename(index=name)
-        return _data, True
+        _data = df.filter(**{axis: value})._data.groupby(cols).sum()
+        _data = pd.Series(
+            [
+                registry.Quantity(v, u)
+                for v, u in zip(_data.values, _data.index.get_level_values("unit"))
+            ],
+            index=_data.reset_index("unit", drop=True).index,
+        )
+        return _data.rename(index=name), True
     else:
         return value, False
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -1,6 +1,6 @@
 import operator
 import pandas as pd
-from pyam.index import append_index_level, get_index_levels
+from pyam.index import append_index_level, get_index_levels, replace_index_values
 from pyam.utils import to_list
 from iam_units import registry
 from pint import Quantity
@@ -118,12 +118,12 @@ def _op_data(df, name, method, axis, fillna=None, args=(), ignore_units=False, *
 
     # separate pint quantities into numerical value and unit (as index)
     if ignore_units is False:
-        rename_args = ("dimensionless", "")
         _value = pd.DataFrame(
-            [[i.magnitude, str(i.units).replace(*rename_args)] for i in result.values],
+            [[i.magnitude, '{:~}'.format(i.units)] for i in result.values],
             columns=["value", "unit"],
             index=result.index,
         ).set_index("unit", append=True)
+        _value.index = replace_index_values(_value, "unit", {"dimensionless": ""})
 
     # otherwise, set unit (as index) to "unknown" or the value given by "ignore_units"
     else:
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -3,23 +3,41 @@ import pandas as pd
 from pyam.index import append_index_level, get_index_levels
 from pyam.utils import to_list
 from iam_units import registry
+from pint import Quantity
 
 
 # these functions have to be defined explicitly to allow calling them with keyword args
 def add(a, b):
+    a, b = _make_series(a, b)
     return operator.add(a, b)
 
 
 def subtract(a, b):
+    a, b = _make_series(a, b)
     return operator.sub(a, b)
 
 
 def multiply(a, b):
+    a, b = _make_series(a, b)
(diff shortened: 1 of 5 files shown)


pyam/_ops.py
```diff
@@ -70,14 +111,33 @@ def _op_data(df, name, method, axis, fillna=None, args=(), **kwds):
...
-        raise ValueError(f"{msg}: {_value}")
-
-    # insert the index level and reset the series name
+        raise ValueError(f"{msg}: {result}")
+
+    # separate pint quantities into numerical value and unit (as index)
+    if ignore_units is False:
+        rename_args = ("dimensionless", "")
+        _value = pd.DataFrame(
+            [[i.magnitude, str(i.units).replace(*rename_args)] for i in result.values],
```
pjuergens, line 123: Pint can return the unit as `EJ / yr` instead of `exajoule / year` with this code snipplet
```python
# format the the unit as string 'EJ / yr' 
unit = '{:~}'.format(c.units)
```
To implement it, the line should read something like
```python
[[i.magnitude, '{:~}'.format(str(i.units)).replace(*rename_args)] for i in result.values],
```
I'm not sure, if the `str()` is still neded then and how it returns dimensionless units

pjuergens:
So it's just a comment and not a full review. I hope some of the others have time for an in-depth-review, cause I'd really like to use this feature :)

danielhuppmann:
Thanks for the suggestion, @pjuergens - implemented it and updated the unit tests, because shorter units are generally preferable, I think. Also updated the description of this PR to reflect this change.

However, note that the underlying issue remains, because a user will generally not receive the same notation for a unit - in our case, `EJ / yr` is now returned as `EJ / a`. I don't see that as a problem, just something to be aware of...

pyam/_ops.py
danielhuppmann: Implemented.

pyam/_ops.py
```diff
@@ -1,24 +1,39 @@
...
+
+
+def _make_series(a, b):
+    """Cast instances of a pint.Quantity to a pd.Series
+
+    Calling an operation on a pd.Series and a pint.Quantity removes the units,
+    therefore a pint.Quantity is transformed to a pd.Series of quantities.
+    """
+    if isinstance(a, Quantity):
+        return pd.Series([a] * len(b), index=b.index), b
```
znicholls, line 33: We should probably add explicit tests for this.

[Pint-pandas](https://github.com/hgrecco/pint-pandas) is an attempt to support something like this. At the moment I don't think it works for pyam because the entire series has to have the same unit, but maybe something to keep in mind for future.

znicholls (commented):
Looks good to me. My major concern would be the casting of each individual value to/from pint.Quantity. That looks like it could be an expensive operation as soon as the data is a moderate size. We could test that, or we could just not worry about it for now.

danielhuppmann:
Thanks for the comment @znicholls! I had a look at [pint-pandas](https://github.com/hgrecco/pint-pandas), but as you write, this assumes that an entire column in a dataframe has the same unit - this is not something that pyam satisfies, not even on a per-variable level. A user could currently have a scenario ensemble where "Primary Energy" is in EJ for some scenarios and "GWh" in others. Something to discuss for a pyam v2.0 release...

About speed and casting, I'd follow the strategy I was taught by @gidden - first make it work, then make it fast...

znicholls:
> About speed and casting, I'd follow the strategy I was taught by @gidden - first make it work, then make it fast...

It's a great strategy!

pyam/_ops.py
```diff
@@ -46,13 +61,39 @@ def _op_data(df, name, method, axis, fillna=None, args=(), **kwds):
...
 
-    _data_kwds = {}
+    _data_kwds, _unit_kwds = {}, {}
     for i, (key, value) in enumerate(kwds.items()):
-        kwds[key], _data_kwds[key] = _get_values(df, axis, value, cols, key)
+        kwds[key], _unit_kwds[key], _data_kwds[key] = _get_values(
+            df, axis, value, cols, key
+        )
+
+    # fast-pass on units: override pint for some methods if all kwds have same the unit
```
pjuergens, line 77:
```suggestion
    # fast-pass on units: override pint for some methods if all kwds have the same unit
```

pyam/core.py
```diff
@@ -1838,13 +1849,16 @@ def subtract(self, a, b, name, axis="variable", fillna=None, append=False):
...
         else:
             return IamDataFrame(_value, meta=self.meta)
 
-    def multiply(self, a, b, name, axis="variable", fillna=None, append=False):
+    def multiply(
+        self, a, b, name, axis="variable", fillna=None, ignore_units=False, append=False
+    ):
         """Compute the product of timeseries data between `a` and `b` along an `axis`
 
         This function computes `a * b`. If `a` or `b` are lists, the method applies
```
pjuergens, line 1864: Should we add a note in the description here and for the divide-function that units might be reformatted by pint? For addition and subtraction this should not be necessary.
Also I was wondering which unit pint returns when multiplying e.g. `PJ` with `1 / PJ`. - Does it recognise it as `dimensionless` or will it return `PJ / PJ`? This might become important when using more complicated functions in `apply` - so we can keep it in mind to test in a following PR.

pjuergens (changes requested):
Basically just two minor comments concerning documentation. As far as I can tell the coding looks good to me. Thanks for the work!

pyam/core.py
danielhuppmann: Added a docstring to all four binary-ops functions and `apply()` - after all, as soon as not all units are identical, all functions will revert to abbreviated-units style of pint.

About more elaborate operations corner cases like you describe, my experience is that pint is quite good at dealing with those...

pjuergens: Addition and subtraction should lead to an error, if it's not done with identical units.

I think you are right about pint that it recognizes `PJ / PJ`. What it doesn't recognize are such things like `TWh / PJ` - but I would leave that to the user to convert it into `dimensionsless`.

pjuergens (approved):
In my opinion it's good to go :)

pyam/core.py
znicholls: > Addition and subtraction should lead to an error, if it's not done with identical units.

Why? If you add PJ and EJ or Mt CO2 and Gt CO2, the units are compatible so why should it error? 

> What it doesn't recognize are such things like `TWh / PJ`

Pint is generally fine with this. E.g. if I had W * s / J then you could convert to dimensionless and it should all work without an issue. Whether pint knows about TWh or not is a different question (is TWh defined in pint or iam_units?)

pjuergens: > Why? If you add PJ and EJ or Mt CO2 and Gt CO2, the units are compatible so why should it error?

I haven't thought about that :D I was thinking about different units like EJ and Mt CO2 and pint should raise and error then, hopefully ;-)

> Pint is generally fine with this. E.g. if I had W * s / J then you could convert to dimensionless and it should all work without an issue. Whether pint knows about TWh or not is a different question (is TWh defined in pint or iam_units?)

I just tried pint with meters and inch and it returned meter/inch, which I had to convert (using pint) into dimensionless later. But I don't see an issue with that - users shout convert the units either before or after by themselves.

PR #547, opened 2021-06-15
update ipcc colors for wg1

gidden:
Skipping formalities for now, as I don't think this needs to be in release notes.

This PR updates the IPCC colors in `plotting.py` to the latest available versions communicated by the TSU. Communicated colors in RGB are attached. RGB converted to hex with the below code snippet

```
import numpy as np
import matplotlib

with open('./WGI SSP colours.txt') as f:
    for l in f.readlines()[1:]:
        data = l.strip().split()
        rgb = np.array([int(x) for x in data[1:]]) / 256
        print(data[0], matplotlib.colors.rgb2hex(rgb))
```
[WGI SSP colours.txt](https://github.com/IAMconsortium/pyam/files/6658400/WGI.SSP.colours.txt)


znicholls (approved):
Easy as

danielhuppmann:
Restarted the tests, but one test (see below) is still failing because the updated colors violate the matplotlib-image-compare tolerance (not on Windows, interestingly).

https://github.com/IAMconsortium/pyam/blob/7d83a9836b01069d9fe8a2a50a5383beeab0fe7d/tests/test_plotting.py#L108

More importantly: can we (in the file or in this PR, or both) have a clear indication to the version of the color scheme?

byersiiasa (approved):
Cool!

gidden:
Hi @ozgeyelekci - this is our PR for updating `pyam` to the latest available IPCC WG1 colors. Given that https://github.com/IPCC-WG1/colormaps appears to be the canonical location for these definitions, would it be possible to tag the current state of that repo with a version number that we can reference here?

PR #551, opened 2021-06-25
Default to IamDataFrame-time-col on the x-axis of plots

danielhuppmann:

- ~Tests Added~
- [x] Documentation Added

# Description of PR

So far, when drawing plots with an IamDataFrame using a continuous-time format (i.e., `datetime`), a user had to specify `x="time"` explicitly. This PR makes the plotting module smarter to check (by default) the time-col attribute of an IamDataFrame.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 # Next Release
 
+- [#551](https://github.com/IAMconsortium/pyam/pull/551) Default to IamDataFrame-time-col on the x-axis of plots
 - [#549](https://github.com/IAMconsortium/pyam/pull/549) Make `plotly` an optional dependency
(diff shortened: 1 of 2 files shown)


danielhuppmann:
@awais307, it would be great if you take a look and see if this solves the issue discussed bilaterally when plotting IamDataFrame instance with hourly resolution.

danielhuppmann:
@awais307, it would be great if you take a look and see if this solves the issue discussed bilaterally when plotting IamDataFrame instance with hourly resolution.

awais307:
@danielhuppmann thanks for this effort. It works now without specifying `x='time` in sub-annual data.

danielhuppmann:
Thanks @awais307 for the nudge! And we should add some plotting tests specifically for continuous-time data at some point...

PR #557, opened 2021-07-06
Swap time for year keeping subannual resolution

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR migrates a feature first developed in https://github.com/openENTRANCE/nomenclature to split out the "time" column of an IamDataFrame into a year column (as int) plus keeping the information on the sub-annual resolution as a new extra-column "subannual".

### DIFF
--- a/tests/test_time.py
+++ b/tests/test_time.py
@@ -32,7 +32,7 @@ def test_swap_time_to_year(test_df, inplace):
     """Swap time column for year (int) dropping subannual time resolution (default)"""
 
     if test_df.time_col == "year":
-        return  # IamDataFrame with time domain `year` not relevant for this test
+        pytest.skip("IamDataFrame with time domain `year` not relevant for this test.")
 
     exp = test_df.data
     exp["year"] = exp["time"].apply(lambda x: x.year)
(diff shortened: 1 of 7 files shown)


tests/test_time.py
```diff
@@ -0,0 +1,93 @@
...
+OE_DATETIME = ["2005-10-01 23:15+01:00", "2010-10-02 23:15+01:00"]
+OE_SUBANNUAL_FORMAT = lambda x: x.strftime("%m-%d %H:%M%z").replace("+0100", "+01:00")
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+def test_swap_time_to_year(test_df, inplace):
+    """Swap time column for year (int) dropping subannual time resolution (default)"""
+
+    if test_df.time_col == "year":
+        return  # IamDataFrame with time domain `year` not relevant for this test
```
znicholls, line 35:
```suggestion
        pytest.skip("IamDataFrame with time domain `year` not relevant for this test")
```

Might be clearer?

tests/test_time.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_swap_time_to_year_errors(test_df):
+    """Assert that swapping time column for year (int) raises the expected errors"""
+
+    # swapping time to year raises when the IamDataFrame has time domain `year`
+    if test_df.time_col == "year":
+        match = "Time domain must be datetime to use this method"
+        with pytest.raises(ValueError, match=match):
+            test_df.swap_time_for_year()
+        return
```
znicholls, line 86: else might be clearer here (or split into two tests) but also fine to leave

znicholls (approved):
Very nice

tests/test_time.py
danielhuppmann: nice, thanks!

tests/test_time.py
danielhuppmann: Splitting into two tests would circumvent the very efficient use of four different versions of `test_df` (one with year, three with alternative datetime formats). So I added an `else` statement to make it clearer...

danielhuppmann:
Thanks for the very swift review, @znicholls!

PR #560, opened 2021-07-08
Add a feature to `swap_year_for_time`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR introduces the "inverse" of the `swap_time_for_year()` method. This should facilitate conversion from and to proper `datetime` timeseries and the "subannual" column currently used in the IIASA database infrastructure.

As shown in the tests, if you have an IamDataFrame with datetime format, then

```python
df.swap_time_for_year(subannual=True).swap_year_for_time()
```
will yield an identical object.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -632,6 +632,10 @@ class IamDataFrame(object):
         ------
         ValueError
             "time" is not a column of `self.data`
+
+        See Also
+        --------
+        swap_year_for_time
         """
         return swap_time_for_year(self, subannual=subannual, inplace=inplace)
 
@@ -652,6 +656,10 @@ class IamDataFrame(object):
         ------
         ValueError
             "year" or "subannual" are not a column of `self.data`
+
+        See Also
+        --------
+        swap_time_for_year
         """
         return swap_year_for_time(self, inplace=inplace)
 
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -11,9 +11,6 @@ import pandas as pd
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
-from pyam._debiasing import _compute_bias
-from pyam.time import swap_time_for_year
-
 try:
     from datapackage import Package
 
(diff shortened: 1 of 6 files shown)


danielhuppmann:
@loeffko @sebastianzwickl @erikfilias, this may be helpful for your ongoing work in the model linkage for openENTRANCE.

pyam/core.py
```diff
@@ -613,25 +612,58 @@ def interpolate(self, time, inplace=False, **kwargs):
...
             "time" is not a column of `self.data`
+
+        See Also
+        --------
+        swap_year_for_time
         """
         return swap_time_for_year(self, subannual=subannual, inplace=inplace)
 
+    def swap_year_for_time(self, inplace=False):
+        """Convert the `year` and `subannual` dimensions to `time` (as datetime).
```
znicholls, line 644: Worth adding that the conversion is done using `dateutil.parser.parse` on the combination of year and subannual?

znicholls (commented):
My one quick thought. Lots of edge cases could be covered here, depends how widely this feature is expected to be used.

pyam/core.py
danielhuppmann: Thanks, great suggestion - added this to the docstring including intersphinx mapping...

danielhuppmann:
> My one quick thought. Lots of edge cases could be covered here, depends how widely this feature is expected to be used.

The unit tests already cover five basic cases, including the one we are using in the openENTRANCE and ECEMF projects - this is the one where I see this function actually being used.

https://github.com/danielhuppmann/pyam/blob/5adccd81a8416aac3ed185b7bd33a7bb7c0cb515/tests/test_time.py#L55

I don't see any other use cases on the horizon (like using a different column name than "subannual" for the non-year parts, or using several columns) - but no reason why these two "swap" functions couldn't be extended in that direction with keyword arguments...

danielhuppmann:
Merging as there were no objections or further comments... Thanks @znicholls for the suggestion to improve the docs!

PR #561, opened 2021-07-16
Make `rename()` work with a custom index

danielhuppmann:

- ~Tests Added~
- ~Documentation Added~

```
```

# Description of PR

This PR generalizes the `rename()` function so that it can be used on an **IamDataFrame** with a custom index (ie., other than `["model", "scenario"]`, see [here](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame).

Two use cases of the custom index:
- retrieving scenario data from the IIASA database API using `default=False`, which returns all versions of scenarios instead of just the default, and which then uses the custom index `["model", "scenario", "version"]`, see [here](https://pyam-iamc.readthedocs.io/en/stable/api/iiasa.html#pyam.iiasa.Connection.query)
- the openMODEX project (afaik) developed a data format similar to the IAMC template but uses *source* instead of *model*. for generality

## To-Do

Rather than writing a single test just for this function, I'm thinking about how to redo `test_df` in the pytest-config so that a custom index can be tested across the entire pyam code base.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1030,11 +1030,12 @@ class IamDataFrame(object):
             raise ValueError(f"Conflicting rename args for columns {duplicate}")
         mapping.update(kwargs)
 
-        # determine columns that are not `model` or `scenario`
-        data_cols = set(self.dimensions) - set(META_IDX)
+        # determine columns that are not in the meta index
+        meta_idx = self.meta.index.names
+        data_cols = set(self.dimensions) - set(meta_idx)
 
         # changing index and data columns can cause model-scenario mismatch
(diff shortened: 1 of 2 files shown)


danielhuppmann:
fyi @jh-RLI @christian-rli

danielhuppmann:
Working on this further, I realized that restructuring the test suite and simultaneously fixing all issues where the custom-index-feature breaks is too much. Hence, I'm changing tactics - trusting that the existing tests cover all officially supported features, extending features to support custom-index IamDataFrame instances, then extending the `test_df` to also cover a case with a custom-index

PR #563, opened 2021-07-21
Enable `colors` as list in `plot.pie()`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR enables passing an explicit `colors` arg as list to to the pie plot, see https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pie.html

It also implements the `title` keyword argument, which was previously ignored in the implementation, and future-proofs pandas (using `min(level=0)` will be disabled).


pyam/plotting.py
```diff
@@ -274,7 +274,7 @@ def pie(
     value="value",
     category="variable",
     legend=False,
-    title=True,
+    title=False,
```
gidden, line 277: Here I would suggest a default value of `None`. In other plotting funcs, title can be `True` if there is logic available to find a reasonable default. If not, then it will be set by the kwarg value. Would suggest also updating docstring to `string, optional`

tests/test_plotting.py
```diff
@@ -410,6 +410,18 @@ def test_pie_plot_legend(plot_df):
...
 
 
+@pytest.mark.mpl_image_compare(**MPL_KWARGS)
+def test_pie_plot_colors(plot_df):
+    fig, ax = plt.subplots(figsize=(8, 8))
+    with update_run_control(RC_TEST_DICT):
+        (
+            plot_df.filter(
+                variable="Primary Energy", model="test_model", year=2010
+            ).plot.pie(ax=ax, category="scenario", colors=["red", "blue"], title="foo")
```
gidden, line 420: I know this is pedantic, but can you use values other than red and blue here so this test is easily visually distinguishable from the default values (e.g. in `test_pie_plot_labels.png`).

gidden (changes requested):
Great new feature! Some small adjustments requested. Is it possible to add a test of some kind for the 'future proofing' via `groupby`? Not required given all tests currently pass, but would be nice for future us's to not change..

tests/test_plotting.py
danielhuppmann: Done

pyam/plotting.py
danielhuppmann: Done - also change the docstring for the `title` keyword argument to follow [Axes.set_title()](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.set_title.html)

danielhuppmann:
Thanks for the suggestions, all implemented!

Not quite sure what you mean by adding a test for resolving a future-warning? If you go back to a previous [test run](https://github.com/IAMconsortium/pyam/runs/3179512237?check_suite_focus=true#step:6:270), section "Test with pytest", line 270, there is a warning that this particular code will stop working. If you compare this to the current test runs, this warning is gone. (There are plenty of others, though, tackling them one at a time).

gidden:
> Not quite sure what you mean by adding a test for resolving a future-warning? If you go back to a previous [test run](https://github.com/IAMconsortium/pyam/runs/3179512237?check_suite_focus=true#step:6:270), section "Test with pytest", line 270, there is a warning that this particular code will stop working. If you compare this to the current test runs, this warning is gone. (There are plenty of others, though, tackling them one at a time).

Specifically the line

```
mins = pd.concat([allmins, intermins]).groupby(level=0).min()
```

I was curious it is was possible to add any test for this. If not, that's fine.

gidden:
Ahh ok, I understand now. Thanks!

PR #573, opened 2021-08-19
Add documentation to `IamDataFrame.rename` for region aggregation use

phackstock:

- [ ] ~~Tests Added~~
- [x] Documentation Added
- [ ] ~~Name of contributors Added to AUTHORS.rst~~
- [ ] ~~Description in RELEASE_NOTES.md Added~~

```
```

# Description of PR

Add to the docstring of `IamDataFrame.rename` to make the fact that this can be used for region aggregation by renaming a number to regions to the same and then summing over them more obvious to the untrained eye. I had to read through the entire code to figure out the connection between renaming and region aggregation.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1010,6 +1010,11 @@ class IamDataFrame(object):
         columns given in `mapping`. Renaming can only be applied to the `model`
         and `scenario` columns, or to other data coordinates simultaneously.
 
+        This function can also be used to compute region aggregations. This is
+        done by renaming several constituent regions to the same name of a
+        common aggregation region and then applying groupby and sum(). Note that
+        for this use case check_duplicates **must be** set to False.
+
         Parameters
         ----------
         mapping : dict or kwargs


danielhuppmann:
-1 on this PR

Using `rename()` for region aggregation is a bad idea, because you would have to extend the note saying that this only works for a straight sum, but not for any other method like weighted average. So better to use the `aggregate_region()` method right for all types of regional aggregation.

If you think `aggregate_region()` doesn't take a mapping of several source-regions to several target-regions and `rename()` does - please start a PR to extend the features of the `aggregate_region()` method.

phackstock:
Ah ok, I wasn't aware that there is an `aggregate_region()` function, I agree using this is much better. 
The reason why I started the PR in the first place is that I was confused by the use of `rename()` in the irp-internal-workflow repository in `workflow.py` https://github.com/iiasa/irp-internal-workflow/blob/main/workflow.py#L54. Here it is used to aggregate the model native regions to the irp comparison regions if I understood that correctly. So when I added the mapping for MESSAGEix-Materials with the help of Peter I was confused about why the `rename()` function was used.

danielhuppmann:
Closing per my earlier comment that this is not the intended use of the `rename()` method.

PR #579, opened 2021-09-10
Improve performance of IamDataFrame initialization

danielhuppmann:
# Please confirm that this PR has done the following:

- ~Tests Added~ -> no functional changes, existing tests should cover all cases where something could go wrong
- [x] Name of contributors Added to AUTHORS.rst
- [x] Description in RELEASE_NOTES.md Added

# Description of PR

This PR implement faster and less memory-intensive initialization of an IamDataframe by doing the sorting and validation on the indexed pd.Series instead of the DataFrame. Based on memory-profiling work by @phackstock.

PR #580, opened 2021-09-15
Improve performance of IamDataFrame initialization (phase 2)

danielhuppmann:

- ~Tests Added~
- ~Documentation Added~

# Description of PR

This PR is a follow-up to #579, implementing further performance improvements.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,8 @@ The dependencies were updated to require `xlrd>=2.0` (previously `<2.0`) and `op
 
 ## Individual updates
 
-- [#579](https://github.com/IAMconsortium/pyam/pull/579) Increase performance of IamDataFrame initialization
+- [#580](https://github.com/IAMconsortium/pyam/pull/580) Improve performance of IamDataFrame initialization (phase 2)
+- [#579](https://github.com/IAMconsortium/pyam/pull/579) Improve performance of IamDataFrame initialization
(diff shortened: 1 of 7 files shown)


danielhuppmann:
Summarizing bilateral discussions with (and manual benchmarking done by) @phackstock - this PR shows again some improvements in memory usage. One interesting observation is that the initial commit here https://github.com/danielhuppmann/pyam/commit/d652297fc39c9176be661bb508d97d42d02b795b, which uses `df.set_index(.., append=True)` performs much *worse* than either the previous implementation or the "manual" adding-to-index using the pyam.index module...

danielhuppmann:
Running benchmarking with **pytest-monitor** and **memory-profiler** on the IAMC 1.5°C scenario ensemble data for all regions (~80MB, xlsx) shows that this PR increases time use by ~20%, but reduces memory use by 30%... Not quite sure if that is a worthwhile trade-off, or if it can be improved...

phackstock:
Regarding your question @danielhuppmann my vote would be in favor of saving memory even if the price for that is a longer execution time. 
My reasoning is that a longer execution time means more waiting for the user while memory savings can decide whether or not a user might be able to open a data set at all. 
Ideally, if you're working in a jupyter notebook you read the data only once and keep it in memory anyway so I don't think that a plus in execution time is that big of a deal.

danielhuppmann:
closing in favor of #729 and #730

PR #590, opened 2021-11-16
Update expected figures to use matplotlib 3.5

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

Alerted by the nightly tests... It seems that matplotlib v3.5 changed the order of the legends in stackplot-figures with totals. This PR simply updates the test-figures to reflect the new behavior (alternative approach: change the way that the "total" is added to the legend).

In addition, the PR silences a warning in the plotting module and changes the readme and install-bat-file from `py.test` to `pytest` (because this seems to be the more common usage).


pyam/plotting.py
```diff
@@ -437,7 +437,7 @@ def stack(
 
     def as_series(index, name):
         _idx = [i[0] for i in index]
-        return pd.Series([0] * len(index), index=_idx, name=name)
+        return pd.Series([0] * len(index), index=_idx, name=name, dtype="int")
```
gidden, line 440: Not clear why this is needed, can you let me know why?

Can this be sent through calling code? e.g., either with

1. move this `dtype` up as a kwarg or
2. add generic `**kwargs` which then passes through to this series call

gidden (commented):
Thanks for fixing the nightlies! Small question in review.

pyam/plotting.py
danielhuppmann: This is a function only used within this function (see just the following line) to select the rows where a timeseries crosses zero. Adding the dtype explicitly is done just to silence a warning in the log.

gidden: Gotcha - won't hold up. But I assume this will not work for non-integer times?

gidden: *facepalm* reread, and realize no issue here. sorry!

Issue #596, opened 2021-12-20
Integrated handling of categorical vs. datetime time formats


danielhuppmann:
# Background

The current implementation of pyam allows to use three different cases of temporal resolution:

1. yearly data (or longer timeframes, e.g., decadal intervals), indicated as `df.time_col = "year"`
2. yearly data with categorical subannual resolution, indicated as `df.time_col = "year"` and an extra-column "subannual".
   Values in that column could be "summer", "winter", or an abbreviated datetime format like `mm-dd hh:mmz` (see [here](https://github.com/openENTRANCE/openentrance/tree/main/definitions/subannual#wide-format-vs-long-format))
3. subannual resolution in a datetime format, `df.time_col = "time"` 

It is possible to mix use cases 1 and 2 (using a subannual column with a value "year"), but pyam currently does not support mixing 1/2 and 3 within one **IamDataFrame**.

This is because the long-format pd.Series `_data` (which holds the timeseries values) either has an index dimension "year" (as integer) or "time" (as datetime).

# Way forward

Going forward, applications and use cases are emerging where an **IamDataFrame** should support all use cases, for example having hourly data (power generation) as well as yearly values (carbon prices, installed capacity).

1. ~Split out the `_data` into two (or three?) distinct objects, one per use case (and refactor all methods that use it)~
2. ~Adapt the `_data` object by having two index dimensions ("year" and "time"), where for each row, only exactly one of year or time is not null~
3. Adapt the `_data` object by making the class of the entry in the "time" column specify the type of value (if integer, the value is a year; if datetime, it is subannual continuous resolution)
4. Like 3, but have a separate index dimension (column in a dataframe sense) that specifies the type (might be more reliable or have higher performance than type-checking-by-row?)
5. Force all entries in the time column to follow a datetime format
    Caveat:
    - Would probably not support data in a subannual non-continuous-time-format (e.g., "winter-night")
    - Difficult to know if a value for "2020-01-01 00:00" is intended for the first hour of a year or the full year
6. Have a wrapper class to hold two IamDataFrames, one for each type (added per suggestion by @znicholls below)

# Related questions

This issue is (for the time being) only intended for the internal handling - related (subsequent) discussions have to touch the i/o-aspects (reading from & writing to csv/xlsx, how to display the output of the `timeseries()` method)...

danielhuppmann:
@gidden @znicholls @Rlamboll, any thoughts?
FYI @phackstock @meksor

[...]

PR #597, opened 2021-12-23
Allow IamDataFrame instance with mixed 'year' and 'datetime' domain

danielhuppmann:

- [x] Tests Added

# Description of PR

This PR is a first step towards implementation of option 3 described in #596: allowing an **IamDataFrame** to have a "time" domain that has both "datetime" and "year" (as integer) values. The scope of this PR is very limited, to only allow initializing an **IamDataFrame** instance that has such a time domain. Filtering or returning the data in wide timeseries format is not tackled here (and fails in any instance that has a mixed index), to keep the PR reviewable (if anyone is interested).

The PR is directed to a feature-branch that will collect further work.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -33,7 +33,6 @@ from pyam.utils import (
     read_file,
     read_pandas,
     format_data,
-    format_time_col,
     merge_meta,
     find_depth,
     pattern_match,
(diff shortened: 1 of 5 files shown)


danielhuppmann:
I compared performance of this branch against the current main with two large year-based dataset: slightly faster (~10%), but also higher memory use. This is because the consistency-check and casting to int/pd.Timestamp now operates on the index instead of the time column - but this requires creating a full copy of the index rather than a copy of only the time column.

PR #598, opened 2021-12-30
Support mixed 'year' and 'datetime' domain

### LINKED ISSUE #596: Integrated handling of categorical vs. datetime time formats

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR implements refactors the IamDataFrame such that it supports "mixed" time domains (yearly, categorical, datetime), per option 3 of #596.

The PR implements the basic i/o workflow. It does not ensure that all methods support mixed time domain.
ToDo:
 - [x] Support initialization of an IamDataFrame instance that has year (int) and datetime time domain
 - [x] Add a "time domain" attribute
 - [x] Support filter (including change of time-column when filtering from mixed to year-only domain)
 - [x] Adjust `append()`
 - [x] Adjust `concat()`
 - [x] Support casting to "wide" format via `timeseries()`

closes #596


danielhuppmann:
Merging per announcement in the Slack workspace.

PR #608, opened 2022-01-21
`assert_iamframe_equals()` passes if all-nan-cols are present

### LINKED ISSUE #515: `assert_iamframe_equal()` fails on meta-columns with all-nan values
The testing function `assert_iamframe_equal()` fails when comparing two `IamDataFrame` instances where one side has a meta-column with all `nan` values and the other side doesn't have that columns. Arguably, all-nan should be treated as identical to "column doesn't exist".

```python
df.meta.dropna(axis="columns", how="all", inplace=True)
```

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

Currently, `assert_iamframe_equal()` fails if one of the IamDataFrames has a meta-column of all-nan's (and the column doesn't exist in the other IamDataFrame). Arguably, all-nan should be treated as identical to "column doesn't exist".

The PR also adds a page for the testing module to the docs.

Closes #515


pyam/testing.py
```diff
@@ -2,10 +2,36 @@
...
-    pdt.assert_frame_equal(a.meta, b.meta, check_dtype=False, check_like=True)
+    pdt.assert_frame_equal(
+        _drop_nan_col(left.meta),
+        _drop_nan_col(right.meta),
+        check_dtype=False,
+        check_like=True,
+    )
+
+
+def _drop_nan_col(df):
```
phackstock, line 36: Does that need to be its own function? I would say a single line would also nicely fit in `pdt.assert_frame_equal` above.

phackstock (approved):
Looks good to me.
Just one small comment and it seems there some conflict in the README that needs to be resolved.

pyam/testing.py
danielhuppmann: You're right, changed as you suggested...

phackstock:
Perfect, good to merge from my side.

Issue #609, opened 2022-01-23
pandas v1.4 breaks the IIASA database API


danielhuppmann:
The just released pandas v1.4 breaks querying data from the IIASA database API.

See https://github.com/IAMconsortium/pyam/runs/4912626605?check_suite_focus=true for the failing GitHub Actions workflow.

danielhuppmann:
closed via #621

PR #621, opened 2022-02-02
Fix pandas v1.4 breaks the IIASA database API #609

coroa:
pd.read_json in pandas 1.4 expects to receive strings not bytes.


danielhuppmann:
It seems that pandas v1.4 breaks a few more items, related to the filter implementation. If you have a minute to investigate what is going on...?

Note that I have a large PR #598, with a specific PR #614 moving the filter-implementation to a new (test) module. I hope to have that finished by end of next week, will then come back to this PR.

Issue #630, opened 2022-02-16
Adding a `IamSlice` feature


danielhuppmann:
# Description

We often want to know which elements exist on an dimension after filtering, e.g., which variables exist for a specific region (see https://github.com/IAMconsortium/nomenclature/pull/99). This can be done by the following

```python
vars = df.filter(region="Region A").variable
```

However, this is inefficient because this creates a full (downselected) copy of the (timeseries) `data` and `meta` tables.

# Proposed Solution

A new class **IamSlice** which is a derivative of the `pd.MultiIndex` of the internal `_data` pd.Series. The IamSlice is returned by the method `slice()`, which takes the same arguments as `filter()`.

## Expected usage

```python
vars = df.slice(region="Region A").variable
```

gidden:
+1, great idea!

danielhuppmann:
I like your idea about a boolean mask, but I would not how to implement it...

On the second idea about allowing `df[df.slice()]`, that sounds great and easily doable...

danielhuppmann:
Closed via #637

PR #637, opened 2022-03-03
Add initial IamSlice suggestion

coroa:
This would be a possible way to add an IamSlice feature proposed in #630 .

- [ ] Tests Added
- [ ] Documentation Added
- [ ] Name of contributors Added to AUTHORS.rst

```
```

# Description of PR

### DIFF
--- a/tests/test_unfccc.py
+++ b/tests/test_unfccc.py
@@ -11,7 +11,7 @@ UNFCCC_DF = pd.DataFrame(
 INDEX_ARGS = dict(model="UNFCCC", scenario="Data Inventory")
 
 
-def _test_unfccc_tier1():
+def test_unfccc_tier1():
     # test that UNFCCC API returns expected data and units
     exp = IamDataFrame(
         UNFCCC_DF,
--- a/tests/test_unfccc.py
+++ b/tests/test_unfccc.py
@@ -11,7 +11,7 @@ UNFCCC_DF = pd.DataFrame(
 INDEX_ARGS = dict(model="UNFCCC", scenario="Data Inventory")
 
 
-def test_unfccc_tier1():
+def _test_unfccc_tier1():
     # test that UNFCCC API returns expected data and units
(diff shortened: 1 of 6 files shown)


danielhuppmann:
I don't feel sufficiently qualified to do a proper review of the implementation, so I wrote a few tests and made some minor modifications - see #642 and #643

coroa:
Hmm ... that's a pickle. Not sure who could do a review. For me this was basically just trying to get something out to start a discussion.

Undesirable properties (i am aware of):
- `__getitem__` looses meta. it might need a couple of extra postprocessing lines, similar to what is currently in filter (maybe that should be factored out into a utility function). But we could also decide to post-pone the getitem implementation.
- The slice does not have the same dimension categorisation information as the full dataframe. ie its info can only talk broadly about dimensions, instead of differentiating between meta, timeseries and others.

danielhuppmann:
> * `__getitem__` looses meta. it might need a couple of extra postprocessing lines, similar to what is currently in filter (maybe that should be factored out into a utility function). But we could also decide to post-pone the getitem implementation.
> * The slice does not have the same dimension categorisation information as the full dataframe. ie its info can only talk broadly about dimensions, instead of differentiating between meta, timeseries and others.

I don't think that this is a problem. I'd even remove the `__repr__` and `info` methods in a first implementation.

tests/test_unfccc.py
```diff
@@ -11,7 +11,7 @@
 INDEX_ARGS = dict(model="UNFCCC", scenario="Data Inventory")
 
 
-def test_unfccc_tier1():
+def _test_unfccc_tier1():
```
danielhuppmann, line 14: This was a quickfix to make unit-tests pass, already correctly fixed via #647 
```suggestion
def test_unfccc_tier1():
```

danielhuppmann:
closing in favor of #657

PR #639, opened 2022-03-07
Un-pin dependency `pint <= 0.17`

### LINKED ISSUE #638: Un-pin dependency `pint <= 0.17`
As of https://github.com/IAMconsortium/pyam/pull/589 `pint` got pinned to version 0.17. To keep compatible with the latest version, currently 0.18, this pin should be removed.

LauWien:
# Description of PR

This PR un-pins the dependency `pint <= 0.17`, which got introduced in #589. 
Closes #638.

### DIFF
--- a/.github/workflows/pytest-dependency.yml
+++ b/.github/workflows/pytest-dependency.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install specific out-dated version of dependencies
       # Update the package requirements when changing minimum dependency versions
       # Please also add a section "Dependency changes" to the release notes
-      run: pip install pandas==1.1.1 numpy==1.19.0 matplotlib==3.5.0 iam-units==2020.4.21 xlrd==2.0 pint==0.17
+      run: pip install pandas==1.1.1 numpy==1.19.0 matplotlib==3.5.0 iam-units==2020.4.21 xlrd==2.0 pint==0.18
 
     - name: Install other dependencies and package
       run: pip install -e .[tests,optional_plotting,optional_io_formats,tutorials]


danielhuppmann:
So the failing test reminded me why I didn't remove the pin at the time. Can you please insert minimum requirements for pint and iam-units and update the legacy-dependency test?

Please also add a section "## Dependency changes" to the release notes?

.github/workflows/pytest-dependency.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - name: Install specific out-dated version of dependencies
       # Update the package requirements when changing minimum dependency versions
       # Please also add a section "Dependency changes" to the release notes
-      run: pip install pandas==1.1.1 numpy==1.19.0 matplotlib==3.5.0 iam-units==2020.4.21 xlrd==2.0
+      run: pip install pandas==1.1.1 numpy==1.19.0 matplotlib==3.5.0 iam-units==2020.4.21 xlrd==2.0 pint==0.17
```
danielhuppmann, line 28: This is not consistent with the specification and is actively overwritten by the installation in the following, see the [Actions log](https://github.com/IAMconsortium/pyam/runs/5448754850?check_suite_focus=true)
```suggestion
      run: pip install pandas==1.1.1 numpy==1.19.0 matplotlib==3.5.0 iam-units==2020.4.21 xlrd==2.0 pint==0.18
```

danielhuppmann (commented):
There is inconsistency between the new requirements and the legacy-dependency test.

More importantly: there was no functional change, so I suspect that the current codebase is actually compatible with pint 0.17 (or even earlier). Did you check? The **iam-units** seems to be backward-compatible back to pint 0.11.

.github/workflows/pytest-dependency.yml
LauWien: Ah, then I misunderstood the lagacy-dependency test.

danielhuppmann:
The legacy-dependency tests specifically takes the oldest supported version of several important packages, to make sure that our information on "oldest supported version" is actually correct.

If you look at the log of the GitHub action workflow, it first installed pint 0.17, and then installed pyam - but because pint 0.17 is not in line with the stated pyam requirements, it actually installed the latest version of pint (currently 0.18).

LauWien:
Superseded by https://github.com/IAMconsortium/pyam/pull/640.

PR #642, opened 2022-03-09
Add tests and set correct return type for `time`

danielhuppmann:
# Description of PR

This PR adds several tests for the `IamSlice` feature. To ensure consistency, the tests are implemented directly as alternatives to the `filter()` tests.

It also implements the different return-type for the `time` attribute (pd.Index, all other dimension-attributes are lists).

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -477,7 +477,9 @@ class IamDataFrame(object):
         - :class:`pandas.Index` if the time domain is 'mixed'
         """
         if self._time is None:
-            self._time = pd.Index(get_index_levels(self._data, self.time_col))
+            self._time = pd.Index(
+                self._data.index.unique(level=self.time_col).values, name="time"
+            )
 
         return self._time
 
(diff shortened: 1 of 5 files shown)


pyam/core.py
```diff
@@ -113,7 +113,7 @@ def __getattr__(self, attr):
 
         if attr in self.dimensions:
             ret = self._iamcache[attr] = self.index[self].unique(level=attr).tolist()
-            return ret
+            return pd.Index(ret) if attr == "time" else ret
```
coroa, line 117: Ah i was not aware that the time representation was returning proper index types instead of lists. Let's save the indices then, instead of round-robin converting:
```python
    def __getattr__(self, attr):
        ret = object.__getattribute__(self, "_iamcache").get(attr)
        if ret is not None:
            return ret.tolist() if attr != "time" else ret

         if attr in self.dimensions:
            ret = self._iamcache[attr] = self.index[self].unique(level=attr)
            return ret.tolist() if attr != "time" else ret
```

coroa (changes requested):
Looks good

danielhuppmann:
Implemented the suggested change. I noticed that this changed the "name"-attribute of the returned pd.Index object (from `None` to `time`). This actually seems to be preferable behavior, so I also implemented the minimal changes to make `IamDataFrame.time` behave in the same manner.

PR #643, opened 2022-03-09
Add a proper length attribute

danielhuppmann:
# Description of PR

This PR fixes the `__len__` attribute, which should only count the timeseries data points actually included in the slice.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -117,6 +117,9 @@ class IamSlice(pd.Series):
 
         return super().__getattr__(attr)
 
+    def __len__(self):
+        return sum(self.values)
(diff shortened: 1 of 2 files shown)


pyam/core.py
```diff
@@ -117,6 +117,9 @@ def __getattr__(self, attr):
 
         return super().__getattr__(attr)
 
+    def __len__(self):
+        return sum(self.values)
```
coroa, line 121: You should always prefer the pandas native methods.

```suggestion
        return self.sum()
```

coroa (changes requested):

PR #650, opened 2022-04-05
Add AR6 category colors to PYAM_COLORS

jkikstra:
# Description of PR

Adds the AR6 Category and IMP colours to pyam.


jkikstra:
Have updated the test figures. 
As expected, only https://github.com/IAMconsortium/pyam/blob/56fcc56f6f38fdd9f089074379babbfb33c9a476/tests/expected_figs/test_line_PYAM_COLORS.png changed.

Tests locally are failing, but that seems to be due to `pint` and/or `pandas_datareader`, so should not really part of this PR.

danielhuppmann:
Thanks @jarmo! I pinned pint as a quickfix, this should resolve the failing unit tests in this PR.

Could you please do the following:

- [ ] Revert the latest commit and then commit only the one changed figure to this branch
- [ ] Add a comment in each line explaining the meaning of the abbreviations and a reference to the description or source in the AR6 WG3 report, so that users have a chance to know what "C2" or "IMP-Neg" are...
- [ ] Add a line to the release notes

Then good to go, and I think this is a nice feature addition that warrants a new release (which I will take of).

jkikstra:
Hi Daniel and Matt, thanks for the guidance.

I did that now. 

Tried to also update the python notebook that is part of the docs,  `ipcc_colors.ipynb`, but my IIASA desktop seems unhappy due to some dll issue in my venv, so would probably be more efficient if I leave that quick update to you.
image

danielhuppmann (approved):
Thanks @jkikstra for adding the colors and docs!

The IPCC-colors tutorial notebook is run automatically as part of building the docs, so this will show the added AR6 colors. No need to change anything explicitly.

PR #657, opened 2022-05-06
Add an `Iamslice`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR implements an `IamSlice` based on the initial work by @coroa in #637. I rebased the branch and added basic documentation.

One question: in the initial work, `__repr__` combined a summary similar to an **IamDataFrame** and a representation of the **pd.Series**. In the usual test case filtered for `scenario="scen_b", this yields.

```python
<class 'pyam.slice.IamSlice'>
Index dimensions and data coordinates:
   model    : model_a (1)
   scenario : scen_b (1)
   region   : World (1)
   variable : Primary Energy (1)
   unit     : EJ/yr (1)
   year     : 2005, 2010 (2)

model    scenario  region  variable             unit   year
model_a  scen_a    World   Primary Energy       EJ/yr  2005    False
                                                       2010    False
                           Primary Energy|Coal  EJ/yr  2005    False
                                                       2010    False
         scen_b    World   Primary Energy       EJ/yr  2005     True
                                                       2010     True
dtype: bool
```

I find that a bit confusing, because the list of variables shows only "Primary Energy" but the series also shows "Primary Energy|Coal" (as False, but still). I therefore removed the second part, so that `__repr__` only shows the summary overview.

If you think that having the slice as **pd.Series**, we could add a method `as_series()`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+# Next release
+
+- [#657](https://github.com/IAMconsortium/pyam/pull/657) Add an `IamSlice` class
+
 # Release v1.4.0
 
 ## Highlights
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -13,6 +13,7 @@ and methods.
    api/general
    api/iamdataframe
    api/database
+   api/slice
    api/filtering
    api/compute
    api/plotting
--- a/doc/source/api/slice.rst
+++ b/doc/source/api/slice.rst
@@ -1,0 +1,7 @@
+.. currentmodule:: pyam
+
+The **IamSlice** class
+======================
+
+.. autoclass:: IamSlice
+   :members: dimensions, time
--- a/doc/source/api/slice.rst
+++ b/doc/source/api/slice.rst
@@ -4,4 +4,4 @@ The **IamSlice** class
 ======================
 
 .. autoclass:: IamSlice
-   :members: dimensions, time
+   :members: dimensions, time, info
--- a/pyam/__init__.py
+++ b/pyam/__init__.py
@@ -10,6 +10,7 @@ except ModuleNotFoundError:
     from importlib_metadata import version, PackageNotFoundError
 
 from pyam.core import *
+from pyam.slice import IamSlice
 from pyam.utils import *
 from pyam.statistics import *
 from pyam.timeseries import *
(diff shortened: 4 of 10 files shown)


coroa:
> One question: in the initial work, __repr__ combined a summary similar to an IamDataFrame and a representation of the pd.Series. In the usual test case filtered for `scenario="scen_b", this yields.

Well it is a `pd.Series`, ie. `isinstance(..., pd.Series)` will be `True`, so there is no point to a separate method (`pd.Series(a_slice)` will work fine).

My main argument for showing the Series representation is that it being a Series opens up several interesting computations, that people might want to know about, which don't make so much sense, if you don't know it is a boolean series.

The most interesting one is that you can do boolean operations, ie if you want from your example
```python
s_scen_a = test_df.slice(scenario="scen_a")
s_coal = test_df.slice(variable="*|Coal")

s_scen_a & ~ s_coal
```
will give you
```python
<class 'pyam.slice.IamSlice'>
Index dimensions and data coordinates:
   model    : model_a (1)
   scenario : scen_a (1)
   region   : World (1)
   variable : Primary Energy (1)
   unit     : EJ/yr (1)
   year     : 2005, 2010 (2)

model    scenario  region  variable             unit   year
model_a  scen_a    World   Primary Energy       EJ/yr  2005    True
                                                       2010    True
                           Primary Energy|Coal  EJ/yr  2005    False
                                                       2010    False
         scen_b    World   Primary Energy       EJ/yr  2005    False
                                                       2010    False
dtype: bool
```

danielhuppmann:
I agree with the implementation and behavior. I'm just worried that implementing `__repr__` as combining the **IamDataFrame**-style overview plus the **Series**-print can be confusing.

Options:
0. Leave as is, ignore my concern because this will be an expert-only feature anyway
1. Only show the Series-representation
2. Only show the IamDataFrame-style-overview

coroa:
I guess the overview is what is going to be the most helpful. So, i'd vote rather for showing the overview (although, i like seeing both). ie. let's keep it as it is here now. Then we can merge!

What this PR is missing and what is leftover as a new PR would be to refactor the second half of IamDataFrame.filter into a new method (could be called `subset` for instance) or generic helper and use that from `__getitem__` and `filter`; so that `idf[idf.slice(**kwargs)]` is equivalent to `idf.filter(**kwargs)`.

coroa (approved):

PR #659, opened 2022-05-13
add offset method to compute offsets relative to base-year values

gidden:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR introduces an `offset()` method which allows a user to compute offsets relative to base-time-periods in the same style as `normalize()`.

        For example, offsetting from `year=2005` will provide data
        *relative* to `year=2005` such that the value in 2005 is 0 and 
        all other values `value[year] - value[2005]`.

        Conceptually this operation performs as:
        ```
        df - df.filter(**kwargs) + padding
        ```

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1269,6 +1269,46 @@ class IamDataFrame(object):
         if not inplace:
             return ret
 
+    def offset(self, padding=0, inplace=False, **kwargs):
+        """Compute new data which is offset from a specific data point
+
+        For example, offsetting from `year=2005` will provide data
+        *relative* to `year=2005` such that the value in 2005 is 0 and 
+        all other values `value[year] - value[2005]`.
+
+        Conceptually this operation performs as:
+        ```
+        df - df.filter(**kwargs) + padding
+        ```
+
+        Note: Currently only supports normalizing to a specific time.
+
+        Parameters
+        ----------
+        padding : float, optional
+            an additional offset padding
+        inplace : bool, optional
+            if :obj:`True`, do operation inplace and return None
+        kwargs
+            the column and value on which to offset (e.g., `year=2005`)
+        """
+        if len(kwargs) > 1 or self.time_col not in kwargs:
+            raise ValueError("Only time(year)-based normalization supported")
+        ret = self.copy() if not inplace else self
+        df = ret.data
+        # change all below if supporting more in the future
+        cols = self.time_col
+        value = kwargs[self.time_col]
+        x = df.set_index(IAMC_IDX)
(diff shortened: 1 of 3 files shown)


pyam/core.py
```diff
@@ -1269,6 +1269,46 @@ def normalize(self, inplace=False, **kwargs):
...
+        df = ret.data
+        # change all below if supporting more in the future
+        cols = self.time_col
+        value = kwargs[self.time_col]
+        x = df.set_index(IAMC_IDX)
+        x["value"] -= x[x[cols] == value]["value"]
+        x["value"] += padding
+
+        x = x.reset_index()
+        ret._data = x.set_index(self.dimensions).value
```
coroa, line 1307: Since `set_index` is an expensive factorization operation on large datasets, i would suggest to avoid it in favour of working directly on the series representation:

```suggestion
        data = ret._data
        base_value = data.loc[data.index.isin([value], level=self.time_col)].droplevel(self.time_col)
        ret._data = data - base_value + padding
```

coroa: Caveat with both implementations: If the base value in the reference year is missing for a timeseries the whole timeseries will be nan and dropped from the iam dataframe.

coroa (commented):
I am not against including such a function, but it is also quite specific. Unsure, whether discoverability will become an issue at some point or already is.

pyam/core.py
danielhuppmann: I think that the behavior "drop values if the reference is missing" is consistent with the default behavior of other methods, e.g., [`add()`](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.add).

Or we could of course add a `fillna` argument (at a later point)...

danielhuppmann:
Thanks @coroa for suggesting the high-performance implementation alternative!

I don't mind adding the function directly for now. If discoverability becomes an issue, we could delegate methods operating only on the timeseries data to a submodule, similar to what I did with the [`compute`](https://pyam-iamc.readthedocs.io/en/stable/api/compute.html) module...

pyam/core.py
coroa: > Or we could of course add a fillna argument (at a later point)...

That would mean the change:

```python
    def offset(..., fillna=None):
        [...]
        ret._data = data.subtract(base_value, fillna=fillna) + padding
```

gidden:
Hi both - have updated based on your suggestions. Should be ready.

coroa:
LGTM

gidden:
Given positive response here, will go ahead with merge. See #662 for an issue on expanding documentation further.

PR #665, opened 2022-05-23
Fix/concat

### LINKED ISSUE #664: concat not working with non-standard index dimensions
When concatenating `IamDataFrame` objects that have non-standard index dimensions (i.e. anything other than `model, scenario`) you currently get an error merging the meta indicators.

### Minimum example to reproduce

Creating two `IamDataFrames` `df1, df2` with index dimensions `model, scenario, version` and running `pyam.concat`:

```python
import pyam
import pandas as pd

df1 = pyam.IamDataFrame(
    pd.DataFrame(
        [["model_a", "scenario_a", 1, "region_a", "variable_a", "unit", 1]],
        columns=["model", "scenario", "version", "region", "variable", "unit", 2005],
    ),
    index=pyam.DEFAULT_META_INDEX + ["version"],
)
df2 = pyam.IamDataFrame(
    pd.DataFrame(
        [["model_a", "scenario_a", 2, "region_a", "variable_a", "unit", 2]],
        columns=["model", "scenario", "version", "region", "variable", "unit", 2005],
    ),
    index=pyam.DEFAULT_META_INDEX + ["version"],
)

pyam.concat([df1, df2])
```

yields the following error: 

```bash
ValueError: Incompatible `index=['model', 'scenario']` with `meta` (index=['model', 'scenario', 'version'])!
```

### Proposed fix

As the last step of `pyam.concat` is the creation of a new `IamDataFrame` it should work so simply carry over the index dimensions of the concatenated meta table, like so:

```python
return IamDataFrame(
        pd.concat(ret_data, verify_integrity=False), meta=ret_meta, index = ret_meta.index.names
    )
```

phackstock:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

closes #664. 

The `pyam.concat` function now carries over the index of the merged meta table.
Added a unit test to check under `test_feature_append_and_concat.test_concat_non_standard_index`.
For documentation I added an additional line to the `notes` section of the `pyam.concat` function.

One open question from my side would be if there is more documentation than the added docstring required. My hunch would be that it's fine as is since the behavior is not really new but rather more working in line with user expectations.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2793,6 +2793,8 @@ def concat(objs, ignore_meta_conflict=False, **kwargs):
     The *meta* attributes are merged only for those objects of *objs* that are passed
     as :class:`IamDataFrame` instances.
 
+    The index dimension of the return value is carried over from index of the *meta*
+    indicators.
     """
     if not islistable(objs) or isinstance(objs, pd.DataFrame):
         raise TypeError(f"'{objs.__class__.__name__}' object is not iterable")
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2793,8 +2793,8 @@ def concat(objs, ignore_meta_conflict=False, **kwargs):
     The *meta* attributes are merged only for those objects of *objs* that are passed
     as :class:`IamDataFrame` instances.
 
-    The index dimension of the return value is carried over from index of the *meta*
-    indicators.
+    The :attr:`dimensions` and :attr:`index` names of all elements of *dfs* must be
+    identical. The returned IamDataFrame inherits the dimensions and index names.
     """
     if not islistable(objs) or isinstance(objs, pd.DataFrame):
         raise TypeError(f"'{objs.__class__.__name__}' object is not iterable")
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2844,7 +2844,11 @@ def concat(objs, ignore_meta_conflict=False, **kwargs):
             )
 
     # return as new IamDataFrame, this will verify integrity as part of `__init__()`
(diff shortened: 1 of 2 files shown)


pyam/core.py
```diff
@@ -2793,6 +2793,8 @@ def concat(objs, ignore_meta_conflict=False, **kwargs):
     The *meta* attributes are merged only for those objects of *objs* that are passed
     as :class:`IamDataFrame` instances.
 
+    The index dimension of the return value is carried over from index of the *meta*
+    indicators.
```
danielhuppmann, line 2797: Add an explanation that index and data dimensions have to be identical.
```suggestion
    The :attr:`dimensions` and :attr:`index` names of all elements of *dfs* must be
    identical. The returned IamDataFrame inherits the dimensions and index names.
```

tests/test_feature_append_concat.py
```diff
@@ -104,6 +104,38 @@ def test_concat(test_df, reverse, iterable):
     npt.assert_array_equal(ts.iloc[2].values, ts.iloc[3].values)
 
 
+def test_concat_non_standard_index():
```
danielhuppmann, line 107: I'd rather use "default" instead of "standard" because model-scenario is the default *index* when initializing an IamDataFrame.
```suggestion
def test_concat_non_default_index():
```

danielhuppmann (approved):
Looks good, two minor suggestions on the documentation...

phackstock:
Thanks for the quick review @danielhuppmann. If the tests all run through I'd go ahead with the merge.

coroa:
Thanks for that! I literally stumbled over it last night.

PR #682, opened 2022-07-12
Remove version pin on pint

coroa:
The original cause for constraining with < 0.19 as explained in
https://github.com/IAMconsortium/units/pull/37 has been addressed by
pint version 0.19.2.


danielhuppmann (approved):
Thanks @coroa, lgtm!

PR #684, opened 2022-07-29
Use new IIASA-manager API with token refresh

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR refactors the IIASA-manager-API authentication to the new (short-lived) token, including automated refresh. The PR also stricter behavior when reading a config file fails, and adds explicit test if arguments are missing from the credentials.

We decided to leave the current legacy-implementation for the anonymous login and application-list, because these features will change with the new ixmp package.

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 # set requests-logger to WARNING only
 logging.getLogger("requests").setLevel(logging.WARNING)
 
-_AUTH_URL = "https://api.manager.ece.iiasa.ac.at/legacy"
+_AUTH_URL = "https://api.manager.ece.iiasa.ac.at"
 _CITE_MSG = """
 You are connected to the {} scenario explorer hosted by IIASA.
  If you use this data in any published format, please cite the
@@ -138,7 +138,7 @@ class Connection(object):
     @property
     @lru_cache()
     def _connection_map(self):
-        url = "/".join([self._auth_url, "applications"])
+        url = "/".join([self._auth_url, "legacy", "applications"])
         r = requests.get(url, headers=self._headers)
         _check_response(r, "Could not get valid connection list")
         aliases = set()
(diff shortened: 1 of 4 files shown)


meksor:
The equivalent to the anonymous login would be logging in as the "guest" user.
Although I don't know if that user has a usable password set yet.

danielhuppmann:
> The equivalent to the anonymous login would be logging in as the "guest" user.
> Although I don't know if that user has a usable password set yet.

Ok, not sure if the "guest" user *should have* a password... Let's discuss next week.

pyam/iiasa.py
```diff
@@ -45,51 +52,93 @@ def set_config(user, password, file=None):
...
-    r = requests.post(url, headers=headers, data=json.dumps(creds))
-    user = creds["username"]
-    _check_response(r, f"Login failed for user {user}")
-    return r.json(), user
+        return {"Authorization": "Bearer " + self.access_token}
+
+    def obtain_jwt(self):
+        r = self.client.post("/v1/token/obtain/", json=self.creds)
+        if r.status_code == 401:
+            raise ValueError(
```
meksor, line 120: Maybe use self.auth_url here?

pyam/iiasa.py
```diff
@@ -119,27 +168,24 @@ def __init__(self, name=None, creds=None, auth_url=_AUTH_URL):
...
-        return {"Authorization": f"Bearer {self._token}"}
-
     @property
     @lru_cache()
     def _connection_map(self):
-        url = "/".join([self._auth_url, "applications"])
-        r = requests.get(url, headers=self._headers)
+        # TODO: application-list will be reimplemented in conjunction with ixmp-server
+        url = "/".join([self._auth_url, "legacy", "applications"])
+        r = requests.get(url, headers=self.auth())
```
meksor, line 188: I would use httpx here as well. You can pass the auth object when instantiating the client and it will be called for every request.

meksor (commented):
Lgtm from an integration standpoint, I made a few notes.

pyam/iiasa.py
danielhuppmann: If a (non-expert) user reads the error message and tries to put the auth-url into his/her browser, this returns an unhelpful "Not Found" page - so I would rather use a url that users can "use".

pyam/iiasa.py
danielhuppmann: Thanks, implemented!

PR #696, opened 2022-08-31
Local file cache

### LINKED ISSUE #681: Easy-to-use data download tools
Chris asked if we can add an easy-use tool to download emissions data (possibly checking if the data is already downloaded) from the AR6 database to Silicone, since I put in such a tool for SR1.5. [https://github.com/GranthamImperial/silicone/issues/148](https://github.com/GranthamImperial/silicone/issues/148). It avoids the need to look up what connections are used for this. 

It would be more logical for these tools to be part of the main pyam repository because 1) they will get better upkeep if you change the database behaviour 2) they are out-of-scope for Silicone and were just added for my convenience. 

Proposed resolutions: 
1) We write a utility function to do this for AR6
2) We transfer the code doing this for SR1.5 to pyam and replace the Silicone utility with a redirect call to the pyam tool in a later version. 

Are there any objections to doing this?

Rlamboll:
Done to resolve #681 

- [x] Tests Added
- [x] Documentation Added

```
```

# Description of PR


pyam/iiasa.py
```diff
@@ -558,7 +559,7 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
     ----------
     name : str
         A valid name of an IIASA scenario explorer instance,
-        see :attr:`pyam.iiasa.Connection.valid_connections`
+        see :attr:`pyam.iiasa.Connection().valid_connections`
```
danielhuppmann, line 562: Did you test whether the automated API documentation correctly links with the added brackets?

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+    save to that location. Does not check that the previously read version is a complete
+    instance of the database.
+
+    Parameters
+    ----------
+    file : str
+        The location to test for valid data and save the data if not up-to-date.
+    name : str
+        A valid name of an IIASA scenario explorer instance,
+        see :attr:`pyam.iiasa.Connection().valid_connections`
```
danielhuppmann, line 598: Same comment as above, does the API link correctly?

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
         Arguments for :meth:`pyam.iiasa.Connection.query`
     """
     return Connection(name, creds, base_url).query(default=default, meta=meta, **kwargs)
+
+
+def lazy_read_iiasa(
+    file, name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **kwargs
+):
+    """Check if the file in a given location is an up-to-date version of an IIASA
```
danielhuppmann, line 587: Please separate into a one-line description, an empty line, and the longer description.

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+    database. If so, load it. If not, load  data from IIASA scenario explorer and
```
danielhuppmann, line 588: I suggest to not mix the terms "database" and "Scenario Explorer". The Scenario Explorer is just a front-end to a database, let us (try to) consistently use the latter.

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+        A valid name of an IIASA scenario explorer instance,
+        see :attr:`pyam.iiasa.Connection().valid_connections`
+    default : bool, optional
+        Return *only* the default version of each scenario.
+        Any (`model`, `scenario`) without a default version is omitted.
+        If :obj:`False`, return all versions.
+    meta : bool or list of strings, optional
+        If `True`, include all meta categories & quantitative indicators
+        (or subset if list is given).
+    creds : str, :class:`pathlib.Path`, list-like, or dict, optional
```
danielhuppmann, line 606: Passing credentials directly is not supported any more.
```suggestion
    creds : str or :class:`pathlib.Path`, optional
```

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+        latest_update = np.nanmax(pd.to_datetime(version_info["update_date"]))
+        latest = pd.Series([latest_new, latest_update]).max()
+        if latest < date_set:
+            old_read = IamDataFrame(file)
+            if kwargs:
+                old_read = old_read.filter(**kwargs)
+            print("Database read from file")
+            return old_read
+        else:
+            print("Database out of date and will be re-downloaded")
```
danielhuppmann, line 631: Please use `logger.info()`

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+        date_set = pd.to_datetime(os.path.getmtime(file), unit="s")
+        version_info = Connection(name, creds, base_url).properties()
+        latest_new = np.nanmax(pd.to_datetime(version_info["create_date"]))
+        latest_update = np.nanmax(pd.to_datetime(version_info["update_date"]))
+        latest = pd.Series([latest_new, latest_update]).max()
+        if latest < date_set:
+            old_read = IamDataFrame(file)
+            if kwargs:
+                old_read = old_read.filter(**kwargs)
+            print("Database read from file")
```
danielhuppmann, line 628: Please use `logger.info()`

tests/test_iiasa.py
```diff
@@ -336,3 +336,22 @@ def test_query_empty_response(conn):
     """Check that querying with an empty response returns an empty IamDataFrame"""
     # solves https://github.com/IAMconsortium/pyam/issues/676
     assert conn.query(model="foo").empty
+
+
+def test_lazy_read(tmpdir):
+    null_file = str(tmpdir / "test_database.csv")
```
danielhuppmann, line 342: Suggest to call this `tmp_file` or similar - null implies no file...

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+        | See :class:`pyam.iiasa.Connection` for details.
+        | Use :meth:`pyam.iiasa.set_config` to set credentials
+          for accessing private/restricted Scenario Explorer instances.
+    base_url : str
+        Authentication server URL
+    kwargs
+        Arguments for :meth:`pyam.iiasa.Connection.query`
+    """
+    assert file[-4:] == ".csv", "We will only read and write to csv format."
+    if os.path.exists(file):
```
danielhuppmann, line 618: Why not cast `file` to **pathlib.Path** for consistency?

pyam/iiasa.py
```diff
@@ -578,3 +579,60 @@ def read_iiasa(name, default=True, meta=True, creds=None, base_url=_AUTH_URL, **
...
+          any public Scenario Explorer instances (i.e., with Guest login).
+        | See :class:`pyam.iiasa.Connection` for details.
+        | Use :meth:`pyam.iiasa.set_config` to set credentials
+          for accessing private/restricted Scenario Explorer instances.
+    base_url : str
+        Authentication server URL
+    kwargs
+        Arguments for :meth:`pyam.iiasa.Connection.query`
+    """
+    assert file[-4:] == ".csv", "We will only read and write to csv format."
```
danielhuppmann, line 617: I'm wondering if this is a good idea - writing to csv means ignoring the meta indicators, which are crucial for a lot of the relevant analysis.

danielhuppmann (changes requested):
Thanks for the PR @Rlamboll!

I'm worried about two issues:
1. By accepting only csv files for the local copy, you loose all meta indicators, which can be very relevant
2. The query is not actually stored, so if you do (as you do in the test
    ```python
    df = lazy_read_iiasa("some/file.csv", TEST_API, model="model_a")
    ```
    and then
    ```python
    df = lazy_read_iiasa("some/file.csv", TEST_API, model="model_b")
    ```
    you will get a different result for the second call depending on whether any data changed in the database in the meantime.

Rlamboll:
I've enabled it to read xls(x) files, but I don't think it's possible to do proper checks that the filters are applied in only a more restrictive sense without spending a significant period of time querying the database, which is what we're trying to avoid in the first place. I've expanded the documentation to make it clear that it's on the user to look out for this. This is consistent with how the code is currently used.

danielhuppmann (approved):
I'm not 100% convinced about this new feature because there might be confusion when a user queries for different filters than what is stored in the local cache. I'll approve, merge, and then create an issue for future improvements.

PR #701, opened 2022-09-12
Add xlsxwriter to improve to_excel performance

phackstock:

- [ ] ~~Tests Added (none needed except for possibly benchmarks)~~
- [ ] ~~Documentation Added~~

# Description of PR

Added `xlsxwriter` to the dependencies of `pyam`. `pandas` uses `xlsxwriter` over `openpyxl` if it's found. This means that if `xlsxwriter` is found on the system it is used without any changes required.
According to benchmarks (https://exchangetuts.com/python-fastest-way-to-write-pandas-dataframe-to-excel-on-multiple-sheets-1640154784194443), `xlsxwriter` is significantly faster than `openpyxl`.
Should I set up some benchmarks of our own to test it for pyam?

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2374,7 +2374,10 @@ class IamDataFrame(object):
         """
         close = False
         if not isinstance(excel_writer, pd.ExcelWriter):
-            excel_writer, close = pd.ExcelWriter(excel_writer), True
+            excel_writer, close = (
+                pd.ExcelWriter(excel_writer, engine="xlsxwriter"),
+                True,
+            )
         write_sheet(excel_writer, sheet_name, self.meta, index=True)
         if close:
             excel_writer.close()
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2374,10 +2374,8 @@ class IamDataFrame(object):
         """
         close = False
         if not isinstance(excel_writer, pd.ExcelWriter):
-            excel_writer, close = (
-                pd.ExcelWriter(excel_writer, engine="xlsxwriter"),
-                True,
-            )
+            excel_writer = pd.ExcelWriter(excel_writer, engine="xlsxwriter")
+            close = True
         write_sheet(excel_writer, sheet_name, self.meta, index=True)
         if close:
             excel_writer.close()
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -2343,7 +2343,7 @@ class IamDataFrame(object):
         close = False
         if not isinstance(excel_writer, pd.ExcelWriter):
             close = True
-            excel_writer = pd.ExcelWriter(excel_writer, engine="openpyxl")
+            excel_writer = pd.ExcelWriter(excel_writer, engine="xlsxwriter")
 
         # write data table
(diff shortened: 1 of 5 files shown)


danielhuppmann:
I just pip-installed pyam in a clean environment and xlsxwriter was installed automatically already. Not sure if it's necessary to add it as an explicit dependency?

phackstock:
Interesting, it's not a dependency of pandas (https://github.com/pandas-dev/pandas/blob/main/setup.cfg#L33). 
Maybe it was installed as a dependency by some other library. 
I just tried the same, install pyam from a clean virtual environment and it did not install it for me. Maybe it's platform dependent.

danielhuppmann:
You're right, looks like I was confused between two environments. Running some profiling now.

danielhuppmann (commented):
Note that **openpyxl** is hard-coded in line https://github.com/IAMconsortium/pyam/blob/759120f01453f2738df6d4126186bb0a21cab54b/pyam/core.py#L195

Also, I guess it's possible to remove **openpyxl** as a dependency?

I just ran some tests and can confirm that
- removing the hard-coded engine works whether or not **xlsxwriter** is installed
- performance is significantly better if **xlsxwriter** is installed, no need to specify it a engine explicitly (on a 30MB xlsx file, the resulting file is smaller by 50%, time is -45%, CPU is -80%)

danielhuppmann:
There is another usage of **pd.ExcelWriter** where the engine is not explicitly specified. Please make the two usages consistent.

https://github.com/IAMconsortium/pyam/blob/759120f01453f2738df6d4126186bb0a21cab54b/pyam/core.py#L2377

pyam/core.py
```diff
@@ -2374,7 +2374,10 @@ def export_meta(self, excel_writer, sheet_name="meta"):
         """
         close = False
         if not isinstance(excel_writer, pd.ExcelWriter):
-            excel_writer, close = pd.ExcelWriter(excel_writer), True
+            excel_writer, close = (
+                pd.ExcelWriter(excel_writer, engine="xlsxwriter"),
+                True,
+            )
```
danielhuppmann, line 2380:
```suggestion
            excel_writer = pd.ExcelWriter(excel_writer, engine="xlsxwriter")
            close = True
```

phackstock:
Updated the usage of xlsxwriter in `pd.ExcelWriter`. 
There's one more thing that I found: 

https://github.com/IAMconsortium/pyam/blob/main/pyam/utils.py#L135

is there a test for `utils.write_sheet`?

danielhuppmann:
> is there a test for utils.write_sheet?

It's tested implicitly via https://github.com/IAMconsortium/pyam/blob/759120f01453f2738df6d4126186bb0a21cab54b/tests/test_io.py#L56

The line you found was a hacky attempt by me to make xlsx files look "nice" by having useful column widths.

phackstock:
Ok, should I check if using xlsxwriter triggered the error?

danielhuppmann:
Let me give this a quick try, I think you have more urgent things on your to-do list...

phackstock:
True, just thought this might be a quick one to get off the list ...

danielhuppmann (approved):

PR #709, opened 2022-10-17
Ops nan hotfix

gidden:
# Overview

As currently implemented, `fillna=0` is not caught by the line `if fillna:`, and is subsequently ignored. This fixes that to allow NaNs to be filled with 0s. Without this fix, scenario data is silently thrown away when doing computations.

I have confirmed that the newly added test fails on current main.

- [x] Tests Added
~- [ ] Documentation Added~

### DIFF
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -104,7 +104,7 @@ def _op_data(df, name, method, axis, fillna=None, args=(), ignore_units=False, *
             )
 
     # merge all args and kwds that are based on `df._data` to apply fillna
-    if fillna is not None:  # can not check only if fillna: as 0s arent caught
+    if fillna is not None:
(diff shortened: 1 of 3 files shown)


pyam/_ops.py
```diff
@@ -104,7 +104,7 @@ def _op_data(df, name, method, axis, fillna=None, args=(), ignore_units=False, *
             )
 
     # merge all args and kwds that are based on `df._data` to apply fillna
-    if fillna:
+    if fillna is not None:  # can not check only if fillna: as 0s arent caught
```
danielhuppmann, line 107: I think that the extra comment makes the code more complicated without helping to understand the logic - and the test would catch it if anyone ever tried to simplify that line...
```suggestion
    if fillna is not None:
```

danielhuppmann (approved):
Looks good to me, one minor suggestion inline...

PR #720, opened 2022-12-21
Different approach to unstack in `timeseries()`

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

```
```

# Description of PR

I ran into a MemoryError yesterday when writing a large **IamDataFrame** to file so I tried to implement a more performant approach for pvioting to the wide IAMC format.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -782,7 +782,7 @@ class IamDataFrame(object):
         # calling `unstack` on the `s` directly runs into MemoryError for large datasets
         cols = [i for i in s.index.names if i != self.time_col]
         return pd.concat(
-            [_data.unstack(level=self.time_col) for i, _data in s.groupby(level=cols)]
+            [_data.unstack(level=self.time_col) for _, _data in s.groupby(level=cols)]
         ).rename_axis(None, axis=1)
 
     def reset_exclude(self):
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -402,9 +402,8 @@ class IamDataFrame(object):
         """
         if self._time is None:
             self._time = pd.Index(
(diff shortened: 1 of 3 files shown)


pyam/core.py
```diff
@@ -779,7 +779,11 @@ def timeseries(self, iamc_index=False):
                 )
             s = s.droplevel(self.extra_cols)
 
-        return s.unstack(level=self.time_col).rename_axis(None, axis=1)
+        # calling `unstack` on the `s` directly runs into MemoryError for large datasets
+        cols = [i for i in s.index.names if i != self.time_col]
+        return pd.concat(
+            [_data.unstack(level=self.time_col) for i, _data in s.groupby(level=cols)]
```
phackstock, line 785: One small improvement could be to use `_` instead of `i` in the loop to follow the convention for unused iteration variables. 
```suggestion
            [_data.unstack(level=self.time_col) for _, _data in s.groupby(level=cols)]
```

phackstock (approved):
Looks all good to me. 
One small comment in line below.
Two quick clarifying questions from my side.
Did you run some memory profiling to find out how much memory this new approach saves? 
Does this improvement also affect reading a data frame? I run into memory troubles regularly when trying to upload large data frames.

danielhuppmann:
Re questions by @phackstock:

I did not yet run any profiling and this change will (unfortunately) not change the performance issues when reading data.

phackstock:
Ah ok thanks for the clarifications @danielhuppmann. Still very good to improve data writing performance.

danielhuppmann:
Performance tests indicate that this was a very bad idea... But I pulled out some minor improvements into #723

PR #730, opened 2023-02-22
Refactor initialization for simpler maintenance

coroa:
# Please confirm that this PR has done the following:

- [ ] Tests Added
- [ ] ~Documentation Added~
- [ ] ~Name of contributors Added to AUTHORS.rst~
- [x] Description in RELEASE_NOTES.md Added

# Description of PR

~On-top of #729.~ Rebased to `main`.

Splits `utils.format_data` into 6 different functions:
1. `_convert_r_columns(df)` - Check and convert R-style year columns
2. `_knead_data(df, **kwargs)` - Replace, rename and concat according to user arguments
3. `_format_from_database(df)` - Post-process database results
4. `_intuit_column_groups(df, index)` - Check and categorise columns in dataframe
5. `_format_data_to_series(df, index)` - Convert a long or wide pandas dataframe to a series with the required columns
6. `_validate_complete_index(df)`

Functionally it should be neutral. Make sure to look at individual commits to follow the refactoring trail.

PR #731, opened 2023-02-22
Add fast-path to format data

coroa:
Co-authored-by: Matthew Gidden <matthew.gidden@gmail.com>

- [ ] Tests Added
- [ ] Documentation Added
- [ ] ~Name of contributors Added to AUTHORS.rst~
- [ ] ~Description in RELEASE_NOTES.md Added~

# Description of PR

Add a fast-path to `format_data` for initialization with a multi-index based Series or DataFrame that has all the required columns.

~I set the base branch for this PR to PR #730 to highlight the small additional changes necessary.~

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -249,10 +249,13 @@ def _format_from_legacy_database(df):
     return df
 
 
-def _intuit_column_groups(df, index):
+def _intuit_column_groups(df, index, include_index=False):
     """Check and categorise columns in dataframe"""
 
-    existing_cols = pd.Index(df.index.names).dropna()  # skip unnamed columns
+    if include_index:
+        existing_cols = pd.Index(df.index.names)
+    else:
+        existing_cols = pd.Index([])
     if isinstance(df, pd.Series):
         existing_cols = existing_cols.union(["value"])
     elif isinstance(df, pd.DataFrame):
@@ -352,7 +355,9 @@ def format_data(df, index, **kwargs):
 
     # Fast-pass if `df` has the index and required columns as a pd.MultiIndex
     if set(df.index.names) >= set(index) | set(REQUIRED_COLS) and not kwargs:
-        time_col, extra_cols, data_cols = _intuit_column_groups(df, index=index)
+        time_col, extra_cols, data_cols = _intuit_column_groups(
+            df, index=index, include_index=True
+        )
 
         if isinstance(df, pd.DataFrame):
             extra_cols_not_in_index = [c for c in extra_cols if c in df.columns]
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -367,10 +367,9 @@ def format_data(df, index, **kwargs):
             else:
                 df = df["value"]
 
-        df = df.reorder_levels(index + REQUIRED_COLS + extra_cols + [time_col])
+        df = df.reorder_levels(index + REQUIRED_COLS + [time_col] + extra_cols).dropna()
 
     else:
-    
         if isinstance(df, pd.Series):
             if not df.name:
                 df = df.rename("value")
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -350,8 +350,8 @@ def _format_data_to_series(df, index):
 def format_data(df, index, **kwargs):
     """Convert a pandas.Dataframe or pandas.Series to the required format"""
 
+    # Fast-pass if `df` has the index and required columns as a pd.MultiIndex
     if set(df.index.names) >= set(index) | set(REQUIRED_COLS) and not kwargs:
-        # Let's try to cut corners here, it's our fast-path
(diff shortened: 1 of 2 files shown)


pyam/utils.py
```diff
@@ -252,41 +252,47 @@ def _format_from_database(df):
 def _intuit_column_groups(df, index):
     """Check and categorise columns in dataframe"""
 
+    existing_cols = pd.Index(df.index.names).dropna()  # skip unnamed columns
```
danielhuppmann, line 255: What's the use case for unnamed columns? My intuition would be that this is invalid and should raise an error, or not?

gidden:
Hey @coroa - I hope I didn't clobber your commits here by merging your previous PR first. Could you rebase this one on main and then I can provide a review? Thanks!

pyam/utils.py
coroa: Well, yes that could be better handled.

I added it during bugfix for the non-fast path case. Since the `reset_index` happens before calling to `_intuit_column_groups`, there will be an unnamed `RangeIndex`, ie. `df.index.names = [None]`.

In that case we just want to start with an empty index.

We could replace this by
```
existing_cols = pd.Index([]) if df.index.names == [None] else pd.Index(df.index.names)

**Review by gidden — APPROVED** [review]
Thanks for this @coroa - I might have a small preference on style (will add suggestion), but not blocking.

Is it possible to add a test for the untested code path?

File: pyam/utils.py [hunk 2]
```diff
@@ -344,29 +350,51 @@ def _format_data_to_series(df, index):
...
+        time_col, extra_cols, data_cols = _intuit_column_groups(df, index=index)
+
+        if isinstance(df, pd.DataFrame):
+            extra_cols_not_in_index = [c for c in extra_cols if c in df.columns]
+            if extra_cols_not_in_index:
+                df = df.set_index(extra_cols_not_in_index, append=True)
 
-    df = _convert_r_columns(df)
+            if data_cols:
+                df = df[data_cols]
```
gidden, line 363: slight preference for 
```suggestion
                df = (
                    df[data_cols]
                    .rename_axis(columns=time_col)
                    .stack()
                    .rename("value")
                    )
```

coroa: Sure, sure, but I'll not commit your suggestion online, since stickler won't like the hanging brace.

pyam/utils.py
danielhuppmann: Ok, let's leave as is and I'll try to write a few more tests for corner cases...

danielhuppmann (approved):
Looks good to me! If I understand it correctly, the way to access the fast-pass is to set the index before initialization, right?

pyam/utils.py
coroa: I was a bit unhappy with this and after some more thinking, i came to the conclusion it is easier to make this choice explicit by adding a `include_index` argument to `_intuit_column_groups`.

That's how it is now.

coroa:
> Looks good to me! If I understand it correctly, the way to access the fast-pass is to set the index before initialization, right?

Indeed. (or even better keep it as it is, ie never reset it :))

danielhuppmann:
Sorry, my earlier comment was badly phrased... What I meant was the following:

In #726, @gidden added an option `fast=False` to the IamDataFrame initialization to explicitly instruct pyam to use the fast-pass (skip some validations) - now, this is implicit. Which means the fast-pass will automatically be applied by any method using `_finalize(append=False)` (see [here](https://github.com/IAMconsortium/pyam/blob/9aa84d499900ab0d47da0a9e993a302e71483929/pyam/core.py#L225)) including aggregation and algebraic operations - but it is not possible to use the fast-pass when initializing from a file (because pandas reads a dataframe).

I think that this is perfectly fine behavior - just wanted to highlight this (or stand corrected if I'm on the wrong track).

Fine to merge (and maybe add a "force-fast-pass"-arg later). Thanks!

coroa:
> Sorry, my earlier comment was badly phrased... What I meant was the following:
> 
> In #726, @gidden added an option `fast=False` to the IamDataFrame initialization to explicitly instruct pyam to use the fast-pass (skip some validations) - now, this is implicit. Which means the fast-pass will automatically be applied by any method using `_finalize(append=False)` (see [here](https://github.com/IAMconsortium/pyam/blob/9aa84d499900ab0d47da0a9e993a302e71483929/pyam/core.py#L225)) including aggregation and algebraic operations - but it is not possible to use the fast-pass when initializing from a file (because pandas reads a dataframe).
> 
> I think that this is perfectly fine behavior - just wanted to highlight this (or stand corrected if I'm on the wrong track).
> 
> Fine to merge (and maybe add a "force-fast-pass"-arg later). Thanks!

You are spot on. The fast-path is not improving file read-in speed (as-is), but only data passing within pyam and pandas, where the index is preserved, like with the `__finalize__` calls you are highlighting.

PR #733, opened 2023-02-28
Switch IIASA-Connection `default` argument

danielhuppmann:

- ~Tests Added~
- [x] Documentation Added

# Description of PR

In anticipation of the release of the re-implemented **ixmp** package, this PR changes the API of the **iiasa.Connection** methods to resemble other filter options (and the forthcoming ixmp API).

Currently, a user could be confused that the query
```python
iiasa.Connection("integration-test").index(default=False)
```
will return a table of runs that have both *True* and *False* in the `is_default` column.

Going forward, a user has to do `is_default=None` to get *all scenarios* (default and non-default), while `is_default=True` will return only the default scenarios (as currently implemented by `default=True`).

The new ixmp package will also support `is_default=False` to return only the non-default scenarios - the current ixmp API does not support that option.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,18 @@
 # Next Release
 
+## API changes
+
+In anticipation of a change in the **ixmp** API, the argument `default` in the
+IIASA-Connection methods was renamed to `is_default` with the following behavior:
+
+- If `True`, return *only* the default version of a model/scenario.
+- If `False`, return all versions for each model/scenario *except the default* (not yet implemented).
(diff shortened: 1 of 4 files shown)


danielhuppmann:
Thanks for the clarification, @meksor, will update the PR

danielhuppmann:
After some thinking, I went back to simply renaming the kwarg from "default" to "default_only" for forward-compatibility (with a deprecation-warning until release 2.0). The "is_default"-option (True, False, None) will not be available in the current (soon-to-be-legacy) API, so it's not necessary to add that here.

PR #734, opened 2023-02-28
Validation for illegal column names in `data`

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

During review of #729, I noticed that there may be issues when initializing an IamDataFrame with an unnamed column (None or empty string) - both could give unexpected behavior. This PR adds these two corner cases to the validation routine.

### DIFF
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -252,7 +252,10 @@ def _format_from_legacy_database(df):
 def _intuit_column_groups(df, index):
     """Check and categorise columns in dataframe"""
 
-    # check that there is no column in the timeseries data with reserved names
+    if None in df.columns:
+        raise ValueError("Unnamed column in `data`: None")
+
+    # check that there is no column in the timeseries data with reserved/illegal names
     conflict_cols = [i for i in df.columns if i in ILLEGAL_COLS]
     if conflict_cols:
         _cols = f"'{', '.join(conflict_cols)}'"
--- a/pyam/utils.py
+++ b/pyam/utils.py
@@ -268,7 +268,8 @@ def _intuit_column_groups(df, index, include_index=False):
     # check that there is no column in the timeseries data with reserved/illegal names
     conflict_cols = [i for i in existing_cols if i in ILLEGAL_COLS]
     if conflict_cols:
-        _cols = f"'{', '.join(conflict_cols)}'"
+        sep = "', '"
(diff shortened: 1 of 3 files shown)


pyam/utils.py
```diff
@@ -262,12 +262,18 @@ def _intuit_column_groups(df, index, include_index=False):
         existing_cols = existing_cols.union(df.columns)
 
     # check that there is no column in the timeseries data with reserved names
+    if None in existing_cols:
+        raise ValueError("Unnamed column in `data`: None")
```
coroa, line 266: This could also be part of the `ILLEGAL_COLS`, i guess, but maybe the message is just clearer this way.

pyam/utils.py
```diff
@@ -262,12 +262,18 @@ def _intuit_column_groups(df, index, include_index=False):
...
+
+    # check that there is no column in the timeseries data with reserved/illegal names
     conflict_cols = [i for i in existing_cols if i in ILLEGAL_COLS]
     if conflict_cols:
-        msg = f"Column name {conflict_cols} is illegal for timeseries data.\n"
-        _args = ", ".join([f"{i}_1='{i}'" for i in conflict_cols])
-        msg += f"Use `IamDataFrame(..., {_args})` to rename at initialization."
-        raise ValueError(msg)
+        _cols = f"'{', '.join(conflict_cols)}'"
```
coroa, line 271: The quotes are slightly weird, since this would become
```
'data, meta'
```

I'd take away the quotes entirely or prefer
```
'data', 'meta'
```

coroa (approved):
LGTM. Only minor comments from my side. Feel free to merge.

pyam/utils.py
danielhuppmann: Right, the reason why this is not part `ILLEGAL_COLS` is that we have an easy solution for string-column-names, which is part of the error message - but this doesn't work (easily) for `None`.

pyam/utils.py
danielhuppmann: Thanks for catching, I fixed that. The reason why I added the quotes is that if you have `data` and an empty string as column names, it would look like

> Illegal columns for timeseries data: , data

whereas with quotes

> Illegal columns for timeseries data: '', 'data'

(Before it was formatted as a list, not pretty).

PR #759, opened 2023-07-05
Excise "exclude" column from meta and add a own attribute

### LINKED ISSUE #755: Refactor the "exclude" feature
All validation methods have a flag "exclude_on_fail" feature to easily track which scenarios fail one or several validation steps. This is implemented via an "exclude" column in the meta indicators.

However, this approach causes some hiccups in the integration with the ixmp database infrastructure - the "exclude" meta indicator is always added by pyam and then has to be removed before importing to the database (or can cause confusion by users).

My proposal: drop the "exclude" column from the "meta" dataframe and move it to an own attribute. in addition, extend the `filter()` method such that `df.filter(exclude=True)` continues to work as is.

Any thoughts @phackstock @gidden @byersiiasa?

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR excises the "exclude" column from the "meta" attribute to reduce confusion among users why an exclude column is added from a timeseries-only data file. Also, this currently requires some tinkering in the IIASA Scenario Explorer upload workflow.

The PR:
- creates a new attribute `exclude` and refactors all validation methods such that the argument `exclude_on_fail` changes this attribute
- extends the filter such that `filter(exclude=<value>)` works as expected
- adds "exclude" to the list of illegal column names for (timeseries) data and the meta attribute
- when importing an xlsx file created with pyam < 2.0 (which has an "exclude" column in "meta"), that column is moved to the new `exclude` attribute (with a log message)
- streamlines a number of tests and updates the documentation

fyi @gidden @byersiiasa @phackstock 

closes #755


danielhuppmann:
Responding to this comment by @byersiiasa 

> But would there be a method to edit the exclude attribute manually?
e.g., perhaps similar to how one might use set_meta('exclude'...)

You can currently do the following in the new API (in addition to the validation methods):

```python
df.exclude = False
```
to change the entire `exclude` pd.Series

```python
df.exclude[idx] = True
```
where `idx` can be an integer, list of integer, or a **pd.MultiIndex**.

So you could do the following
```
index = df.filter(...).index
df.exclude[index] = True
```

Is that good enough for your use case? Or are there other uses that you encountered?

byersiiasa:
Yes - that sounds good!

danielhuppmann:
Merging after discussions in the MESSAGE group meeting...

PR #763, opened 2023-08-03
Bugfix for incorrect dimension-accessors when initializing from pandas 

### LINKED ISSUE #762: Helpful accessors confounded by pandas regression
The test below fails and I can't see why. This is the underlying cause of https://github.com/iiasa/climate-assessment/pull/36 I think.

In short: if you subset an `IamDataFrame` and then create new instances within some loop (probably a bad pattern, but let's ignore that for now), the metadata is still based on the original data rather than the subset. In the example, below this means that it looks like a scenario provides a variable when it actually doesn't.

@phackstock cc @danielhuppmann in case it helps with your search for a cause and making sure the bug doesn't come back.

```python
def test_climate_assessment_bug():
    test_df = pd.DataFrame(
        [
            ["model_a", "scen_a", "World", "Primary Energy", "EJ/yr", 1, 6.0],
            ["model_a", "scen_a", "World", "Primary Energy|Coal", "EJ/yr", 0.5, 3],
            ["model_a", "scen_b", "World", "Primary Energy", "EJ/yr", 2, 7],
        ],
        columns=["model", "scenario", "region", "variable", "unit", 2020, 2030],
    )
    test_df = IamDataFrame(test_df)
    for (model, scen), df_scen in test_df.timeseries().groupby(["model", "scenario"]):
        if not (model == "model_a" and scen == "scen_b"):
            continue

        # Primary Energy|Coal isn't part of the dataset from which `df_scen_pyam` is initialised
        assert "Primary Energy|Coal" not in df_scen.index.get_level_values("variable").unique()

        df_scen_pyam = IamDataFrame(df_scen)
        # Yet Primary Energy|Coal appears in the model_a, scen_b subset's `.variable` attribute
        assert "Primary Energy|Coal" not in df_scen_pyam.variable

        # I think this is a Pandas bug
```

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR introduces an explicit removal of unused levels when initializing from a pandas object. This bug could lead to inconsistent return-values of the dimension-accessors.

closes #762

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -128,6 +128,17 @@ def test_init_df_from_timeseries(test_df):
     pd.testing.assert_frame_equal(df.timeseries(), test_df.timeseries())
 
 
+def test_init_df_from_timeseries_unused_levels(test_df):
+    # this test guards against regression for the bug
+    # reported in https://github.com/IAMconsortium/pyam/issues/762
+
+    for _, df in test_df.timeseries().groupby(["model", "scenario"]):
+        # we're only interested in the second model-scenario combination
+        pass
+
+    # pandas does not remove unused levels (here: "Primary Energy|Coal") in a groupby
+    assert IamDataFrame(df).variable == ["Primary Energy"]
+
 def test_init_df_with_extra_col(test_pd_df):
     tdf = test_pd_df.copy()
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -132,12 +132,14 @@ def test_init_df_from_timeseries_unused_levels(test_df):
     # this test guards against regression for the bug
     # reported in https://github.com/IAMconsortium/pyam/issues/762
 
-    for _, df in test_df.timeseries().groupby(["model", "scenario"]):
+    for (model, scenario), data in test_df.timeseries().groupby(["model", "scenario"]):
         # we're only interested in the second model-scenario combination
-        pass
+        if model == "model_a" and scenario == "scen_b":
+            df = IamDataFrame(data)
 
-    # pandas does not remove unused levels (here: "Primary Energy|Coal") in a groupby
-    assert IamDataFrame(df).variable == ["Primary Energy"]
+    # pandas 2.0 does not remove unused levels (here: "Primary Energy|Coal") in groupby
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -128,6 +128,18 @@ def test_init_df_from_timeseries(test_df):
...
+def test_init_df_from_timeseries_unused_levels(test_df):
+    # this test guards against regression for the bug
+    # reported in https://github.com/IAMconsortium/pyam/issues/762
+
+    for _, df in test_df.timeseries().groupby(["model", "scenario"]):
+        # we're only interested in the second model-scenario combination
+        pass
+
+    # pandas does not remove unused levels (here: "Primary Energy|Coal") in a groupby
+    assert IamDataFrame(df).variable == ["Primary Energy"]
```
znicholls, line 140:
```suggestion
    for (model, scenario), df in test_df.timeseries().groupby(["model", "scenario"]):
        # we're only interested in this particular model-scenario combination
        # because it doesn't have Primary Energy|Coal
        if (model == "model_a" and scenario == "scen_b"):
            # pandas does not remove unused levels (here: "Primary Energy|Coal") in a groupby
            assert IamDataFrame(df).variable == ["Primary Energy"]
```

I don't know if groupby guarantees order so I always get nervous assuming it does. I think this change is a bit more defensive but also fine to leave how it is if I'm jumping at shadows.

znicholls (approved):
lgtm, one suggestion for a more defensive way to write the test (not an essential change)

phackstock (approved):
Beautiful thanks a lot @danielhuppmann :tada:

tests/test_core.py
danielhuppmann: Thanks! Implemented your suggestion slightly differently to ensure that the test is actually triggered - your approach might end up never calling the assertion because it's inside the if-statement...

znicholls: Smart

coroa:
ok, thanks, indeed, that is the issue i encountered today.

i would have suggested to replace calls to `get_index_levels` method with its pandas version, ie something like:
```
def get_index_levels(index, level):
     return index.unique(level)
```

so they can decide what is the most efficient solution, but your fix is also fine.

Issue #768, opened 2023-08-12
Support logical-and in `require_data()`


danielhuppmann:
# Background

The method `reuqire_data()` applies a logical-or for each dimension, so a scenario is ok if any variable/region from a given list is present.

We should add a keyword argument "method" or "logical" that allow a user to select whether any or all elements should be present.

# Context

In @l-welder & @coroa's **pathways-ensemble-analysis** tool, the following code is used (simplified):

```python
def require_variables(df, list_of_variables):
    for v in list_of_variables:
        df.require_data(variable=v)
```

The following would be much more straightforward:

```python
df.require_data(variable=list_of_variables, how="all")
```

# Default usage

I think that `how="all"` is the more common use case, so we could set the default to `None` (for pyam ^2.0) with a warning that *all* is used by default. In pyam 3.0, we can change the default to *all* and remove the warning.

@phackstock

phackstock:
Is there really a logical _or_ being applied? I just ran the following to double check:
```python
import pandas as pd
from pyam import IamDataFrame, IAMC_IDX

df = IamDataFrame(
    pd.DataFrame(
        [
            ["model_a", "scen_a", "World", "Primary Energy", "EJ/yr", 1],
        ],
        columns=IAMC_IDX + [2020],
    )
)

print(df.require_data(variable=["Primary Energy", "Primary Energy|Gas"]))
# returns model_a, scen_a indicating that there are variables missing
```

and I get a return for `model_a`, `scen_a` which according to the docs (https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.require_data) means that for those indices the criteria are not satisfied.
Granted the output could be more precise, indicating that only `Primary Energy|Gas` is missing, but in principle the logical operation currently used is an _and_, right?

danielhuppmann:
To be more precise, indeed `all` items are required *per dimension*, so as you correctly point out, the method will fail if a variable from that list is not included.

This issue discusses how to deal with multiple dimensions.

```python
df.require_data(
    variable=["Primary Energy", "Primary Energy|Gas"]
    region=["Austria", "Germany"]
)
````
this will currently pass if there is (only) data of "Primary Energy" for Austria and "Primary Energy|Gas" for Germany. 

However, I think a more common use case would be to require that both variables exist for both regions...

phackstock:
Ah right, ok that makes sense. That would be a nice addition indeed. 
[...]

PR #771, opened 2023-08-21
Refactor to use validation module

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR does three things:
- move the utility-method `make_index()` to the utils module
- move the utility-method `_exclude_on_fail()` to a new validation module
- add tests to show that `require_data()` implements logical-and, i.e., all combinations across dimensions are required, (see #768

### DIFF
--- a/pyam/validation.py
+++ b/pyam/validation.py
@@ -6,7 +6,7 @@ from pyam.utils import make_index, s
 logger = logging.getLogger(__name__)
 
 
-def on_failed_validation(df, index):
+def exclude_on_fail(df, index):
     """Assign a selection of scenarios as `exclude: True`"""
 
     if not isinstance(index, pd.MultiIndex):
--- a/pyam/validation.py
+++ b/pyam/validation.py
@@ -1,0 +1,19 @@
+import logging
+import pandas as pd
+
+from pyam.utils import make_index, s
+
+logger = logging.getLogger(__name__)
+
+
+def on_failed_validation(df, index):
+    """Assign a selection of scenarios as `exclude: True`"""
+
+    if not isinstance(index, pd.MultiIndex):
+        index = make_index(index, cols=df.index.names)
+
+    df.exclude[index] = True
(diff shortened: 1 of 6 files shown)


tests/test_feature_validation.py
```diff
@@ -3,43 +3,61 @@
...
+    columns=IAMC_IDX + TEST_YEARS,
+)
 
 
 @pytest.mark.parametrize(
     "kwargs",
     (
         dict(),
         dict(variable="Primary Energy"),
+        dict(variable="Primary Energy", year=[2005, 2010]),
```
phackstock, line 26: Is there a specific reason you're using `dict()` over `{}`?
```suggestion
        {"variable": "Primary Energy", "year": [2005, 2010]},
```

pyam/validation.py
```diff
@@ -0,0 +1,19 @@
+import logging
+import pandas as pd
+
+from pyam.utils import make_index, s
```
phackstock, line 4: I'd suggest renaming the function `s` into something more descriptive.

phackstock (commented):
Looks good to me. Two small comments below.

tests/test_feature_validation.py
danielhuppmann: I think readability is better thanks to the syntax-highlighting...

pyam/validation.py
phackstock: Maybe `plural_s`? Although I typed that before I'd seen the use in a docstring. There it's quite readable, so ok from my side to keep it this way if we don't have any good ideas.

tests/test_feature_validation.py
phackstock: Fair

phackstock (approved):
Good to be merged from my side. 
If you find a good alternative name for the `s` function you can change it otherwise it's fine as well to keep it as is and merge.

PR #772, opened 2023-08-22
Show all missing rows for `require_data()`

### LINKED ISSUE #768: Support logical-and in `require_data()`

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR changes the returned dataframe of the `require_data()` method to not only show the failing scenarios (dataframe of model-scenario combinations), but a dataframe that has all missing combinations of elements.

So if *model_a* - *scen_a* does not have the variable*Emissions|CO2* at the global level, the method will return the following:

model | scenario | region | variable
------|------|------|------
model_a | scen_a | World | Emissions\|CO2

closes #768

### DIFF
--- a/tests/test_feature_validation.py
+++ b/tests/test_feature_validation.py
@@ -63,9 +63,9 @@ def test_require_data(test_df_year, kwargs, exclude_on_fail):
     pdt.assert_frame_equal(obs, exp)
 
     if exclude_on_fail:
-        list(df.exclude) == [False, True]
+        assert list(df.exclude) == [False, True]
     else:
-        list(df.exclude) == [False, False]
+        assert list(df.exclude) == [False, False]
 
 
 def test_require_variable_pass(test_df):
--- a/tests/test_feature_validation.py
+++ b/tests/test_feature_validation.py
@@ -51,7 +51,15 @@ def test_require_data(test_df_year, kwargs, exclude_on_fail):
     df = test_df_year.append(IamDataFrame(DATA_GAS))
 
     obs = df.require_data(**kwargs, exclude_on_fail=exclude_on_fail)
+
     exp = pd.DataFrame([["model_a", "scen_b"]], columns=["model", "scenario"])
+    # add parametrization-dependent columns to expected output
+    if kwargs["variable"] == "Primary Energy|Coal":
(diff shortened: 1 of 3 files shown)


tests/test_feature_validation.py
```diff
@@ -51,7 +51,15 @@ def test_require_data(test_df_year, kwargs, exclude_on_fail):
...
+    # add parametrization-dependent columns to expected output
+    if kwargs["variable"] == "Primary Energy|Coal":
+        exp["variable"] = ["Primary Energy|Coal"]
+    else:
+        exp["variable"] = ["Primary Energy"]
+        exp["year"] = [2010]
+
     pdt.assert_frame_equal(obs, exp)
 
     if exclude_on_fail:
```
phackstock, line 65: In the lines below: 

```python
    if exclude_on_fail:
        list(df.exclude) == [False, True]
    else:
        list(df.exclude) == [False, False]
```

I think there's an assertion missing

phackstock (changes requested):
One small comment other than that looks good to me.
I've tried to wrap my head around `require_data` in its entirety but wasn't quite able to. As long as it's working though that's fine.

tests/test_feature_validation.py
danielhuppmann: Good catch, thanks!

phackstock (approved):
Very nice, that should help a lot in finding missing data :+1:.
Good to merge from my side.

Issue #785, opened 2023-10-12
aggregate_time() issue when uodating to pyam 2.0


adrivinca:
Hi, I realized updating from pyam 1.7 to 2.0 a piece of code I used in sub-annual reporting fails

`rep_iamSE_sub.aggregate_time(vv, column="subannual", value="year", append=True)`

gives the following error, although the input iamDF has values. Also, the code of the function aggregate_time does not seem to have changed, so I wonder what I could do to use the function.
Any suggestions on how to solve this is very welcome. Since we instructed people in a workshop to use these functions. maybe pandas pivot_table has changed behaviour?

```
for vv in rep_iamSE_sub.variable:
    rep_iamSE_sub.aggregate_time(vv, column="subannual", value="year", append=True)
Traceback (most recent call last):

  Cell In[26], line 2
    rep_iamSE_sub.aggregate_time(vv, column="subannual", value="year", append=True)

  File ~\AppData\Local\anaconda3\envs\message_env\Lib\site-packages\pyam\core.py:1653 in aggregate_time
    _df = _aggregate_time(
![pyamaggregation py, line 185](https://github.com/IAMconsortium/pyam/assets/22025365/33ec2b39-6a44-4826-a138-3f14d3cce179)

  File ~\AppData\Local\anaconda3\envs\message_env\Lib\site-packages\pyam\aggregation.py:186 in _aggregate_time
    .value.rename_axis(None, axis=1)

  File ~\AppData\Local\anaconda3\envs\message_env\Lib\site-packages\pandas\core\generic.py:6202 in __getattr__
    return object.__getattribute
![rep_iamSE_sub timeseries() head()](https://github.com/IAMconsortium/pyam/assets/22025365/31927fc7-255c-461a-906f-2ba2ccc63049)
__(self, name)

AttributeError: 'DataFrame' object has no attribute 'value'
```

I attach a spyder screenshot from `rep_iamSE_sub.timeseries().head()`
and an excel screenshot of the df after the pivot table, before running into the error (pyam/aggregation.py, line 185)
![rep_iamSE_sub timeseries() head()](https://github.com/IAMconsortium/pyam/assets/22025365/a4d42207-b939-4367-b2df-6703223712a8)
![pyamaggregation py, line 185](https://github.com/IAMconsortium/pyam/assets/22025365/9d5d186f-e0f7-460f-a726-9b246f054c12)

I will possibly attach a simple example if I have time, but would be great to get some insights in the meanwhile.
Thanks

[...]

PR #790, opened 2023-10-16
Set name of series returned by `aggregate_time()`

danielhuppmann:
# Description of PR

Quickfix for #785.

FYI @adrivinca


adrivinca:
Thanks for the fix. I am not completely clear on what changed in the new version and caused the error. However the remaining questions are:
- why this issue is not tested? should you add test to avoid similar errors happen in the future?
- does this occurs with the `_aggregate*` ? they need to be tested otherwise the package is not stable

PR #792, opened 2023-10-17
Facilitate region-aggregation with inconsistent      model scenario region       time

### LINKED ISSUE #558: df.aggregate_region() when inconsistent index (e.g. years)
There are instances when doing weighted aggregation of  regions, when years between the variable and the weight variable may not be aligned or one variable has more years than the other.  One gets error message:
```
  File "c:\github\pyam\pyam\aggregation.py", line 216, in _agg_weight
    raise ValueError("Inconsistent index between variable and weight!")

ValueError: Inconsistent index between variable and weight!
```
which comes from [here](https://github.com/IAMconsortium/pyam/blob/6de47ccec5d477a1a35425cea557849df699238c/pyam/aggregation.py#L215):
```
    if not data.droplevel(["variable", "unit"]).index.equals(weight.index):
        raise ValueError("Inconsistent index between variable and weight!")
```

It would be nice if there's a way to improve the logic here - e.g. for example, the minimum requirement should be not that the indexes are equal, but that the index of the variable A being weighted, is 100% present within the weight variable B. Im rusty on set notation, but basically that index A is a subset of index B?

Any quick  suggestion  how this should be done and/or arguments against?

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR improves the handling of region-aggregation when variables and weights have an inconsistent index.

Now, if weights are missing, a ValueError is raised with (the head of) a list of missing rows. If variable data is missing, a warning is written to the log with (the head of) a list of missing rows.

closes #558

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 # Release v2.0.0
 
+- [#789](https://github.com/IAMconsortium/pyam/pull/789) Support region-aggregation with weights-index >> data-index
+
 ## Highlights
 
 - Use **ixmp4** as dependency for better integration with the IIASA Scenario Explorer database infrastructure 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 # Release v2.0.0
 
-- [#789](https://github.com/IAMconsortium/pyam/pull/789) Support region-aggregation with weights-index >> data-index
+- [#792](https://github.com/IAMconsortium/pyam/pull/792) Support region-aggregation with weights-index >> data-index
 
 ## Highlights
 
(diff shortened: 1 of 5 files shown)


danielhuppmann:
I vaguely recall that @orichters expressed interest in a more flexible handling of the region-aggregation-with-weights...?

orichters:
> I vaguely recall that @orichters expressed interest in a more flexible handling of the region-aggregation-with-weights...?

I do not 😃. Maybe @Renato-Rodrigues, if it was someone from the REMIND team?

phackstock (approved):
Looks good to me. 
Can #754 be closed then since you're no longer raising an issue if weights are a superset of the data?

PR #797, opened 2023-11-03
Add `to_ixmp4()` method

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a method `to_ixmp4()` to easily save all scenarios in an IamDataFrame (including meta indicators) to an ixmp4 platform database instance.

The method name `to_ixmp4()` doesn't seem elegant, any alternative suggestions of comments?

### DIFF
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -34,7 +34,7 @@ def write_to_ixmp4(df, platform: ixmp4.Platform):
     for model, scenario in df.index:
         _df = df.filter(model=model, scenario=scenario)
 
-        run = platform.Run(model=model, scenario=scenario, version="new")
+        run = platform.runs.create(model=model, scenario=scenario)
         run.iamc.add(_df.data)
         run.meta = dict(_df.meta.iloc[0])
         run.set_as_default()
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -26,7 +26,7 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df):
         ("regions", df.region, RegionModel),
         ("units", df.unit, UnitModel),
     ]:
-        platform_values = platform.__getattribute__(dimension).tabulate().name.values
+        platform_values = getattr(platform, dimension).tabulate().name.values
         if missing := set(values).difference(platform_values):
             raise model.NotFound(
                 ", ".join(missing)
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -27,7 +27,7 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df):
         ("units", df.unit, UnitModel),
     ]:
         platform_values = platform.__getattribute__(dimension).tabulate().name.values
-        if missing := [i for i in values if i not in platform_values]:
+        if missing := set(values).difference(platform_values):
             raise model.NotFound(
                 ", ".join(missing)
                 + f". Use `Platform.{dimension}.create()` to add the missing {dimension}."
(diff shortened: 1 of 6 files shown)


danielhuppmann:
Nice catch, @phackstock. A roll-back is not possible in ixmp4 (because the creation of a run and adding of a run are separate actions), and there is no possibility (yet) to delete a run., but I'll add a to-do.

phackstock:
Ah I see, then the quick fix might be to still use a try except and delete the whole run if we encounter an error. Just to avoid having half finished ghost runs.

pyam/ixmp4.py
```diff
@@ -0,0 +1,42 @@
...
+        platform = ixmp4.Platform(platform)
+
+    # TODO: implement a try-except to roll back changes if any error writing to platform
+    # depends on https://github.com/iiasa/ixmp4/issues/29
+    # quickfix: ensure that units and regions exist before writing
+    for dimension, values, model in [
+        ("regions", df.region, RegionModel),
+        ("units", df.unit, UnitModel),
+    ]:
+        platform_values = platform.__getattribute__(dimension).tabulate().name.values
```
phackstock, line 29: I think using `.getattr` is suggested to use instead of the "dunder" method:
```suggestion
        platform_values = platform.getattr(dimension).tabulate().name.values
```

pyam/ixmp4.py
```diff
@@ -0,0 +1,42 @@
...
+        if missing := [i for i in values if i not in platform_values]:
```
phackstock, line 30: I prefer using an expressive name for the iterator:
```suggestion
        if missing := [value for value in values if i not in platform_values]:
```
or you could solve this directly using sets for values and platform_values

tests/test_ixmp4.py
```diff
@@ -0,0 +1,35 @@
...
+    platform.regions.create(name="World", hierarchy="common")
+    platform.units.create(name="EJ/yr")
+
+    if test_df.time_domain == "year":
+        test_df.to_ixmp4(platform=platform)
+    else:
+        with pytest.raises(NotImplementedError):
+            test_df.to_ixmp4(platform=platform)
+
+    # TODO add test for reading data from ixmp4 platform
```
phackstock, line 35: Maybe remove here and turn into an issue instead?

tests/test_ixmp4.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_ixmp4_integration(test_df):
+    """Write an IamDataFrame to the platform"""
+    platform = Platform(_backend=SqliteTestBackend())
+    platform.regions.create(name="World", hierarchy="common")
+    platform.units.create(name="EJ/yr")
+
+    if test_df.time_domain == "year":
+        test_df.to_ixmp4(platform=platform)
```
phackstock, line 30: Do you have a way of verifying that the data made it into the database correctly?

phackstock (commented):
Looks good to me in principle. 
Some minor comments and questions in line.

tests/test_ixmp4.py
danielhuppmann: This will be added in a follow-up PR adding `read_ixmp4()`, but this depends on ixmp4 v0.7.

pyam/ixmp4.py
danielhuppmann: Done.

tests/test_ixmp4.py
phackstock: Alright, understood

pyam/ixmp4.py
phackstock: Ah sorry I got the suggestion wrong, it should have been
```python
getattr(platform, dimension).tabulate()...
```

phackstock (approved):
Thanks for implementing my comments, good to merge from my side.

PR #800, opened 2023-11-18
Avoid misleading attribute-reference in iiasa.Connection docstring

danielhuppmann:
# Description of PR

Per an observation by @Rlamboll, the current docstring of the **iiasa.Connection** class is misleading because it states 

> See [pyam.iiasa.Connection.valid_connections](https://pyam-iamc.readthedocs.io/en/latest/api/iiasa.html#pyam.iiasa.Connection.valid_connections) for a list of available APIs.

when a user actually has to do `pyam.iiasa.Connection().valid_connections`.

This PR changes the docstring to only show the attribute-name in the docstring (instead of `pyam.iiasa...`) while maintaining a functioning link to the docs of that attribute.

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -139,8 +139,8 @@ class Connection(object):
     ----------
     name : str, optional
         The name of a database API.
-        See :attr:`pyam.iiasa.Connection.valid_connections` for a list
-        of available APIs.
+        Use :attr:`valid_connections <pyam.iiasa.Connection.valid_connections>`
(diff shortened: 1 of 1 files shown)


Rlamboll:
This works, but `read_iiasa` and `lazy_read_iiasa` have basically identical docstrings with the same problem (and you aren't referring to attributes of that class there).

Rlamboll (approved):
I hope people will understand it better this way! Thanks!

PR #804, opened 2023-12-11
Extend the `validate()` signature

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR extends the signature of the `validate()` method to take filter-arguments directly instead of using a (nested) dictionary, so

```python
df.validate({"Primary Energy": {"up": x, year=y}})
```

can now be called as
```python
df.validate(variable="Primary Energy", year=y, upper=x)
```

This allows to validate only a selection of the datapoints, e.g., filtered by region or model/scenario name. The signature can also more easily be called from a (to-be-implemented) processor in the **nomenclature** package.

The option to use nested dicts is marked as deprecated with release 3.0.

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1819,9 +1819,10 @@ class IamDataFrame(object):
         -----
         The following arguments are available for filtering:
 
-         - 'meta' columns: filter by string value of that column
          - 'model', 'scenario', 'region', 'variable', 'unit':
            string or list of strings, where `*` can be used as a wildcard
+         - 'meta' columns: mapping of column name to allowed values
+         - 'exclude': values of :attr:`exclude`
          - 'index': list of model, scenario 2-tuples or :class:`pandas.MultiIndex`
          - 'level': the "depth" of entries in the variable column (number of '|')
            (excluding the strings given in the 'variable' argument)
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1047,30 +1047,49 @@ class IamDataFrame(object):
         # TODO: deprecated, remove for release >= 2.1
         raise DeprecationWarning("Use `df.require_data()` instead.")
 
-    def validate(self, criteria={}, exclude_on_fail=False):
-        """Validate scenarios using criteria on timeseries values
+    def validate(
+        self,
+        criteria: dict = None,
+        *,
+        upper: float = None,
+        lower: float = None,
+        exclude_on_fail: bool = False,
+        **kwargs,
+    ) -> pd.DataFrame:
+        """Validate scenarios using bounds on (filtered) timeseries 'data' values.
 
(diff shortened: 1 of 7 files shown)


tests/test_feature_validation.py
```diff
@@ -68,77 +68,166 @@ def test_require_data(test_df_year, kwargs, exclude_on_fail):
...
     # checking exclude on fail
-    obs = test_df.validate(
-        {"Primary Energy": {"up": 6, "year": 2010}}, exclude_on_fail=True
-    )
+    obs = test_df.validate(**args, exclude_on_fail=True)
     pdt.assert_frame_equal(obs, test_df.data[5:6].reset_index(drop=True))
     assert list(test_df.exclude) == [False, True]
 
 
+def test_validate_multiple_criteria(test_df):
```
phackstock, line 213: You could assert here in addition that the deprecation warning is issued.

pyam/validation.py
```diff
@@ -2,13 +2,47 @@
...
-def _validate(df, criteria, exclude_on_fail):
-    _df = _apply_criteria(df._data, criteria, in_range=False)
+def _validate(df, criteria, upper, lower, exclude_on_fail, **kwargs):
+    # TODO: argument `criteria` is deprecated, remove for release >= 3.0
+    if criteria is not None:
+        deprecation_warning(
+            "Use `upper`, `lower`, and filter-arguments instead.", "Argument `criteria`"
+        )
+        if upper or lower is not None and not kwargs.empty:
+            raise NotImplementedError(
```
phackstock, line 18: Seems that Codecov and I came to the same conclusion here :D

pyam/core.py
```diff
@@ -1047,30 +1047,49 @@ def require_variable(self, *args, **kwargs):
         # TODO: deprecated, remove for release >= 2.1
         raise DeprecationWarning("Use `df.require_data()` instead.")
 
-    def validate(self, criteria={}, exclude_on_fail=False):
-        """Validate scenarios using criteria on timeseries values
+    def validate(
+        self,
+        criteria: dict = None,
+        *,
+        upper: float = None,
```
phackstock, line 1054: If you want you could make this more explicit with 
```suggestion
        upper_bound: float = None,
```
but I'm totally fine with just `upper` as well.

pyam/core.py
```diff
@@ -1047,30 +1047,49 @@ def require_variable(self, *args, **kwargs):
...
+        lower: float = None,
```
phackstock, line 1055: Same as above comment 
```suggestion
        lower_bound: float = None,
```

danielhuppmann:
Thanks @phackstock - implemented your suggestion for `upper_bound` and `lower_bound`, but I don't think that checking deprecation-warnings is worth the effort...

PR #810, opened 2024-02-07
Update the docstring of the iiasa-connection credentials to ixmp4

danielhuppmann:
# Description of PR

@phackstock noticed that the iiasa.Connection docstring was not updated to reflect the new ixmp4-login method.


phackstock (approved):
Thanks for the quick fix.

PR #818, opened 2024-02-27
Use context manager for reading Excel file

### LINKED ISSUE #817: Pyam read Excel does not close file
While working together with @adrivinca we discovered a bug in how pyam reads Excel files. 
In https://github.com/IAMconsortium/pyam/blob/0762055f318c9adf8df0275e8c11de3544419bdd/pyam/utils.py#L102-L130

a `pd.ExcelFile` instance is created but never closed. This means that during runtime, Excel files that were opened to read cannot be deleted since they're never closed. 
When using a temporary directory that is cleaned up at some point, this causes an error on Windows systems since the file will still be open.

Closing the file properly might also significantly decrease pyam's memory hunger when reading Excel files.

phackstock:

- [ ] ~~Tests Added~~
- [ ] ~~Documentation Added~~
- [ ] ~~Name of contributors Added to AUTHORS.rst~~

# Description of PR

Closes #817.
Use a context manager to ensure that an Excel file is closed after reading.

FYI @adrivinca

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -25,12 +25,12 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} py${{ matrix.python-version }}
-  
+
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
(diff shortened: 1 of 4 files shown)


danielhuppmann (approved):
Looks good, thanks!

phackstock:
Pinned `ixmp4==0.6.0` after a test broke when using `ixmp4` version `0.7.0`.

PR #819, opened 2024-02-27
implementation for fast categorize

gidden:
This is only a prototype for now, but I hope shows how we could implement a fast categorization in line with previous work on `df.validate()`

- [x] ~Tests Added~ no user-facing api changes
- [x] ~Documentation Added~ no user-facing api changes

# Description of PR

This PR uses the machinery from `validate()` now in `categorize()` resulting in huge speedups


gidden:
Not clear to my why ruff CI is failing

glatterf42:
The error message seems instructive to me:
```python
pyam/core.py:1:1: I001 [*] Import block is un-sorted or un-formatted
pyam/core.py:69:29: F401 [*] `pyam.validation._apply_criteria` imported but unused
Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

We have recently introduced ruff in this repo, which can handle several nice things at once. For example, it sorts your import blocks and cleans up unused imports. All these fixes have been applied to the main branch, so if you rebase this PR on top of that, you'll automatically get them. If you don't want to do that for whatever reason, run `ruff check . --fix` in your local environment (or use `--diff` instead of `--fix` first to see what will happen) and commit the changes again :)

glatterf42:
If you've done that and the action is still failing, please compare your local version with the one used for the check. Per default, GHA will use the [latest pypi release](https://pypi.org/project/ruff/), so 0.3.2.

glatterf42 (changes requested):
Looks good to me in principle, but for this to be a proof of concept, I'd like to see a test case using the new behaviour and confirming that it works as intended.

Or am I misunderstanding this? The PR this one is following up on changed the call signature of `df.validate()`, but this one keeps `categorize()` the same, isn't it?

glatterf42 (approved):
Then the existing tests are enough for me :)

gidden:
Correct, all we do here is update the internals of `categorize()` to utilize another top-level API (`validate()`). I'd argue this is even more robust than depending on other internal-only functions. Maybe @danielhuppmann can quickly give a thumbs up?

danielhuppmann:
Will review tonight

danielhuppmann:
This PR does not actually implement the new signature of `validate()` using direct filters and upper/lower bound arguments, hence the deprecation warning (inherited from `validate()`) without being able to use the new signature would be probably very confusing to users. So I fully implemented this in a new branch building on the prototype by @gidden..

Closing in favor of #837

gidden:
As an FYI, I was utilizing this prototype including the filters and upper/lower bound arguments. Code snippets that I was using are here:

in a `config.yaml`
```
  categorize:
    name: foo
    label: Medium Risk
    color: xkcd:baby blue
  criteria:
    upper_bound: 1000000
    lower_bound: 800000
    year: 2100
    region: World
 ```
 
 functioning code
```
        for key, data in self.definitions.items():
            if "categorize" not in data:
                continue
            logger.info(f"Preparing categorization for {key}")

            args = data["categorize"]
            self.db.categorize(
                args["name"],
                args["label"],
                variable=data["variable"],
                **data["criteria"],
                **{k: v for k, v in args.items() if k not in ["name", "label"]},
            )
```

File: pyam/core.py [hunk 1]
```diff
@@ -940,18 +946,25 @@ def categorize(
...
             if arg:
                 run_control().update({kind: {name: {value: arg}}})
-        # find all data that matches categorization
-        rows = _apply_criteria(self._data, criteria, in_range=True, return_test="all")
-        idx = make_index(rows, cols=self.index.names)
 
-        if len(idx) == 0:
+        # find all data that matches categorization
+        # TODO: if validate returned an empty index, this check would be easier
+        not_valid = self.validate(criteria=criteria, **kwargs)
```
gidden, line 960: Here @danielhuppmann the validate kwargs like `upper_bound` and other filtering args are taken in

danielhuppmann:
Ok, right, the method works with the new signature, but it was not described in the docstring and there were no tests added in this PR - this is what I implemented in the new branch and PR.

PR #829, opened 2024-03-04
Add a `pyam.iiasa.platforms()` function

danielhuppmann:

- ~Tests Added~
- [x] Documentation Added

# Description of PR

This PR adds a method `pyam.iiasa.list_platforms()` as a shorthand for the ixmp4 function down in the config hierarchy, producing an output like:

```
IIASA platform      Access    Notice

public-test         public    This is a public ixmp4 test instance hosted by the ... 
ecemf               private   
genie               public    This is a development instance for the GENIE knowle... 
ssp-extensions      public    
ecemf-internal      private   

5 total 
```

I modified this from the corresponding method in ixmp4 showing name, access level and notice (instead of dsn), because this seems more relevant for a Python-API-user.

@meksor, if you agree, I can implement the corresponding change in ixmp4?

# Update

Renamed the function to `pyam.iiasa.connections()` to avoid naming conflict with the ixmp4 method.

### DIFF
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -46,7 +46,12 @@ DEFAULT_IIASA_CREDS = Path("~").expanduser() / ".local" / "pyam" / "iiasa.yaml"
 
 
 def platforms() -> None:
-    """Print all available ixmp4 platforms hosted by IIASA"""
+    """Print all available ixmp4 platforms hosted by IIASA
+
+    See Also
+    --------
+    ixmp4.conf.settings.manager.list_platforms
+    """
 
     _platforms = ixmp4.conf.settings.manager.list_platforms()
     utils.echo("IIASA platform".ljust(20) + "Access".ljust(10) + "Notice\n")
--- a/pyam/iiasa.py
+++ b/pyam/iiasa.py
@@ -46,6 +46,7 @@ DEFAULT_IIASA_CREDS = Path("~").expanduser() / ".local" / "pyam" / "iiasa.yaml"
 
 
 def list_platforms():
+    """Print all available ixmp4 platforms hosted by IIASA"""
 
     platforms = ixmp4.conf.settings.manager.list_platforms()
     utils.echo("IIASA platform".ljust(20) + "Access".ljust(10) + "Notice\n")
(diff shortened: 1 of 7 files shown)


meksor:
Hm it would be useful to see if one is connecting via rest or sqlalchemy, maybe just showing the protocol part of the dsn? 
Something like `dsn.split('://')[0]`

danielhuppmann:
> Hm it would be useful to see if one is connecting via rest or sqlalchemy, maybe just showing the protocol part of the dsn?

My suggestion only refers to the attributes when connecting via the manager - the toml would still have the dsn.

pyam/iiasa.py
```diff
@@ -197,6 +217,10 @@ def _connection_map(self):
     @lru_cache()
     def valid_connections(self):
         """Return available resources (database API connections)"""
+        logger.warning(
+            "IIASA is migrating to a database infrastructure using the ixmp4 package."
```
phackstock, line 221: This could make a user think that all current projects running on the old infrastructure will be migrated.

pyam/iiasa.py
```diff
@@ -35,13 +36,32 @@
...
 
 # path to local configuration settings
 DEFAULT_IIASA_CREDS = Path("~").expanduser() / ".local" / "pyam" / "iiasa.yaml"
 
 
+def list_platforms():
+    """Print all available ixmp4 platforms hosted by IIASA"""
+
+    platforms = ixmp4.conf.settings.manager.list_platforms()
+    utils.echo("IIASA platform".ljust(20) + "Access".ljust(10) + "Notice\n")
```
phackstock, line 52: Is there a specific reason to use the ixmp functions for displaying information other than the colors?

phackstock (approved):
Good to merge from my side, just two small comments

pyam/iiasa.py
phackstock: Alright, fine with me just wanted to know if I was missing something

pyam/iiasa.py
phackstock: Fair point.

meksor:
Ah I see, alright convinced!

PR #832, opened 2024-03-07
Add `test-platform` fixture for ixmp4 tests

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR implements a few improvements related to the tests of the ixmp4-integration:

- Add a `test_platform` fixture in conftest
- Add a test for reading default-only runs from ixmp4
- Drop the "version" column from the meta-indicators when writing to an ixmp4 database (to avoid having a version-meta-indicator and a version-attribute of a Run) and add a test

### DIFF
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -1,7 +1,11 @@
+import logging
+
 import ixmp4
 import pandas as pd
 from ixmp4.core.region import RegionModel
 from ixmp4.core.unit import UnitModel
+
+logger = logging.getLogger(__name__)
 
 
 def read_ixmp4(platform: ixmp4.Platform | str, default_only: bool = True):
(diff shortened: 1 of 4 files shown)


pyam/ixmp4.py
```diff
@@ -70,10 +74,21 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df):
...
 
+    # The "version" meta-indicator, added when reading from an ixmp4 platform,
+    # should not be written to the platform
+    if "version" in df.meta.columns:
+        logger.warning(
+            "The `meta.version` column was dropped when writing to the ixmp4 platform."
+        )
+        meta = df.meta.drop(columns="version")
+    else:
+        meta = df.meta.copy()
```
phackstock, line 85: Do you need the explicit `copy()` here?

pyam/ixmp4.py
```diff
@@ -70,10 +74,21 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df):
                 f"{dimension}."
             )
 
+    # The "version" meta-indicator, added when reading from an ixmp4 platform,
+    # should not be written to the platform
+    if "version" in df.meta.columns:
+        logger.warning(
+            "The `meta.version` column was dropped when writing to the ixmp4 platform."
+        )
+        meta = df.meta.drop(columns="version")
```
phackstock, line 83: Could this result in an empty data frame except for the model and scenario index?

tests/test_ixmp4.py
```diff
@@ -1,59 +1,64 @@
...
+
+
+def test_ixmp4_reserved_columns(test_platform, test_df_year):
+    """Make sure that a 'version' column in `meta` is not written to the platform"""
+
+    # test writing to platform with a version-number as meta indicator
+    test_df_year.set_meta(1, "version")  # add version number added from ixmp4
+    test_df_year.to_ixmp4(platform=test_platform)
+
+    assert "version" not in test_platform.runs.get("model_a", "scen_a").meta
```
phackstock, line 64: Would the meta here just yield nothing?

phackstock (approved):
Looks good to me in principle. 
Just a couple of clarification questions below

tests/test_ixmp4.py
danielhuppmann: `meta` yields an empty dictionary here, added a test for it

pyam/ixmp4.py
danielhuppmann: yes, added a test-variant for this case

PR #833, opened 2024-03-09
Refactor to poetry groups instead of extras

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR does a few things:
- Refactor to use groups in pyproject.toml instead of extras
- Only import **wbdata** and **unfccc** when needed, skip tests if not installed
   This as a reaction to https://github.com/OliverSherouse/wbdata/issues/74


danielhuppmann:
Please review @glatterf42

.github/workflows/nightly.yml
```diff
@@ -68,13 +68,13 @@ jobs:
...
       if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-      run: poetry install --no-interaction --extras "docs optional_io_formats optional_plotting tests tutorials" --no-root
+      run: poetry install --no-interaction --with dev,docs,optional_io_formats,optional_plotting,tutorials,wbdata --no-root
 
     #------------------------
     #  install root project
     #------------------------
     - name: Install library
-      run: poetry install --no-interaction --extras "optional_io_formats optional_plotting tests tutorials" --only-root
+      run: poetry install --no-interaction --with dev,optional_io_formats,optional_plotting,tutorials,wbdata
```
glatterf42, line 77: Is there any specific reason to remove `--only-root`? 
The idea is that the step `Install dependencies` just before this one installs the dependencies if they have not been recovered from a cache. Thus, the only thing left to install in the `Install library` step is the project itself. When optional dependencies were defined as extras, they needed to be mentioned here because poetry removes dependencies from unmentioned extras upon `poetry install`. This is not true for groups, though, so we should be fine without `--with ...`. So this lines should be either
`run: poetry install --no-interaction`
or:
`run: poetry install --no-interaction --only-root`, I think.

.github/workflows/pytest-legacy.yml
```diff
@@ -60,13 +60,13 @@ jobs:
...
       if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-      run: poetry install --no-interaction --extras "optional_io_formats optional_plotting tests tutorials" --no-root
+      run: poetry install --no-interaction --with dev,optional_io_formats,optional_plotting,tutorials --no-root
 
     #------------------------
     #  install root project
     #------------------------
     - name: Install library
-      run: poetry install --no-interaction --extras "optional_io_formats optional_plotting tests tutorials" --only-root
+      run: poetry install --no-interaction --with dev,optional_io_formats,optional_plotting,tutorials
```
glatterf42, line 69: Same comment as in nightly workflow.

.github/workflows/build-docs.yml
```diff
@@ -43,12 +43,12 @@ jobs:
       with:
         path: .venv
         key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
-
+
```
glatterf42, line 46: These white spaces seem unnecessary, but not too tragic.

pyam/__init__.py
```diff
@@ -13,21 +11,21 @@
     require_variable,
     validate,
 )
-from pyam.datareader import read_worldbank  # noqa: F401  # noqa: F401
 from pyam.iiasa import lazy_read_iiasa, read_iiasa  # noqa: F401  # noqa: F401
 from pyam.ixmp4 import read_ixmp4  # noqa: F401
+from pyam.logging import configure_logging  # noqa: F401
 from pyam.run_control import run_control  # noqa: F401
-from pyam.statistics import Statistics
+from pyam.statistics import Statistics  # noqa: F401
```
glatterf42, line 18: In pyproject.toml, we have two lines that read:
```toml
[tool.ruff.lint.per-file-ignores]
"__init__.py" = ["F401"]
```
This should make it unnecessary to manually specify `# noqa: F401` in `__init__.py` files. Is there any specific reason to write it? A tool like mypy would not read this configuration setting, so requires its own setting, but ruff should not complain. If something like flake8 is still running in the background of your editor, feel free to switch it off since ruff provides the same functionality. If ruff itself is complaining, we should check its docs if the configuration syntax has changed for the version we're using.

pyam/worldbank.py
```diff
@@ -63,7 +57,7 @@ def read_worldbank(model="World Bank", scenario="WDI", **kwargs) -> IamDataFrame
         unit="n/a",
         region="country",
     )
-    # TODO use wb.get_indicators to retrieve corrent units (where available)
+    # TODO use wb.get_indicators to retrieve correct units (where available)
```
glatterf42, line 60: Thanks for clarifying this, I wasn't sure if this should be `current` or `correct` :)

glatterf42 (approved):
Looks good overall and tests are passing, some minor notes on questions in-line :)

pyam/__init__.py
danielhuppmann: Done

.github/workflows/build-docs.yml
danielhuppmann: Removed the whitespaces

.github/workflows/pytest-legacy.yml
danielhuppmann: Done

.github/workflows/nightly.yml
danielhuppmann: Done

PR #836, opened 2024-03-14
Building the docs fails with poetry 1.8.* on ReadTheDocs

danielhuppmann:
# Description of PR

Building the docs failed on ReadTheDocs for all of our projects, seems to e related to poetry 1.8.*. I'll make a corresponding PR on nomenclature - @glatterf42, can you please do the same on ixmp4?

FYI @phackstock

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -13,42 +13,42 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


danielhuppmann:
Ok, seems that it was an issue that the new poetry installs to the wrong virtual environment, see https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry.

Correctly rendered docs at https://pyam-iamc.readthedocs.io/en/devops-docs-poetry/

glatterf42 (approved):
Thanks, this looks good to me :)

PR #849, opened 2024-04-18
Filter/square brackets

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR fixes an issue where filtering by a string with square brackets is interpreted as regexp and doesn't work as expected.

### DIFF
--- a/pyam/str.py
+++ b/pyam/str.py
@@ -4,6 +4,8 @@ import numpy as np
 import pandas as pd
 from pandas.api.types import is_list_like
 
+REGEXP_CHARACTERS = r".+()[]{}|$"
+
 
 def concat_with_pipe(x, *args, cols=None):
     """Concatenate a list or pandas.Series using ``|``, drop None or numpy.nan"""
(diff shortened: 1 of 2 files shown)


pyam/str.py
```diff
@@ -4,6 +4,8 @@
 import pandas as pd
 from pandas.api.types import is_list_like
 
+REGEXP_CHARACTERS = r".+()[]{}|$"
```
phackstock, line 7: Not that I'd ever seen it but we can add the caret symbol as well `^`, especially since you have the dollar sign:
```suggestion
REGEXP_CHARACTERS = r".+()[]{}|$^"
```

phackstock (approved):
Looks good to me, one small suggestion but good to merge either way.

pyam/str.py
danielhuppmann: Thanks for this comment - went through the list of regex-characters at https://docs.python.org/3/library/re.html and also added `?`...

PR #850, opened 2024-04-19
Use recommended dictConfig for logging configuration

glatterf42:
This PR is in tandem with https://github.com/iiasa/ixmp4/pull/83, where you can find an extensive description of the underlying issue. For pyam, there is no `server.conf`, so these parts can be disregarded.

Does this PR need to be mentioned in the release notes?
Also, can we please have a quick patch release after merging this?


pyam/logging.py
```diff
@@ -11,7 +12,10 @@
 
 def configure_logging():
     """Configure logging"""
-    config.fileConfig(here / "logging.conf", disable_existing_loggers=False)
+    logging_config = here / "logging.json"
+    with open(logging_config, "r") as file:
```
danielhuppmann:
```suggestion
    with open(logging_config) as file:
```

pyam/logging.json
```diff
@@ -0,0 +1,23 @@
...
+      }
+    },
+    "handlers": {
+      "console": {
+        "class": "logging.StreamHandler",
+        "level": "NOTSET",
+        "formatter": "generic"
+      }
+    }
+  }
```
danielhuppmann:
```suggestion
  }

```

pyam/logging.py
```diff
@@ -11,7 +12,10 @@
...
+        config_dict = json.load(file)
+    config.dictConfig(config_dict)
```
danielhuppmann, line 18:
```suggestion
        config.dictConfig(json.load(file))
```

danielhuppmann (approved):
Thanks, a few minor suggestions inline...

pyam/logging.py
glatterf42: Ah, one more than anticipated ;)

danielhuppmann (approved):
Thanks @glatterf42, hope that this will resolve some downstream issues...

Issue #852, opened 2024-04-24
how to install pyam


AZFARHAD24511:
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyam
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\Admin\miniconda3\Lib\site-packages\pyam\__init__.py", line 9, in <module>
    from . initial_guesses import OrthogonalPolynomialInitialGuess
  File "C:\Users\Admin\miniconda3\Lib\site-packages\pyam\initial_guesses.py", line 10, in <module>
    import pycollocation
ModuleNotFoundError: No module named 'pycollocation'

(base) C:\Users\Admin>conda install -c conda-forge pyam
Channels:
 - conda-forge
 - defaults
Platform: win-64
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: C:\Users\Admin\miniconda3

  added / updated specs:
    - pyam

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    alembic-1.13.1             |     pyhd8ed1ab_1         155 KB  conda-forge
    annotated-types-0.6.0      |     pyhd8ed1ab_0          17 KB  conda-forge
    brotli-1.1.0               |       hcfcfb64_1          19 KB  conda-forge
    brotli-bin-1.1.0           |       hcfcfb64_1          20 KB  conda-forge
    ca-certificates-2024.2.2   |       h56e8100_0         152 KB  conda-forge
    certifi-2024.2.2           |     pyhd8ed1ab_0         157 KB  conda-forge
    colour-0.1.5               |     pyhd8ed1ab_1          21 KB  conda-forge
    conda-23.11.0              |  py311h1ea47a8_1         1.2 MB  conda-forge
    contourpy-1.2.1            |  py311h005e61a_0         202 KB  conda-forge
    cycler-0.12.1              |     pyhd8ed1ab_0          13 KB  conda-forge
    et_xmlfile-1.1.0           |     pyhd8ed1ab_0          10 KB  conda-forge
    fastapi-0.103.0            |     pyhd8ed1ab_0          55 KB  conda-forge
    furl-2.1.3                 |     pyhd8ed1ab_0          22 KB  conda-forge
    glib-2.78.4                |       hd77b12b_0         491 KB
    glib-tools-2.78.4          |       hd77b12b_0          91 KB
[...]

PR #853, opened 2024-04-25
Add installation-warning to Readme

danielhuppmann:
# Description of PR

This PR adds a warning to the Readme that the package is distributed under a different name on pypi, to avoid issues like #852.


phackstock (approved):
Looks good to me.

danielhuppmann:
I downgraded the Mac-OS-image to 13 as a quickfix because of failing actions with numpy, most likely to the switch to ARM64 processors.

PR #866, opened 2024-07-20
Validate timeseries data using explicit `value` and `rtol`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds keyword arguments to `validate()` to use *value* and *rtol* directly instead of *upper_bound* and *lower_bound*, in response to the discussion in https://github.com/IAMconsortium/common-definitions/pull/104. However, now that I've implemented this, I'm wondering whether it wouldn't make more sense to keep the pyam method lean and do the translation from *value* and *rtol* as part of a **nomenclature.DataValidator** class...

Any thoughts @phackstock?

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -1073,6 +1073,8 @@ class IamDataFrame:
         self,
         criteria: dict = None,
         *,
+        value: float = None,
+        rtol: float = None,
         upper_bound: float = None,
         lower_bound: float = None,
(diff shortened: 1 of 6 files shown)


danielhuppmann:
After some thoughts, I concluded that implementing this feature in nomenclature makes more sense.
Closing in favor of https://github.com/IAMconsortium/nomenclature/pull/370.

PR #868, opened 2024-07-26
Add option to filter by `measurand`

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a feature to support filtering by a "measurand", i.e., a tuple (or iterable of tuples) of variable and unit

See the related discussion at https://github.com/IAMconsortium/nomenclature/pull/360

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -545,13 +545,13 @@ def test_meta_idx(test_df):
     assert len(_meta_idx(test_df.data)) == 2
 
 
-def test_filter_by_bool(test_df):
+def test_filter_meta_by_bool(test_df):
     test_df.set_meta([True, False], name="meta_bool")
     obs = test_df.filter(meta_bool=True)
     assert obs.scenario == ["scen_a"]
 
 
-def test_filter_by_int(test_df):
+def test_filter_meta_by_int(test_df):
     test_df.set_meta([1, 2], name="meta_int")
     obs = test_df.filter(meta_int=[1, 3])
     assert obs.scenario == ["scen_a"]
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -447,6 +447,19 @@ def test_filter_variable_and_depth(test_df, filter_args):
     assert obs == []
 
 
+def test_filter_measurand_list(test_df):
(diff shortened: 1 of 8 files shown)


tests/test_core.py
```diff
@@ -447,6 +447,19 @@ def test_filter_variable_and_depth(test_df, filter_args):
...
 
+def test_filter_measurand_list(test_df):
+    data = test_df.data
+    data.loc[3, "variable"] = "foo"
+    data.loc[4, "unit"] = "bar"
+    df = IamDataFrame(data)
+
+    obs = df.filter(measurand=(("foo", "EJ/yr"), ("Primary Energy", "bar")))
+
+    assert obs.variable == ["foo", "Primary Energy"]
```
phackstock: Comparisons where unintentionally the order matters are always prone to breaking, I'd change it to:
```suggestion
    assert set(obs.variable) == {"foo", "Primary Energy"}
```

tests/test_core.py
```diff
@@ -447,6 +447,19 @@ def test_filter_variable_and_depth(test_df, filter_args):
...
+    assert obs.unit == ["EJ/yr", "bar"]
```
phackstock, line 459: Same as above, using a set means the test won't break if the order of units changes

```suggestion
    assert set(obs.unit) == {"EJ/yr", "bar"}
```

phackstock (approved):
One small comment, after that good to merge.

tests/test_core.py
```diff
@@ -431,14 +431,35 @@ def test_filter_empty_df():
...
 
+def test_filter_measurand_list(test_df):
+    data = test_df.data
+    data.loc[4, "variable"] = "foo"
+    data.loc[5, "unit"] = "bar"
+    df = IamDataFrame(data)
+
+    obs = df.filter(measurand=(("foo", "EJ/yr"), ("Primary Energy", "bar")))
+
+    assert obs.variable == ["Primary Energy", "foo"]
```
danielhuppmann, line 458:
```suggestion
    assert set(obs.variable) == {"Primary Energy", "foo"}
```

PR #870, opened 2024-08-06
Improve the docs of the `filter()` arguments

danielhuppmann:

- ~Tests Added~
- [x] Documentation Added

# Description of PR

This PR moves the documentation of the filter options/arguments to an own page of the docs rather that having them as part of the `slice()` docstring.

### DIFF
--- a/docs/api/filtering.rst
+++ b/docs/api/filtering.rst
@@ -1,17 +1,20 @@
-Filtering
-=========
+.. currentmodule:: pyam
 
-Filtering of an :class:`IamDataFrame`
--------------------------------------
+Filtering and slicing
+=====================
+
+Arguments for filtering an :class:`IamDataFrame`
+------------------------------------------------
 
 The |pyam| package provides several methods to filter an :class:`IamDataFrame` by its
-(timeseries) **data** or **meta** values. Read more about the `Data Model <data.html>`_.
+(timeseries) **data** or **meta** values. Read more about the `Data Model <data.html>`_
+that is implemented by an :class:`IamDataFrame`.
 
 The following arguments are available for filtering and can be combined as needed:
 
 Index
 ^^^^^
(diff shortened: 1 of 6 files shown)


docs/api/filtering.rst
```diff
@@ -1,12 +1,72 @@
...
+Methods for filtering and slicing an :class:`IamDataFrame`
+----------------------------------------------------------
+
+.. automethod:: pyam.IamDataFrame.filter
+   :noindex:
+
+.. automethod:: pyam.IamDataFrame.slice
+   :noindex:
+
+The **IamSlice** class
```
phackstock, line 56: Is that one necessary to include in the documentation for users?

docs/api/filtering.rst
```diff
@@ -1,12 +1,72 @@
...
+----------------------
+
+This class is an auxiliary feature to streamline the implementation of the
+:meth:`IamDataFrame.filter` method.
+
+.. autoclass:: pyam.slice.IamSlice
+   :members: dimensions, time, info
+
+Filtering using a proxy :class:`IamDataFrame`
```
phackstock, line 65: I've been trying to imagine how that would work and I think a short example would help here.

phackstock (commented):
Looks good to me in principle. I think it would be easier for the users if you had a short example of each filter. You could use the same data frame every time and filter in different ways.

danielhuppmann:
Thanks @phackstock - agree that we should extend the docs to include examples and I'll start an issue for it to be tackled before release 3.0, but for now, my aim was to get the docs out of `core.py`.

phackstock (approved):
Thanks for the clarification @danielhuppmann, in this case I'd say good to merge.

docs/api/filtering.rst
phackstock: Fair point

PR #877, opened 2024-08-26
Handling kwargs to be passed to pandas.ExcelFile

### LINKED ISSUE #876: kwargs are not passed to `pandas.ExcelFile` when reading Excel files with `IamDataFrame`
When reading an Excel file into a new `IamDataFrame` instance, `IamDataFrame.__init__` propagates keyword arguments to `read_file` and further to `read_pandas`. But in both of these methods, the Excel file is read by creating a `pandas.ExcelFile` instance with no kwargs. This results in an exception if you use kwargs that need to be passed to the `pandas.ExcelFile' init method.

In particular, this affects the `engine` kwarg. Pandas uses openpyxl as the Excel engine by default. But openpyxl is quite slow. For a moderately large IAMC-format Excel file in my current project (22 MB, 153k rows, 24 year columns) I see a 10x speed-up if I read it with `pandas.read_excel` and use `engine="calamine"` instead of the default, about 3 second load time vs. around 30 with openpyxl. 

The total times are similar if I read the file by passing it to `pyam.IamDataFrame.__init__` (which then uses openpyxl) vs first reading it with `pandas.read_excel` and `engine="calamine"` and then pass the resulting DataFrame to `pyam.IamDataFrame`. Which is a workable workaround, but not very elegant, and results in quite a bit of extra boilerplate if you also need to read metadata.

The fix should be relatively simple: In the two places where `pd.ExcelFile` is used ([core.py line 200](https://github.com/IAMconsortium/pyam/blob/bc56c53f5a835b857aadab21e4a3f8cb74d9dab6/pyam/core.py#L200) in `IamDataFrame._init`, and [utils.py line 103](https://github.com/IAMconsortium/pyam/blob/bc56c53f5a835b857aadab21e4a3f8cb74d9dab6/pyam/utils.py#L103), in `read_pandas`), search `kwargs` for any keywords that are accepted by `pandas.ExcelFile` (including `engine`), pop those off the kwargs and send them to `pd.ExcelFile` instead.

I'm working on an implementation of this locally that I can submit as a PR. But please let me know if you see a reason why kwargs should be handled in the original way instead. Which basically means any reason that the user should not be able to specify `engine` or `storage_options` when reading from an Excel file.

korsbakken:
Detects kwargs passed to `IamDataFrame.__init__` and to `read_pandas` that need to be passed to `pandas.ExcelFile.__init__` to be handled properly.

This PR closes #876 

- ~~[ ] Tests Added~~
- ~~[ ] Documentation Added~~
- ~~[ ] Name of contributors Added to AUTHORS.rst~~

**Tests:** I have run existing tests, and tested locally that the new code reads Excel files correctly with `engine="calamine"` as a kwarg to `pyam.IamDataFrame`. But a new test under `/tests/` that tests the added/fixed functionality (rather than just regression testing) would require that probably aren't intended to be part of the core requirements. If the lead developers would like me to do this, I would appreciate guidance on how to add a test that requires extra optional packages, without breaking the existing test suite.

**Documentation:** As far as I can tell, this is a bug fix, and shouldn't require new documentation. The one new helper function, `utils.get_excel_file_with_kwargs`, has a docstring that explains how it's used.

**Name in AUTHORS.rst:** This is a fairly trivial addition, and I don't need my name in the AUTHORS file for it. But I can add it in a update commit if you do want it there.

# Description of PR

The PR modifies lines in `pyam.core.IamDataFrame.__init__` and `pyam.utils.read_pandas` that create a `pandas.ExcelFile` instance to read an Excel file. The modifications ensure that keyword arguments that are recognized by `pandas.ExcelFile.__init__` are passed on to the `pandas.ExcelFile` instance and removed from calls to later methods (which would cause, e.g., `pandas.read_excel` to raise an exception).

The modifications add a new helper function `get_excel_file_with_kwargs` to `pyam.utils`. The function takes an excel file path and arbitrary kwargs, extracts the keyword arguments that are accepted by `pandas.ExcelFile.__init__`, creates a new `ExcelFile` instance with those keyword arguments, and returns it along with a dict of the other keyword arguments that were not used.

The new functionality makes it possible to use the keyword arguments `engine`, `storage_options` and `engine_kwargs` with `IamDataFrame.__init__` and `utils.read_pandas` when reading Excel files, which could cause an exception to be raised in the previous version.


danielhuppmann:
Thanks!

One suggestion for adding a test: add a few more keyword arguments, like header, usecols, skiprows, nrows, then make a copy of https://github.com/IAMconsortium/pyam/blob/bc56c53f5a835b857aadab21e4a3f8cb74d9dab6/tests/test_io.py#L116 with

```python
import_df = IamDataFrame(TEST_DATA_DIR / "test_df.xls", nrows=2)
assert_iamframe_equal(test_df_year.filter(scenario="scen_a"), import_df)
```

tests/test_io.py
```diff
@@ -118,6 +124,37 @@ def test_read_xls(test_df_year):
...
+
+
+@pytest.mark.skipif(not has_calamine, reason="Package 'python_calamine' not installed.")
+def test_read_xlsx_calamine(test_df_year):
+    # Test that an xlsx file is read correctly when using the calamine engine,
+    # and that excel kwargs such as `sheet_name` are still handled correctly
+    import_df = IamDataFrame(
+        TEST_DATA_DIR / "test_df.xlsx",
+        engine="calamine",
+        sheet_name="custom data sheet name",
```
danielhuppmann, line 153:
```suggestion
        # TODO this argument is currently ignored by pyam
        # see https://github.com/IAMconsortium/pyam/issues/878
        sheet_name="custom data sheet name",
```

pyam/utils.py
```diff
@@ -94,13 +94,30 @@ def write_sheet(writer, name, df, index=False):
...
 
+def get_excel_file_with_kwargs(path, **kwargs):
+    """Return a `pandas.ExcelFile` and a dict of unused kwargs.
+
+    When reading an Excel file, this function finds keyword arguments that
+    should be passed to `pandas.ExcelFile`, and returns a `pandas.ExcelFile`
+    instance along with the remaining keyword arguments (which presumably
+    will be used for other purposes by the calling function).
+    """
+    EXCEL_FILE_KWS = ('engine', 'storage_options', 'engine_kwargs')
```
danielhuppmann, line 105:
```suggestion
    EXCEL_FILE_KWS = ('engine', 'storage_options', 'engine_kwargs', 'header', 'usecols', 'skiprows', 'nrows')
```

danielhuppmann: And you could move that variable to the top of the file...

danielhuppmann (commented):
Looks good, but you need to add the pandas-arguments to make that test strategy work.

pyam/utils.py
korsbakken: Thanks, but that shouldn't be needed. I think you are confusing kwargs to `pandas.read_excel` and `pandas.ExcelFile` here. It's only the latter that needed to be addressed by this PR (and included in `EXCEL_FILE_KWS`). But since there was no pre-existing test for kwargs to `pandas.read_excel`, or correct handling of both simultaneously, I included that too.

`header`, `usecols`, `skiprows` and `nrows` are all arguments to `pandas.read_excel` and handled in that function, not passed on to `pandas.ExcelFile` (which does not accept those keywords). The code before this PR already supported keywords that are handled by `pandas.read_excel` and not passed to `pandas.ExcelFile`. The problem was the specific arguments that need to be passed to `pandas.ExcelFile` (engine, storage_options, engine_kwargs). `pyam.utils.read_pandas` passed its kwargs on to `pandas.read_excel`, but it first created a `pandas.ExcelFile` instance without any kwargs. pandas would then raise an exception, because it got both a pre-created `pandas.ExcelFile` instance and duplicate kwargs that are only accepted when `pandas.read_excel` is called with a Path and creates its own `pandas.ExcelFile` instance from that.

The `sheet_name` argument here is not ignored. If it were, the test would fail (I checked that first by leaving it out and confirming that the test indeed failed). I included it precisely to check that `pandas.read_excel` kwargs that are *not* passed to `pandas.ExcelFile` are not ignored. The xlsx file that is loaded in the test does not contain any sheet named `data` or otherwise recognized by pyam, so it relies on `sheet_name` being specified.

danielhuppmann: You are of course right, I was trying to review the PR in between two minutes and didn't take the necessary time, sorry. I overlooked that you had added a new test_df.xlsx file.

tests/test_io.py
danielhuppmann: Never mind...

danielhuppmann:
I guess that the pytest-legacy test (with pandas 2.1.2) is failing because `engine_kwargs={"data_only": False}` is not supported there? Can you add a pytest-skip for versions lower than 2.2 for that test? Then this PR is good to be merged!

korsbakken:
> I guess that the pytest-legacy test (with pandas 2.1.2) is failing because `engine_kwargs={"data_only": False}` is not supported there? Can you add a pytest-skip for versions lower than 2.2 for that test? Then this PR is good to be merged!

Yes. In fact, both the added tests fail with pandas < 2.2 for different reasons. `test_read_xlsx_calamine` fails because calamine wasn't added as an engine before 2.2.0. And `test_read_xlsx_kwargs` fails because pandas included hard-coded `engine_kwargs` that couldn't be overridden until the same version. I will add a warning about the latter if anyone tries to use `engine_kwargs` with pandas < 2.2 in the main code, and skip the tests in that case.

korsbakken:
I pushed a new version now, with a warning if anyone tries to use `engine_kwargs` with pandas version lower than 2.2.0, and skipping the tests if pandas is lower than that version.

In order to actually test the new functionality in CI, I also added a new optional dependency group `calamine` to pyproject.toml and updated the pytest workflows to include it. In order to make that work, I also had to update poetry.lock. If you would rather not include that, you can either reject the last two commits if that's possible, or just let me know and I will push a revert commit.

pyam/utils.py
```diff
@@ -94,13 +97,36 @@ def write_sheet(writer, name, df, index=False):
...
+    kwargs = kwargs.copy()
+    excel_file_kwargs = {
+        k: kwargs.pop(k) for k in EXCEL_FILE_KWS if k in kwargs
+    }
+    if packaging.version.parse(importlib.metadata.version("pandas")) \
+            < packaging.version.parse("2.2.0"):
```
danielhuppmann, line 114: Very nice idea to warn the users, but it's a bit too much to warn all users (where 99% won't even know what engine-kwargs are... Hence a proposal for a more targeted warning.
```suggestion
    # TODO remove when bumping minimum pandas dependency to >= 2.2
    if "engine_kwargs" in excel_file_kwargs and packaging.version.parse(
        importlib.metadata.version("pandas")
    ) < packaging.version.parse("2.2.0"):
```

korsbakken: > Very nice idea to warn the users, but it's a bit too much to warn all users (where 99% won't even know what engine-kwargs are)... Hence a proposal for a more targeted warning.

Yes, I totally agree. Thanks!

danielhuppmann (approved):
Thanks so much, very nice improvement!

PR #879, opened 2024-09-05
Add read_netcdf function

LinhHo:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a read_netcdf() function to read netcdf data (.nc) and return a :class:`IamDataFrame`. Note that this function currently does not support sub-annual data. The meta indicators have the exact indices from META_IDX (currently ‘model’ and ‘scenario’).


danielhuppmann (commented):
Thanks for this first stab, looks nice!

There are a few suggestions inline - most importantly, pyam can handle long- and wide-format natively, no need to do these transformations in your code.

One bigger suggestion is to place the entire new code in a new file `netcdf.py` (to avoid `core.py` becoming ever longer) and add
```python
from pyam.netcdf import read_netcdf
```
to `__init__.py`. 

Also, please take a look at pep8 and use ruff or black to format your code.
 
And one question: What's the difference between "standard" netcdf vs. the aggregated calliope netcdf?

tests/test_io.py
```diff
@@ -254,3 +255,13 @@ def test_io_datapackage(test_df, tmpdir):
     # read from csv assert that IamDataFrame instances are equal
     import_df = read_datapackage(file)
     assert_iamframe_equal(test_df, import_df)
+
+def test_io_netcdf(test_df, tmpdir):
+    # add column to `meta` and write to datapackage
+    test_df.set_meta(["a", "b"], "string")
+    
+    # read from csv assert that IamDataFrame instances are equal
```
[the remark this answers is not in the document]
LinhHo: Thanks for the information! When calling `test_df` to compare with dataframe from `read_netcdf`, an AttributeError occurs as below. As `test_df` is usually passed to functions such as `pyam.to_csv(file)`, I think another `else` clause is needed to get data directly from `test_df` for comparison in this case.
error_test_df

LinhHo:
Thanks a lot for your comments! To answer your question about aggregated Calliope netCDF, the current output from Calliope (“standard” netCDF) has five dimensions: nodes, techs, carriers, timesteps, and costs. They contain variables according to the model’s specifications, e.g., flow cap, flow in, flow out, flow export, etc. 

Aggregated Calliope netCDF (the input data for pyam.read_netcdf) are data that are aggregated or disintegrated from standard Calliope netCDF files into variables compatible with the IAMC standard, such as Primary Energy | Coal. They have four dimensions: model, scenario, region, and time. Unit is an attribute of each variable. In the future, Calliope will be updated to also produce aggregated netCDF as an optional output.

I will be on vacation in the next two weeks and will address any further comments from you when I'm back.
Best regards,
Linh

pyam/netcdf.py
```diff
@@ -0,0 +1,90 @@
...
+    import pip
+    pip.main(['install', '--user', 'xarray'])
+    import xarray as xr
+    
+from pyam.core import IamDataFrame
+from pyam.utils import META_IDX, IAMC_IDX
+
+def read_netcdf(path):
+    """Read timeseries or year-based data and meta-indicators from aggregated netCDF
+    (intermediate from Euro Calliope netCDF)
```
danielhuppmann, line 16: Good practice to keep the first sentence of a long docstring to one line. Also, this function should be generic, not limited to Calliope.
```suggestion
    """Read timeseries or year-based data and meta-indicators from netCDF
```

pyam/netcdf.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Parameters
+    ----------
+    path : string or path object to the netCDF file
```
danielhuppmann, line 20:
```suggestion
    path : :class:`pathlib.Path` or file-like object
        Scenario data file in netCDF format.
```

pyam/netcdf.py
```diff
@@ -0,0 +1,90 @@
...
+
+    output
```
danielhuppmann, line 22:
```suggestion
    Returns
```

pyam/netcdf.py
```diff
@@ -0,0 +1,90 @@
...
+                    _ds[_var]
+                    .to_dataframe()
+                    .reset_index(drop=False)
+                    .rename(columns={_var: "value"})
+                )
+
+            _tmp["variable"] = _ds[_var].long_name
+            _tmp["unit"] = _ds[_var].unit
+            _tmp = _tmp.reindex(columns=_list_cols)
+            _full_df = pd.concat([_full_df, _tmp])
```
danielhuppmann, line 80: Concatenating repeatedly is an expensive operation, enter to just add `_tmp` to a list and then do a concat after the loop.

danielhuppmann (changes requested):

danielhuppmann:
Sorry for the delay in the review, due to your and my vacation and then too many other ongoing activities...

danielhuppmann:
Pinging @coroa @Rlamboll @znicholls, @LinhHo started implementation of a `read_netcdf` method in pyam. Could you take a look at whether the nc file that she is using is structured in a way that is useful for the IAMC-format domain (including datetime support)?

znicholls:
There's lots of ways to do something like this with netCDF (similar to CSV, lots of options). The trick is more about just picking something that works.

If all you're doing is reading data, then it doesn't really matter. I would probably give the method a more specifc name. E.g. `read_netcdf_caliope`, because you don't have an equivalent `write_netcdf` function that makes it more obvious what format is expected.

There are probably some performance things you could improve in the PR, but if you test that it can read data with time columns and test that it can read data with year columns, you're probably good to go and you can fix performance later.

<details>
<summary>If you want to also support reading more arbitrary netCDF and writing netCDF, read this</summary>

The first thing to understand is that there isn't a standard netCDF format. People do crazy things, so supporting general netCDF isn't really a thing. If you have a `write_netcdf` method though, it's easier to be like, "We support reading netCDF written with `write_netcdf` or equivalents thereof.".

The thing we found difficult in scmdata was supporting arbitrary metadata columns when writing. That makes your life much trickier. If you don't care about capturing the meta table, that would make things simpler.

The other thing that can be tricky is the sparsity of data. If you want to store lots of model-scenario combinations, but not all models have run all scenarios, depending on how you format the data, you can end up with a lot of nans which can be annoying for memory or data storage/handling.

If you want the full details, our docs go into this sparsity issue here: https://scmdata.readthedocs.io/en/latest/notebooks/netcdf.html

In terms of how we implemented it, handling all our use cases was tricky (see https://github.com/openscm/scmdata/blob/main/src/scmdata/netcdf.py and https://github.com/openscm/scmdata/blob/main/src/scmdata/_xarray.py). You could do something much simpler if you make more restrictions than we did.
</details>

znicholls (commented):
See comment in main thread for the rest. This was the only thing that really stood out for me when looking at the code

pyam/netcdf.py
znicholls: I think you'll struggle to write a generic reader if you don't have an equivalent writer (more complete thoughts here: https://github.com/IAMconsortium/pyam/pull/879#issuecomment-2470035042)

tests/test_io.py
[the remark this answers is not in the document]
LinhHo: @danielhuppmann thanks for the comment! I've updated the test but there is still one problem when comparing the meta data. The error message says `AssertionError: DataFrame.iloc[:, 0] (column name="string") are different`, and more specifically  `nan != nan`.

Since np.nan is different from np.nan by default, I am not sure how to solve this problem. Could you help me with this please?

znichollscr:
> please see [this Jupyter notebook](https://github.com/LinhHo/pyam/blob/netcdf/docs/tutorials/read_netcdf.ipynb)

Looks very nice!

znicholls:
> please see [this Jupyter notebook](https://github.com/LinhHo/pyam/blob/netcdf/docs/tutorials/read_netcdf.ipynb)

Looks very nice!

gidden:
Awesome, thanks @LinhHo !!

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    # Check if the xarray dataset has the correct coordinates, then get column names
+    if is_year_based:
+        _list_cols = IAMC_IDX + ["year", "value"]
+    elif is_datetime:
+        _list_cols = IAMC_IDX + ["time", "value"]
+    else:
+        raise TypeError(
+            "Time coordinate needs to be integer (year-based) or datetime (timeseries), neither was given."
```
danielhuppmann, line 50:
```suggestion
            "Time coordinates can year (integer) or datetime format, found: " + ds.coords["time"]
```

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns
+    ----------
+    IamDataFrame with meta indicators if available
+
+    """
+
+    _ds = xr.open_dataset(path)
+    NETCDF_IDX = ["time", "model", "scenario", "region"]
+    _list_variables = [i for i in _ds.to_dict()["data_vars"].keys()]
+    _meta = []
```
danielhuppmann, line 32: Suggest to introduce `_meta` at the same place as `dfs`.

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+        _list_cols = IAMC_IDX + ["year", "value"]
+    elif is_datetime:
+        _list_cols = IAMC_IDX + ["time", "value"]
+    else:
+        raise TypeError(
+            "Time coordinate needs to be integer (year-based) or datetime (timeseries), neither was given."
+        )
+
+    # read `data` table
+    dfs = []
```
danielhuppmann, line 54: If you call the other internal object `_meta`, I suggest to call this `_data`.
```suggestion
    _data = []
```

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+                )
+
+            _tmp["variable"] = _ds[_var].long_name
+            _tmp["unit"] = _ds[_var].unit
+            dfs.append(_tmp)
+        else:
+            raise TypeError(
+                f"Cannot define {_var}, different indices from META_IDX and IAMC_IDX."
+            )
+    _full_df = pd.concat(dfs).reset_index(drop=True)
```
danielhuppmann, line 86: "Full df" sounds like there is an empty df somewhere, better to stick closer to the pyam-conventions.
```suggestion
    data = pd.concat(_data).reset_index(drop=True)
```

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+                    .reset_index(drop=False)
+                    .rename(columns={"time": "year", _var: "value"})
+                )
+            # if timeseries, keep the time coordinate as "time"
+            elif is_datetime:
+                _tmp = (
+                    _ds[_var]
+                    .to_dataframe()
+                    .reset_index(drop=False)
+                    .rename(columns={_var: "value"})
```
danielhuppmann, line 76: You could do this column-renaming after the concatenation, would probably be more efficient to do it once only (and not for every timeseries).

pyam/netcdf.py
```diff
@@ -0,0 +1,91 @@
...
+            _meta.append(_var)
+        elif set(_ds[_var].dims) == set(NETCDF_IDX):
+            # convert the data into the IamDataframe format
+            # if year-based data, get the time coordinate as "year"
+            if is_year_based:
+                _tmp = (
+                    _ds[_var]
+                    .to_dataframe()
+                    .reset_index(drop=False)
+                    .rename(columns={"time": "year", _var: "value"})
```
danielhuppmann, line 68: You could do this column-renaming after the concatenation, would probably be more efficient to do it once only (and not for every timeseries).

danielhuppmann (commented):
Looks good, a few technical improvement suggestions inline.

pyam/netcdf.py
```diff
@@ -0,0 +1,93 @@
...
+    HAS_XARRAY = True
+except ModuleNotFoundError:
+    xr = None
+    HAS_XARRAY = False
+from pyam.core import IamDataFrame
+from pyam.utils import META_IDX, IAMC_IDX
+
+
+def read_netcdf(path):
+    """Read timeseries or year-based data and meta-indicators from netCDF
```
danielhuppmann, line 17: Yearly data is also called timeseries, no need to differentiate here
```suggestion
    """Read timeseries data and meta indicators from a netCDF file
```

pyam/netcdf.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Parameters
+    ----------
+    path : :class:`pathlib.Path` or file-like object
+        Scenario data file in netCDF format.
+
+    Returns
+    ----------
+    IamDataFrame with meta indicators if available
```
danielhuppmann, line 26:
```suggestion
    :class:`IamDataFrame`
```

pyam/netcdf.py
```diff
@@ -0,0 +1,93 @@
...
+            _tmp["unit"] = _ds[_var].unit
+            dfs.append(_tmp)
+        else:
+            raise TypeError(
+                f"Cannot define {_var}, different indices from META_IDX and IAMC_IDX."
+            )
+    data = pd.concat(_data).reset_index(drop=True)
+
+    return IamDataFrame(
+        _full_df,
```
danielhuppmann: This probably went when committing my earlier suggestions
```suggestion
            pd.concat(_data).reset_index(drop=True),
```

tests/test_io.py
```diff
@@ -254,3 +255,9 @@ def test_io_datapackage(test_df, tmpdir):
     # read from csv assert that IamDataFrame instances are equal
     import_df = read_datapackage(file)
     assert_iamframe_equal(test_df, import_df)
+
+
+def test_io_netcdf(test_df_year):
+    file = Path(TEST_DATA_DIR / "test_df.nc")
+    import_df = read_netcdf(file)
+    assert_iamframe_equal(test_df_year, import_df)
```
danielhuppmann: Compare obs(erved) to exp(ected) outcome (but exp comes directly from the function).
```suggestion
    obs = read_netcdf(Path(TEST_DATA_DIR / "test_df.nc"))
    assert_iamframe_equal(obs, test_df_year)
```

pyam/netcdf.py
danielhuppmann: I still think that `_data` and `_meta` should be defined at the same place in the script.

tests/test_io.py
```diff
@@ -254,3 +255,11 @@ def test_io_datapackage(test_df, tmpdir):
...
     import_df = read_datapackage(file)
     assert_iamframe_equal(test_df, import_df)
+
+
+def test_io_netcdf(test_df_year):
+    file = Path(TEST_DATA_DIR / "test_df.nc")
+    import_df = read_netcdf(file)
+    # temporarily fix for np.nan being converted to 'nan' when saving netcdf file
+    import_df.meta.string[1] = np.nan
+    assert_iamframe_equal(test_df_year, import_df)
```
danielhuppmann, line 265: Suggest to use `obs` to be consistent with the general obs(erved) vs. exp(ected) structure of tests.
```suggestion
    obs = read_netcdf(TEST_DATA_DIR / "test_df.nc")
    # temporarily fix for np.nan being converted to 'nan' when saving netcdf file
    obs[1] = np.nan
    assert_iamframe_equal(obs, test_df_year)
```

danielhuppmann (commented):
Thanks, this is progressing nicely. I guess the remaining issue with the 'nan' as string  has to be fixed on the side of writing to netcdf?

danielhuppmann:
Looks like you found a very inconvenient corner case...
https://docs.xarray.dev/en/stable/user-guide/io.html
image

I suggest the following: replace all 'None's in string-columns in `meta` by "NaN" when writing to netcdf, and then replace them again when reading from file.

pyam/netcdf.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+    data = pd.concat(_data).reset_index(drop=True)
+    # if year-based data, get the time coordinate as "year"
+    # if timeseries, keep the time coordinate as "time"
+    if is_year_based:
+        data = data.rename(columns={"time": "year"})
+
+    return IamDataFrame(
+        data,
+        meta=_ds[_meta].to_dataframe() if _meta else None,
```
danielhuppmann, line 87: Actually, making the test pass is as easy as this.
```suggestion
        meta=_ds[_meta].to_dataframe().replace("nan", np.nan) if _meta else None,
```

danielhuppmann:
@LinhHo, I think with the quickfix I proposed above (replacing "nan"-strings by `np.nan` and a little bit of clean-up (like adding a readme, removing the print statement), this should be good to go, right?

pyam/netcdf.py
LinhHo: The column _var has a different variable name in every loop and thus, it needs to be changed into the same column name "value" before the concatenation.

pyam/netcdf.py
LinhHo: I moved the column-renaming for "year" after the concatenation. The _var to "value" renaming however needs to stay in the loop. Please see the comment above.

pyam/netcdf.py
danielhuppmann: Good point.

LinhHo:
@danielhuppmann Yes, I think so. Thanks a lot for all the comments! I have just updated the release note and cleaned up the code. Could you have an overall look again if we can finalise this PR, please?
By readme did you mean a documentation like this Jupyter notebook?
https://github.com/LinhHo/pyam/blob/netcdf/docs/tutorials/read_netcdf.ipynb

RELEASE_NOTES.md
```diff
@@ -26,6 +26,7 @@ Bumped minimum version of pandas and numpy to fit **ixmp4**'s requirement.
 
 ## Individual updates
 
+- (#879)[https://github.com/IAMconsortium/pyam/pull/879] Add read_netcdf() function
```
danielhuppmann, line 29: This has to go to the top of this file. These are the individual changes for a previous release, not the latest version.

danielhuppmann:
> readme did you mean a documentation like this Jupyter notebook?

Sorry, I meant release notes. Which you have already added but in the wrong place.

danielhuppmann:
Tests are failing because the lock file isn't up to date and the release-note ist still in the wrong place - I made https://github.com/LinhHo/pyam/pull/1 to fix this for you. Please review and approach that PR, then tests for this PR should pass.

danielhuppmann:
I took the liberty of pushing an update of the lock file directly to your branch

danielhuppmann:
Looks like tests are failing because matplotlib published a new release, so the plots generated in the tests do not exactly match the expected figures. I will take care of that next week.

danielhuppmann (approved):
This PR is ready to be merged once the matplotlib issue is fixed (in a separate PR).

Going forward, @LinhHo, it would be great if you could implement an equivalent method `IamDataFrame.to_netcdf()`, and I will do a PR adding this method to the docs.

danielhuppmann:
I realized that if I fix the matplotlib expected-plots in a separate branch, that would again create merge conflicts on poetry.lock. So I added the expected plots to this branch.

PR #893, opened 2024-12-09
Do not sort on initialization or append

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

So far, pyam sorted the timeseries data upon initialization and when appending data (including aggregation, and binary operations). This is probably a performance drag in scenario-processing where multiple aggregation steps are performed iteratively.

Also, there was an issue where the result of `timeseries()` did to always show the years in order.

This PR is the first of a sequence of planned improvements, removing the automated sorting and ensuring that timeseries in wide format are always returned in the right order.

As a next step, I will implement a `sort_data()` method (in case this is needed), remove automated sorting from other methods (e.g., `rename()`), and make sure that `compare()` and `check_*()` methods word on unsorted data.


byersiiasa:
This is a very welcome improvement!

Not using pyam much these days and I couldn't see the functionality of sort_data() quickly - but, if not included already, being able to choose the ordering of whether sort by model/scenario/variable could be useful downstream when plotting. IIRC there are now options to provide some manual sort argument when plotting, but this could be an alternative, or useful for other purposes!

danielhuppmann:
>  sort by model/scenario/variable could be useful downstream when plotting

Not quite what this PR had in mind, the plotting library already supports an `order` keyword argument for area and stacked bar charts.

PR #905, opened 2025-01-09
Pin Poetry to v1.8.5

dc-almeida:


danielhuppmann (approved):
Thank you @dc-almeida

PR #907, opened 2025-02-13
Add fast route for operations with non-SI units

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

```
```

# Description of PR

This PR enables binary operations with non-SI units without having to explicitly use the `ignore_units` argument.

### DIFF
--- a/pyam/_ops.py
+++ b/pyam/_ops.py
@@ -83,7 +83,7 @@ def _op_data(df, name, method, axis, fillna=None, args=(), ignore_units=False, *
         and fillna is None
         and len(_unit_kwds["a"]) == 1
         and len(_unit_kwds["b"]) == 1
-        and registry.Unit(_unit_kwds["a"][0]) == registry.Unit(_unit_kwds["b"][0])
+        and _unit_kwds["a"][0] == _unit_kwds["b"][0]
     ):
         # activate ignore-units feature
         ignore_units = _unit_kwds["a"][0] if method in [add, subtract] else ""
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -97,6 +97,21 @@ def test_add_variable(test_df_year, arg, df_func, fillna, ignore_units, append):
 
 
 @pytest.mark.parametrize("append", (False, True))
+def test_add_scenario_non_si_unit(test_df_year, append):
+    df = test_df_year.rename(unit={"EJ/yr": "foo"})
+    exp = df_ops_variable(operator.add, "Sum", unit="foo", meta=test_df_year.meta)
+    args = ("Primary Energy", "Primary Energy|Coal", "Sum")
+
+    if append:
+        obs = df.copy()
+        obs.add(*args, append=True)
+        assert_iamframe_equal(df.append(exp), obs)
+    else:
+        obs = df.add(*args)
+        assert_iamframe_equal(exp, obs)
+
+
+@pytest.mark.parametrize("append", (False, True))
 def test_add_scenario(test_df_year, append):
     """Verify that in-dataframe addition works on a custom axis (`scenario`)"""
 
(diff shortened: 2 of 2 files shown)


tests/test_ops.py
```diff
@@ -96,6 +101,21 @@ def test_add_variable(test_df_year, arg, df_func, fillna, ignore_units, append):
         assert_iamframe_equal(exp, obs)
 
 
+@pytest.mark.parametrize("append", (False, True))
+def test_add_scenario_non_si_unit(test_df_year, append):
+    df = test_df_year.rename(unit={"EJ/yr": "foo"})
+    exp = df_ops_variable(operator.add, "Sum", unit="foo", meta=test_df_year.meta)
+    args = ("Primary Energy", "Primary Energy|Coal", "Sum")
+
+    if append:
```
phackstock: I'm not a big fan of tests in general where there's an if inside the test that changes the test. I find that it makes it harder to read and understand. 
In this case I get the idea of wanting to avoid code duplication so perhaps this would make it more legible:
```python
if append:
    obs = df.copy()
    obs.add(*args, append=True)
    exp = df.append(exp)
else:
    obs = df.add(*args

assert_iamframe_equal(exp, obs)
```

this way the test ends with the same command in both ways and find it easier to read because my brain can go "ah ok we're simply preparing the data that we want to compare in two different ways but the comparison at the end is the same", as opposed to having to keep both branches of the if (admittedly small) in my mind.

tests/test_ops.py
```diff
@@ -159,6 +185,22 @@ def test_subtract_variable(test_df_year, arg, df_func, fillna, append, ignore_un
         assert_iamframe_equal(exp, test_df_year.subtract(*args, **kwds))
 
 
+@pytest.mark.parametrize("append", (False, True))
+def test_subtract_scenario_non_si_unit_unit(test_df_year, append):
+    df = test_df_year.rename(unit={"EJ/yr": "foo"})
+    exp = df_ops_variable(operator.sub, "Diff", unit="foo", meta=test_df_year.meta)
+    args = ("Primary Energy", "Primary Energy|Coal", "Diff")
+
+    if append:
```
phackstock: Same as for the comment in `test_add_scenario_non_si_unit`

tests/test_ops.py
```diff
@@ -288,6 +330,21 @@ def test_divide_variable(test_df_year, arg, df_func, fillna, append, ignore_unit
         assert_iamframe_equal(exp, test_df_year.divide(*args, **kwds))
 
 
+@pytest.mark.parametrize("append", (False, True))
+def test_divide_variable_non_si_unit_unit(test_df_year, append):
+    df = test_df_year.rename(unit={"EJ/yr": "foo"})
+    exp = df_ops_variable(operator.truediv, "Ratio", unit="", meta=test_df_year.meta)
+    args = ("Primary Energy", "Primary Energy|Coal", "Ratio")
+
+    if append:
```
phackstock: Same as for the comment for `test_add_scenario_non_si_unit`.

pyam/_ops.py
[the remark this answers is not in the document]
danielhuppmann: Done.

tests/test_ops.py
```diff
@@ -78,8 +78,13 @@ def test_add_variable(test_df_year, arg, df_func, fillna, ignore_units, append):
             inplace=True,
         )
 
-    unit = "EJ/yr" if ignore_units is False else ignore_units
-    exp = df_func(operator.add, "Sum", unit=unit, meta=test_df_year.meta)
+    if ignore_units:
```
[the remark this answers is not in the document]
danielhuppmann: I split out the four main tests (for the four binary operations) into two each, one "straight" and one with ignore-units.  Makes this PR quite dense, but definitely a good improvement for long-term maintenance.

PR #908, opened 2025-03-20
Bump to Poetry v2

dc-almeida:
Bumping pyam to Poetry v2


phackstock:
> The line causing the pytest-legacy CI to fail is `poetry add xlrd@2.0.1 --optional --lock`, as in v2, `--optional`: Add as an optional dependency to an extra, see documentation [here](https://python-poetry.org/docs/cli/#add). How should this be changed?

I think changing line 46 in pytest-legacy.yaml to:
```yaml
          poetry add xlrd@2.0.1 --lock --group optional_io_formats
```
should do the trick. Poetry treats optional dependencies now as groups with names. I looked it up and xlrd is part of the `optional_io_formats` so explicitly adding it to this group should do the trick.

pyproject.toml
phackstock, line 1: Please add `requires-poetry = ">=2.0,<3.0"` to make the update explicit (https://python-poetry.org/docs/pyproject/#requires-poetry).

phackstock (changes requested):
Some requested changes below.
I think the `--optional` fix should be a quick one.

phackstock (approved):
Thanks a lot @dc-almeida, good to be merged from my side.

danielhuppmann (approved):
Thanks @dc-almeida!

PR #912, opened 2025-05-28
Ixmp4/datetime

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR is a first step to support writing subannual (datetime) timeseries to an ixmp4 platform.

### DIFF
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -1,9 +1,13 @@
 import logging
 
 import ixmp4
+import numpy as np
 import pandas as pd
 from ixmp4.core.region import RegionModel
 from ixmp4.core.unit import UnitModel
+from ixmp4.data.abstract import DataPoint
+
+from pyam.index import get_index_levels_codes
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


phackstock (commented):
Good to be merged in principle, just a couple of comments:
* The release notes entry seems to be missing for now
* An entry to the docs page about firstly the datetime upload option and the datetime formatting rules would be good. 
* As codecov points out, a test with a 'time' dimension  with datetime and years combined would be good.

pyam/ixmp4.py
```diff
@@ -114,7 +123,15 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df):
         _df = df.filter(model=model, scenario=scenario)
 
         run = platform.runs.create(model=model, scenario=scenario)
-        run.iamc.add(_df.data)
+
+        if df.time_domain == "year":
+            run.iamc.add(_df.data)
+        else:
```
phackstock, line 129: I know this is technically not required since we only reach this point if we have datetime or year anyway but I always prefer explicit code:
```suggestion
        elif df.time_domain == "datetime":
```
since the code inside the if clause will only work for datetime

phackstock (approved):
One tiny, optional comment, good to be merged either way.

PR #952, opened 2026-02-02
Add a simple `aggregate_kyoto_gases()` method

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR adds a method to compute the aggregate Kyoto GHG from a set of species using any GWP metric supported by **iam-units**.

The method is similar to [add_gwp100_kyoto()](https://github.com/iiasa/climate-assessment/blob/640db7e885fcaee3c3d8642f0d97c11553b99406/src/climate_assessment/utils.py#L174) in the **climate-assessment** package.

~One issue is that there is a discrepancy in the unit used for HFC43-10 between ScenarioMIP data, **iam-units** and **common-definitions**, so this is commented out for now. See https://github.com/iiasa/emissions_harmonization_historical/issues/161~
*Fixed by implementing automatic handling*.

The units for each species and the minimum set of species (required) are hard-coded based on **common-definitions**, the returned variable has the GWP-metric as a suffix. This could be generalized in follow-up PR (or immediately if there are strong feelings).

Specific question to @jkikstra @znicholls @chrisroadmap: 
- Is this the right set of species?
- Should the list of minimum/required species be extended?

### DIFF
--- a/pyam/emissions.py
+++ b/pyam/emissions.py
@@ -1,0 +1,40 @@
+from pyam.exceptions import raise_data_error
+
+REQUIRED_SPECIES = ["Emissions|CO2", "Emissions|CH4", "Emissions|N2O"]
+
+SPECIES_UNIT_MAPPING = {
+    "Emissions|CO2": "Mt CO2/yr",
+    "Emissions|CH4": "Mt CH4/yr",
+    "Emissions|N2O": "kt N2O/yr",
+    "Emissions|SF6": "kt SF6/yr",
+    "Emissions|C2F6": "kt C2F6/yr",
+    "Emissions|C6F14": "kt C6F14/yr",
+    "Emissions|CF4": "kt CF4/yr",
+    "Emissions|HFC|HFC125": "kt HFC125/yr",
+    "Emissions|HFC|HFC134a": "kt HFC134a/yr",
+    "Emissions|HFC|HFC143a": "kt HFC143a/yr",
+    "Emissions|HFC|HFC227ea": "kt HFC227ea/yr",
+    "Emissions|HFC|HFC23": "kt HFC23/yr",
+    "Emissions|HFC|HFC32": "kt HFC32/yr",
+    # inconsistent notation between iam-units and common-definitions
+    #    "Emissions|HFC|HFC43-10": "kt HFC43-10/yr",
+}
+
+
+def aggregate_kyoto_gases(df: "IamDataFrame", metric: str):
+    """Internal implementation of the `aggregate_kyoto_gases` function"""
+
+    missing = df.require_data(variable=REQUIRED_SPECIES)
+    if missing is not None:
+        raise_data_error("Missing species for aggregation", missing)
+
+    df_list = list()
+    for species, unit in SPECIES_UNIT_MAPPING.items():
+        if species in df.variable:
+            df_list.append(
+                df.filter(variable=species).convert_unit(
+                    unit, "Mt CO2-equiv/yr", context=metric
+                )
+            )
+
+    return df_list
--- a/pyam/emissions.py
+++ b/pyam/emissions.py
@@ -1,22 +1,24 @@
+from pyam.aggregation import aggregate_data
 from pyam.exceptions import raise_data_error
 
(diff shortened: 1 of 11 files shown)


pyam/emissions.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def aggregate_kyoto_gases(df, metric: str):
+    """Internal implementation of the `aggregate_kyoto_gases` function"""
+
+    missing = df.require_data(variable=REQUIRED_SPECIES)
+    if missing is not None:
+        raise_data_error("Missing species for aggregation", missing)
+
+    df_list = list()
```
phackstock: This could be a list comprehension:
```suggestion
    return [df.filter(variable=species).convert_unit(unit, "Mt CO2-equiv/yr", context=metric) for species, unit in SPECIES_UNIT_MAPPING.items() if species in df.variable]
```

pyam/emissions.py
```diff
@@ -0,0 +1,50 @@
+from pyam.aggregation import aggregate_data
+from pyam.exceptions import raise_data_error
+
+REQUIRED_KYOTO_SPECIES = ["Emissions|CO2", "Emissions|CH4", "Emissions|N2O"]
+
+
+ALL_KYOTO_SPECIES = {
```
phackstock: Since we were talking about this a lot in the context of climate-processor, are there naming changes in this list between ar6 and the newer common-definitions variable names? If so, maybe specify in a comment what variable nomenclature this list is following.

pyam/emissions.py
```diff
@@ -0,0 +1,50 @@
...
+    "Emissions|cC4F8",
+}
+
+
+def aggregate_kyoto_ghg(df, metric: str, target_variable: str, target_unit: str):
+    """Internal implementation of the `aggregate_kyoto_ghg` function"""
+
+    _df = df.filter(variable=ALL_KYOTO_SPECIES)
+
+    missing = _df.require_data(variable=REQUIRED_KYOTO_SPECIES)
```
phackstock, line 54: Probably a bit of a niche case but does this requirement check if for each year where have the required species, we have ALL the required species?

pyam/emissions.py
```diff
@@ -0,0 +1,50 @@
...
+    "Emissions|C5F12",
+    "Emissions|C6F14",
+    "Emissions|C7F16",
+    "Emissions|C8F18",
+    "Emissions|CF4",
+    "Emissions|cC4F8",
+}
+
+
+def aggregate_kyoto_ghg(df, metric: str, target_variable: str, target_unit: str):
```
phackstock, line 45: Do you need `target_unit` as an input variable or will it always be "Mt CO2-equiv/yr"? In that case you could of course remove it as an input variable.

phackstock (approved):
Good to be merged from my side. Couple of questions below.

pyam/emissions.py
danielhuppmann: Replaced this by a `convert_unit` inplace anyway

pyam/emissions.py
danielhuppmann: There could be a use case where someone wants to use `Mt CO2e/yr` instead because `-` can cause issues for **pint**

pyam/emissions.py
danielhuppmann: Added references to common-definitions. There is now also some special handling for different spellings (HF43-10, HF4310 and HF4310mee), but I doubt that this needs to be explained in the docs.

pyam/emissions.py
danielhuppmann: Yeah, there could be models with incomplete or strange reporting, but to properly test that, one would have to check the required variables for each scenario separately (because what if one scenario only goes until 2050 and another one goes until 2100), and I think that this goes beyond the scope of pyam to test this (and write tests for it).

danielhuppmann:
Thank you @phackstock for the comments. Added a clear reference to common-definitions to the docs and added some special handling for HFC4310 where **gcages** and **common-definitions** are not aligned.

pyam/emissions.py
phackstock: Fair enough, flaky reporting should be a model issue, not pyam's.

phackstock:
Perfect, thanks, good to be merged from my side then.

PR #972, opened 2026-05-08
Fix ixmp4 platform lookup via `read_iiasa`

meksor:

- [X] Tests Added
- [ ] Documentation Added
  This is a bug fix so its not necessary.
- [X] Name of contributors Added to AUTHORS.rst

# Description of PR

Fixes the ixmp4 platform lookup via `read_iiasa` which currently only works per **display name** but should work per **slug** instead.


danielhuppmann (approved):
Looks good, thank you!

PR #974, opened 2026-05-12
Adjust SceSeAuth to work with new authentication classes

meksor:

- [ ] Tests Added
- [ ] Documentation Added

# Description of PR

Adjust SceSeAuth to work with new authentication classes, since it expects auth.access_token to be a string which is no longer true.


danielhuppmann:
Is this in addition to #972 or as a replacement?

meksor:
In addition. Another bug that surfaced while helping folks out with the new version.

danielhuppmann (approved):
Looks good, thank you!

PR #979, opened 2026-06-03
Support subannual (categorical) data with ixmp4.Platform

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR adds functionality to write and read subannual (categorical) datapoints to an **ixmp4.Platform**.

**pyam** currently translates all "None" to empty strings `''` in the index to allow filtering and avoid some hiccups when plotting. But this may need to be revised. I'm still thinking about this. But beyond the scope of this PR.

@meksor, I had to remove the subannual column before adding to the **Run** because otherwise the automatic type-inference wouldn't work. I think that this is ok, just highlighting in case I'm misunderstanding something or you have a different idea for implementation (or handling this in ixmp4).

### DIFF
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -179,7 +179,11 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df, checkpoint_message: str):
                     if "subannual" in _df_year.extra_cols:
                         # import the yearly vs. subannual-categorical data separately
                         # TODO revise handling of yearly values in mixed time domain
-                        run.iamc.add(_df_year.filter(subannual="").data)
+                        run.iamc.add(
+                            _df_year.filter(subannual="").data.drop(
+                                columns=["subannual"]
+                            )
+                        )
                         run.iamc.add(_df_year.filter(subannual="", keep=False).data)
                         # remove subannual column for type-infering in ixmp4
                         run.iamc.add(
--- a/pyam/ixmp4.py
+++ b/pyam/ixmp4.py
@@ -182,8 +182,11 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df, checkpoint_message: str):
                         run.iamc.add(_df_year.filter(subannual="").data)
                         run.iamc.add(_df_year.filter(subannual="", keep=False).data)
                         # remove subannual column for type-infering in ixmp4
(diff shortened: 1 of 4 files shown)


meksor:
Are empty string values the reason why you can't just give the entire DF to IXMP4 or is there another reason?

danielhuppmann:
> Are empty string values the reason why you can't just give the entire DF to IXMP4 or is there another reason?

You are right, I changed the behaviour of the "subannual" column to **None** for annual and datetime-datapoints, and now passing year and categorical as one data frame works like a charm! (But year vs. datetime still has to be separate, AFAIK). 

And I added a test to explicitly check the types of the datapoints saved to the database...

pyam/ixmp4.py
```diff
@@ -139,7 +141,7 @@ def write_to_ixmp4(platform: ixmp4.Platform | str, df, checkpoint_message: str):
     checkpoint_message : str
         The message for the ixmp4 checkpoint (similar to a commit message).
     """
-    if df.extra_cols:
+    if df.extra_cols and df.extra_cols != ["subannual"]:
         raise NotImplementedError(
             "Only data with standard IAMC columns can be written to an ixmp4 platform."
```
phackstock: I guess you'd have to update the error message, right?

phackstock (approved):
One small comment about an error message below, good to be merged without an additional review though.

pyam/ixmp4.py
danielhuppmann: Thanks, good catch!

meksor:
> You are right, I changed the behaviour of the "subannual" column to None for annual and datetime-datapoints, and now passing year and categorical as one data frame works like a charm! (But year vs. datetime still has to be separate, AFAIK).

Yeah they have to be separate columns. Ill see what I can do on the ixmp4 side to implement that as well, but for now this looks good to me.

meksor (approved):

danielhuppmann:
> Ill see what I can do on the ixmp4 side to implement that as well, but for now this looks good to me.

Thanks @meksor, but no worries, the current ixmp4 behaviour makes a lot of sense, I was just confused between None vs. whitespace in the test cases. I'll update the related pyam behaviour accordingly.

PR #981, opened 2026-06-10
Refactor implementation of `subannual` timeslices

danielhuppmann:

- [x] Tests Added
- ~Documentation Added~

# Description of PR

This PR follows up on an API change in #980, where the yearly aggregate is now identified as `None` rather than the string `year` in IamDataFrame instances with mixed time domain.

The following changes are implemented specifically:
- a warning when importing data that uses "year" or an empty string in the subannual column (legacy format)
- rename the method `aggregate_time()` to `aggregate_subannual()` for clarity (with deprecation warning)
- add an option
  ```python
  df.filter(subannual=True/False)
  ```
  to easily filter for all timeseries datapoints that are related to subannual categorical time resolution (or the inverse)
- rework the existing test to cover both the legacy and new formulation

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -5,6 +5,7 @@ import os
 import sys
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import Callable
 
 import ixmp4
 import numpy as np
@@ -31,7 +32,7 @@ from pyam.aggregation import (
     aggregate_data,
 )
 from pyam.compute import IamComputeAccessor
-from pyam.exceptions import format_log_message, raise_data_error
+from pyam.exceptions import format_log_message, raise_data_error, deprecation_warning
 from pyam.filter import (
     datetime_match,
     filter_by_col,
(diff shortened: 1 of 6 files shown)


danielhuppmann:
I pinky-promise to add a documentation page on all filter-options in a follow-up PR before the next major release

pyam/core.py
```diff
@@ -2059,6 +2073,12 @@ def _apply_filters(self, level=None, depth=None, **filters):  # noqa: C901
 
                 keep_col = datetime_match(self.get_data_column("time"), values)
 
+            elif col == "subannual" and isinstance(values, bool):
+                keep_col = filter_by_col(self._data, "subannual", "*", regexp=False)
+                # if filter is `subannual=False`, keep only where values are `None`
+                if not values:
```
phackstock: I find this version easier to read
```suggestion
                if values is False:
```

phackstock (approved):
Looks good to me 👍

pyam/core.py
danielhuppmann: Thanks!

PR #986, opened 2026-07-08
Extend `set_meta_from_data()` to apply on different column

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description of PR

This PR (finally) implements a feature that a user can check where some function applied to a timeseries in `data` holds and use the corresponding value in a different column for the meta value.

I've done this (manually) multiple times for getting the year where peak-warming (i.e., maximum of the temperature timeseries) is reached.

A few items to discuss:
- I chose "on" as argument name for the column *on* which to apply the "method", following the logic of `groupby(by)`... But I'm open to suggestions for alternative argument names
- The resulting meta-value is not necessarily unique, like when the temperature maximum is reached in two years. I (arbitrarily) use the first item and write a warning to the log.
- The implementation of `apply_method()` is very "hands-on", there may be more elegant approaches with lumpy...?
- For whatever reason, `df.groupby()[column].apply("min")` yields a different result from `df.groupby()[column].apply(np.min)` (because the latter evaluates element-wise).

# Example

To set the year where peak-temperature is reached as meta-indicator "Year of Peak Warming":

```python
df.set_meta_from_data(
    name="Year of Peak Warming",
    variable="Climate Assessment|Surface Temperature (GSAT)|Median",
    method="max",
    column="year",
    on="value",
)
```

### DIFF
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -941,15 +941,15 @@ class IamDataFrame:
         Parameters
         ----------
         name : str
-            Column name of the 'meta' table.
-        method : function, optional
+            Resulting column name in the 'meta' table.
+        method : function or str, optional
             Method for aggregation
             (e.g., :func:`numpy.max <numpy.ndarray.max>`);
             required if downselected data do not yield unique values.
         column : str, optional
             The column from `data` to be used to derive the indicator.
         on : str, optional
-            If given, apply the `method` on this column and use correspind value from
+            If given, apply the `method` on this column and use corresponding value from
             `column` as meta indicator.
         **kwargs
             Passed to :meth:`slice` for downselection of data.
--- a/pyam/core.py
+++ b/pyam/core.py
@@ -935,31 +935,82 @@ class IamDataFrame:
         self._new_meta_column(name)
         self.meta[name] = meta[name].combine_first(self.meta[name])
 
-    def set_meta_from_data(self, name, method=None, column="value", **kwargs):
+    def set_meta_from_data(self, name, method=None, column="value", on=None, **kwargs):
         """Add meta indicators from downselected timeseries data
 
         Parameters
         ----------
         name : str
-            Column name of the 'meta' table
+            Column name of the 'meta' table.
         method : function, optional
             Method for aggregation
             (e.g., :func:`numpy.max <numpy.ndarray.max>`);
-            required if downselected data do not yield unique values
+            required if downselected data do not yield unique values.
         column : str, optional
-            The column from `data` to be used to derive the indicator
+            The column from `data` to be used to derive the indicator.
+        on : str, optional
+            If given, apply the `method` on this column and use correspind value from
+            `column` as meta indicator.
         **kwargs
-            Passed to :meth:`slice` for downselected data
+            Passed to :meth:`slice` for downselection of data.
+
+        Raises
+        ------
+        ValueError
+            If the resulting meta-indicators are not unique for each element
+            of the :attr:`index`.
+
+        Examples
+        --------
+        A simple use case for this method is to compute a meta-indicator for the peak
+        (maximum) temperature from annual temperature timeseries data for each scenario:
+
+        .. code:: python
+
+           df.set_meta_from_data(
+               name="Climate Assessment|Peak Warming [°C]",
+               method="max",
+               variable="Climate Assessment|Surface Temperature",
+           )
+
+        Alternatively, the *method* can be applied on a column *on* and the
+        corresponding value from the *column* is set as meta indicator. This can be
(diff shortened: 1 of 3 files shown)


pyam/core.py
```diff
@@ -935,31 +935,91 @@ def set_meta(self, meta, name=None, index=None):  # noqa: C901
         self._new_meta_column(name)
         self.meta[name] = meta[name].combine_first(self.meta[name])
 
-    def set_meta_from_data(self, name, method=None, column="value", **kwargs):
+    def set_meta_from_data(self, name, method=None, column="value", on=None, **kwargs):
```
phackstock, line 949: For me as a user, type hints would help here so that I can get an easier idea of what the function can do, especially for `method`

pyam/core.py
```diff
@@ -935,31 +935,91 @@ def set_meta(self, meta, name=None, index=None):  # noqa: C901
...
         """Add meta indicators from downselected timeseries data
 
         Parameters
         ----------
         name : str
-            Column name of the 'meta' table
+            Column name of the 'meta' table.
```
phackstock: I would add something here to emphasize that this is the name of the resulting column:
```suggestion
            Resulting column name in the 'meta' table.
```

pyam/core.py
```diff
@@ -935,31 +935,91 @@ def set_meta(self, meta, name=None, index=None):  # noqa: C901
...
         method : function, optional
```
phackstock: In your example, you put a string which is contradicting the type hint here:

```suggestion
        method : function or str, optional
```

pyam/core.py
```diff
@@ -935,31 +935,91 @@ def set_meta(self, meta, name=None, index=None):  # noqa: C901
...
             Method for aggregation
             (e.g., :func:`numpy.max <numpy.ndarray.max>`);
-            required if downselected data do not yield unique values
+            required if downselected data do not yield unique values.
         column : str, optional
-            The column from `data` to be used to derive the indicator
+            The column from `data` to be used to derive the indicator.
+        on : str, optional
+            If given, apply the `method` on this column and use correspind value from
```
phackstock:
```suggestion
            If given, apply the `method` on this column and use corresponding value from
```

pyam/core.py
```diff
@@ -935,31 +935,91 @@ def set_meta(self, meta, name=None, index=None):  # noqa: C901
...
+
+            @staticmethod
+            def apply_method(x):
+                if callable(method):
+                    value = x[x[on] == method(x[on])][column].unique()
+                else:
+                    value = x[x[on] == x[on].apply(method)][column].unique()
+
+                if len(value) > 1:
+                    logger.warning(f"Non-unique result from {method} on column {on}.")
```
phackstock, line 1012: Should this just be a warning? I'd imagine that producing a non-unique value for a meta value can only be ill-conceived. We might want to raise an actual error here.

tests/test_feature_set_meta.py
```diff
@@ -158,3 +158,13 @@ def test_set_meta_from_data_nonunique(test_df):
...
+
+
+@pytest.mark.parametrize("method", ("min", np.min))
+def test_set_meta_from_data_method_on_other_column(test_df_year, method):
+    # get the variable that has the lowest data value for each scenario
+    test_df_year.set_meta_from_data("foo", method=method, column="variable", on="value")
+    exp = pd.Series(
+        data=["Primary Energy|Coal", "Primary Energy"], index=EXP_IDX, name="foo"
+    )
+    pdt.assert_series_equal(test_df_year["foo"], exp)
```
phackstock: I know that this implementation is working but for improved clarity, I'd suggest:
```suggestion
    pdt.assert_series_equal(test_df_year.meta["foo"], exp)
```

phackstock (changes requested):
Looks good to me. Just a few minor comments and questions below.
On the topic of using `on` for the name of the column that is used to populate the values in the meta table, I think using `on` is good. I always like taking design patterns, or variable names in this case, from other, established libraries.

pyam/core.py
danielhuppmann: pyam was started before typehints were cool (or before I knew about them), and I would like to add them as one documentation-harmonisation update for all methods.

pyam/core.py
danielhuppmann: In my original use case, it could be that the temperature stabilises at (for example) 1.7°,, in particular if we round the output from the climate assessment. So there could be multiple years where peak temperature is reached, without this being necessarily incorrect.

The truly correct solution would be to also provide an *on_method* argument that makes this unique?

